PR #11, opened 2023-12-05
User crud UN-140

PThorpe92:
Tests are included for the controller 

TODO: add pagination and sorting for Index, look into middleware

### DIFF
--- a/app/Http/Controllers/v1/UserController.php
+++ b/app/Http/Controllers/v1/UserController.php
@@ -1,12 +1,12 @@
 <?php
 
-namespace App\Http\v1\Controllers;
+namespace App\Http\Controllers\v1;
 
 use App\Http\Requests\StoreUserRequest;
 use App\Http\Resources\NewUserResource;
 use App\Http\Resources\UserResource;
 use App\Http\Controllers\Controller;
-use Illuminate\Http\Request;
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/v1/UserController.php
chrissantillan, line 1: @PThorpe92 should this file be in `app/Http/Controllers/**Api**/v1`?

PThorpe92: Yeah this is something we need to figure out:

The route itself would actually be `api/v1` either way if you define it in the `routes/api.php`

If you look into the controller, you'll see it has pre-generated methods (commented out) that return views, this confused me, and kind of seems to blend the two. I guess we need to determine if we are going to separate everything.

chrissantillan: Yeah. Was just wondering if we were keeping the same directory structure from Starfish in v2, where API controllers are kept in `Api/V1` and regular controllers (?) are kept in the parent Controllers dir.

chrissantillan (approved):
Approved. Going forward, team will need to agree upon directory structure for Controllers (front end view-based and back end API based)

nokierae (approved):

PR #19, opened 2023-12-07
feat: remove forgot pwd, change login to username, impl Auth flow

PThorpe92:
Login is now done with Username and there is no more `forgot password`
image

Changed Auth controller to instead set Temp password, and trigger `password_reset`

Login flow needs to be interrupted to check for `password_reset` in order to prompt user to set a new password. Probably going to just re-use UI for signing up?


chrissantillan (approved):
Request approved.

PR #29, opened 2023-12-13
feat: add ProviderPlaform CRUD functionality UN-183

chrissantillan:


app/Http/Controllers/v1/ProviderPlatformController.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        $providerPlatforms = ProviderPlatform::paginate(10);
+
+        return PaginateResource::make($providerPlatforms, ProviderPlatformResource::class);
+    }
+
+    /**
+     * Show the form for creating a new resource.
+     */
+    public function create()
```
PThorpe92, line 27: I know these are generated but I believe I removed them from the User one after our meeting as they aren't following the separation of concerns we outlined. not a big deal but they can be removed

app/Http/Requests/UpdateProviderPlatformRequest.php
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @return array<string, \Illuminate\Contracts\Validation\ValidationRule|array<mixed>|string>
+     */
+    public function rules(): array
+    {
+        return [
+            'name' => 'required|string|max:255',
+            'type' => [Rule::enum(ProviderPlatformType::class)],
+            'description' => 'nullable|string|max:255',
+            'icon_url' => 'required|url:http,https',
```
PThorpe92, line 32: For an Update on the provider platform request, I believe we should treat these as PATCH requests.

Since it requires an ID pulled from the uri to make the request in the first place(plus eventually `authorize`), I think we should make every field nullable, so the frontend can only send a JSON object with the 1-2 fields they wish to update, which will come from the Form. It will avoiding having to fetch or include unnecessary info in the request

PThorpe92 (approved):

PR #34, opened 2023-12-15
feat: add oxlint linting system for frontend code to githook and add github action

PThorpe92:
Currently we have `Prettier` running as a formatter but as the front end code grows, we will need something like `eslint` which is well known for being slow and __not__ written in Rust.  So the clear option is `oxlint` which is neither

### DIFF
--- a/.github/cliff.toml
+++ b/.github/cliff.toml
@@ -17,6 +17,7 @@ body = """
         #### {{ group | upper_first }}
         {% for commit in commits %}
             - {{ commit.message | upper_first }}\
+            Author: {{ commit.author.name }}
         {% endfor %}
     {% endfor %}\
 {% endfor %}\n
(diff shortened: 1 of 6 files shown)


nokierae:
This PR has oxlint run In the pre-commit hook, but I don't see it added as a dev dependency in the package.json file. It also looks like you've kept prettier as a dependency and the prettier ignore file. Do we still need those?

Otherwise looks great!

PThorpe92:
Good catch! This does need to be added as a dependency to the package.json file, 

It sounds unclear in my comment, but this is added in place of something like `eslint` for semantic and syntactic errors, and actually doesn't provide formatting like prettier, so we will be continuing to use prettier for formatting.

nokierae (approved):

PR #36, opened 2023-12-17
Base authenticated layout

nokierae:


.husky/check_unstaged.py
```diff
@@ -4,12 +4,12 @@
 
 
 def print_red(text):
-    return f"\033[91m{text}\033[0m"
+    return "\033[91m{}\033[0m".format(text)
```
PThorpe92, line 7: you dislike python `f`strings?

PThorpe92 (approved):

.husky/check_unstaged.py
nokierae: I love them! I was just making the script work in both python2 and 3.

PThorpe92: ahhhhh that makes sense lol

PR #46, opened 2024-01-03
feat: add Enrollments model and migration

coledykstra:


database/migrations/2024_01_03_145604_create_enrollments_table.php
```diff
@@ -0,0 +1,45 @@
...
+            $table->unsignedBigInteger('user_id');
+            $table->unsignedBigInteger('course_id');
+            $table->string('provider_id')->nullable();
+            $table->enum('enrollment_state', ['active', 'inactive', 'completed', 'deleted']);
+            $table->json('links');
+            $table->timestamp('provider_start_at')->nullable();
+            $table->timestamp('provider_end_at')->nullable();
+            $table->timestamps();
+
+            $table->primary('id');
```
nokierae, line 27: Id is automatically marked primary, so this line is causing issues and you can remove it.

database/migrations/2024_01_03_145604_create_enrollments_table.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('enrollments', function (Blueprint $table) {
+            $table->id();
+            $table->unsignedBigInteger('user_id');
```
nokierae, line 18: Instead of modifying the table on line 31, this line can just be `$table->foreignId('user_id')`. Laravel uses convention to know that it is referencing the user table and id field.

coledykstra:
I fixed these things but i don't understand whats happening

database/migrations/2024_01_03_145604_create_enrollments_table.php
```diff
@@ -0,0 +1,45 @@
...
+            $table->unsignedBigInteger('course_id');
```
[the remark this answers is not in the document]
chrissantillan: I second Nokie's comments re: Laravel's auto referencing convention

database/migrations/2024_01_03_145604_create_enrollments_table.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('enrollments', function (Blueprint $table) {
+            $table->primary('id');
```
nokierae, line 17: This can just be `$table->id();` which should fix the issue we are seeing in the build.

nokierae:
The conventional commits check is failing because it should be 'fix:' and not 'bug:'. I have linked the docs for how to change commit messages below. There is some complexity, so feel free to reach out to me and or preston if you get stuck.

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

PR #49, opened 2024-01-04
feat: User management

nokierae:
This isn't 100% complete with every feature and still needs some refactoring, but I need to push my updates to the UI so that Carolina has access to them. This also includes a couple of backend fixes.


PThorpe92 (approved):
LGTM :+1:

PR #51, opened 2024-01-10
Base oauth configuration, add canvas service and ProviderUserMapping and Endpoint

PThorpe92:


app/Http/Controllers/v1/ProviderLoginController.php
```diff
@@ -0,0 +1,126 @@
...
+
+        $logins = $query->paginate($perPage);
+
+        return ProviderUserMappingResource::collection($logins);
+    }
+
+    /**
+     * Create a new login to a given provider for a given User.
+     */
+    public function create(CreateProviderUserMappingRequest $request)
```
PThorpe92, line 48: oops, this wasn't completed yet. still working on it

app/Models/ProviderUserMapping.php
```diff
@@ -0,0 +1,29 @@
...
+class ProviderUserMapping extends Model
+{
+    use HasFactory;
+
+    protected $fillable = [
+        'user_id',
+        'provider_platform_id',
+        'external_user_id',
+        'external_username', // if the user has a field in the external platform that is different from 'username'
+        'authentication_provider_id',
```
nokierae, line 17: This might need to go on the provider platform or might not be needed at all.

PThorpe92: Yeah this will depend on whether it is possible that a 'login' for a user from our platform would ever have a login assigned to anything but our oidc server. If this exists on the Provider Platform model, then I think it would need to be nullable. 

Hard to tell at this point what unique info we may need about a user related to a provider login elsewhere

PThorpe92:
closing this in favor of #57

PR #53, opened 2024-01-16
Pulls and renders categories in left menu

calisio:
Adds new feature that pulls categories from the database and renders these categories and its links in the left menu.

### DIFF
--- a/resources/js/Components/LeftMenu.tsx
+++ b/resources/js/Components/LeftMenu.tsx
@@ -1,48 +1,69 @@
+import axios from "axios";
 import Brand from "./Brand";
 import {
     HomeIcon,
     ArchiveBoxIcon,
     BookOpenIcon,
 } from "@heroicons/react/24/solid";
+import useSWR from "swr";
 
 interface CategoryParameters {
     categoryName: string;
     linksArray: Array<LinksArrayParameters>;
+    rank: number;
 }
-// had to add undefined to this, but not sure if that is not allowed
 interface LinksArrayParameters {
-    [linkName: string]: string | undefined;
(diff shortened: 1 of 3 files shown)


nokierae:
Thanks @calisio ! I'll leave some feedback later this afternoon.

PThorpe92 (approved):

nokierae:
Update: I ran out of time tonight, so I'll have to leave feedback tomorrow.

resources/js/Components/LeftMenu.tsx
```diff
@@ -1,11 +1,70 @@
+import axios from "axios";
 import Brand from "./Brand";
 import {
     HomeIcon,
     ArchiveBoxIcon,
     BookOpenIcon,
 } from "@heroicons/react/24/solid";
+import useSWR from "swr";
+
+interface CategoryParameters {
```
nokierae, line 10: I would move these types to the resources/js/common.ts file and rename them to just Category and CategoryLink.

resources/js/Components/LeftMenu.tsx
```diff
@@ -1,11 +1,70 @@
...
+                    {categoryName}
+                </summary>
+                <ul>{linksList}</ul>
+            </details>
+        </li>
+    );
+}
+
+function getCategoryItems() {
+    const { data, error, isLoading } = useSWR("/api/v1/categories", (url) =>
```
nokierae, line 42: Technically, all hooks need to be in the body of the component and shouldn't be placed in functions. I would moev useSWR into the LeftMenu component and pass the data into this function.

PR #56, opened 2024-01-18
feat: add enrollments crud api

coledykstra:


nokierae:
I've reviewed, but I'd love a second pair of eyes on it @PThorpe92

app/Http/Controllers/v1/CategoryController.php
```diff
@@ -30,7 +29,7 @@ public function index()
 
         $query->orderBy($sortBy, $sortOrder);
 
-        $categories = $query->paginate($perPage);
+        $categories = $query->get();
```
PThorpe92, line 32: What is the reason we are getting rid of pagination here?

app/Http/Controllers/v1/EnrollmentController.php
```diff
@@ -0,0 +1,85 @@
...
+        return new EnrollmentResource($Enrollment);
+    }
+
+    public function store(StoreEnrollmentRequest $request)
+    {
+        $validated = $request->validated();
+
+        $Enrollment = Enrollment::create($validated->all());
+
+        return EnrollmentResource::collection($Enrollment);
```
PThorpe92, line 55: Why are we returning a collection here? is this expecting creating/many enrollments at once?

app/Http/Controllers/v1/CategoryController.php
coledykstra: Nokie suggested that we will have never enough categories to paginate.

app/Http/Controllers/v1/EnrollmentController.php
coledykstra: Thats how I did it in category, so i did the same thing for enrollments. I do not know if we will return many at once or not.

app/Http/Controllers/v1/CategoryController.php
PThorpe92: makes sense :+1:

app/Http/Controllers/v1/EnrollmentController.php
PThorpe92: as stated in slack, it's likely that the return json won't be used, and the frontend/consumer will simply be looking for a 201/created. i think worst case all it's doing is makes the front end have to index an additional field on the response. I'm fine with leaving it. this can be merged :+1:

PThorpe92 (approved):

PR #57, opened 2024-01-18
OIDC support #UN-219

PThorpe92:
This PR handles the integration of OAuth2 using laravel's Passport library, as well as an additional library https://github.com/jeremy379/laravel-openid-connect to handle the OIDC id_token layer on top.


app/Services/CanvasServices.php
```diff
@@ -63,7 +63,7 @@ public function __construct(int $providerId, int $accountId, string $apiKey, str
         $this->account_id = $accountId;
         $this->access_key = $apiKey;
         $this->base_url = $url;
-        $this->client = $this->getClient();
+        $this->client = new Client();
```
chrissantillan, line 66: Here is my solution to the Bearer token problem. In the GuzzleHttp docs it says when creating a new Client you can add a base_uri and any other request options. So here it is: 

`$this->client = new Client([
            'base_uri' => $this->api_url,
            'headers' => [
                'Authorization' => 'Bearer '.$this->access_key
            ]
        ]);`

Additionally the base_uri will prefix any relative URIs given as the first parameter in any request. See https://docs.guzzlephp.org/en/stable/quickstart.html

PThorpe92: i think since now all the methods have to do is outline the `url` string and the `body`, it should be fine. if we were having to include the headers in every request in every method than we would def want to figure out how to not duplicate code. also it is strange that it didn't work previously using that same method

chrissantillan (approved):

PR #65, opened 2024-01-24
auth working, providerusermapping controller, migration, fixed canvas services, dependency, etc etc

PThorpe92:


app/Models/ProviderUserMapping.php
```diff
@@ -0,0 +1,29 @@
...
+class ProviderUserMapping extends Model
+{
+    use HasFactory;
+
+    protected $fillable = [
+        'user_id',
+        'provider_platform_id',
+        'external_user_id',
+        'external_username', // if the user has a field in the external platform that is different from 'username'
+        'authentication_provider_id',
```
nokierae, line 17: What is the field supposed to be and why do we need it? In the factory its a number, in the migration it is a string, and in a test it is "openid_connect".

app/Models/ProviderPlatform.php
```diff
@@ -21,6 +21,7 @@ class ProviderPlatform extends Model
         'access_key',
         'base_url',
         'state',
+        'auth_provider_id',
```
nokierae, line 24: Should this be an 'external_' value?

tmp/private.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
nokierae, line 1: What are these tmp keys used for?

nokierae:
Yesterday it looked like there were some updates to the .env.example file and some instructions we probably needed to add to the README. Something to do with Passport I think. Do we still need those?

PThorpe92:
> Yesterday it looked like there were some updates to the .env.example file and some instructions we probably needed to add to the README. Something to do with Passport I think. Do we still need those?

Yes, good call some documentation around that will be needed for the readme

app/Models/ProviderUserMapping.php
PThorpe92: This is supposed to be the type of auth we are using with the provider, currently it defaults to openid_connect, but in the future it might be SAML or OAuth2. I'm going to make a new Enum for this, and we can enforce it like that, if that makes sense to you

nokierae: Just making a note that we talked about making this field 'authentication_provider_status'.

nokierae (approved):

PR #67, opened 2024-01-24
feat: create action/RPC for registering auth provider in canvas

PThorpe92:


app/Http/Controllers/v1/Actions/RegisterCanvasAuthProviderAction.php
```diff
@@ -0,0 +1,36 @@
...
+
+        $provider = ProviderPlatform::where(['id' => $valid['provider_platform_id']])->firstOrFail();
+
+        if (! $provider || ! in_array($provider->type, ['canvas_cloud', 'canvas_oss'])) {
+            return response()->json(['Provider Platform was either not found or is not a Canvas instance', 404]);
+        }
+
+        $cs = CanvasServices::byProviderId($provider->id);
+
+        $resp = $cs->createAndRegisterAuthProvider($valid->auth_provider_url);
```
nokierae, line 24: Don't we need to check to see if the auth provider hasn't already been created (and info stored on the provider platform)? Otherwise we'd create a whole new auth provider for this platform.

nokierae:
One minor comment but then LGTM.

app/Http/Controllers/v1/Actions/RegisterCanvasAuthProviderAction.php
PThorpe92: Yes and no, if fresh migrations are ran, the client id/secret is invalidated, so it's likely it will need to be ran again, and the database (might) not reflect this.. obviously fresh migrations would in fact reset this as well, but there are indeed situations where keys will be invalidated, but that field might not reflect it.

its honestly just a tough one.. i dont believe there is a clear right answer.

I dont want to make it impossible in that situation to run it again, but i dont want it to be ran twice accidentally either. this might have to be somehting that is done on the front end, where the UI will let them know that it exists, and makes sure the want to overwrite? lmk your thoughts

nokierae: I can see your point. There are different use cases for this in both dev and production. I'll think on it more and we can talk about it tomorrow.

nokierae: I think in the short term we will just be using this manually to set up the environment. In the future we can revisit this when we try to make it self service.

PR #72, opened 2024-01-29
feat: canvas course enrollment service UN-215

chrissantillan:


PThorpe92:
As soon as my PR is merged, there will be some conflicts and honestly i think i fixed most of the issues with the tests in this one, so I'll finish whatever else needs to be fixed after that's merged.

PThorpe92 (approved):
LGTM

app/Http/Controllers/v1/StoreUserEnrollmentController.php
```diff
@@ -0,0 +1,56 @@
...
+use App\Models\Enrollment;
+use App\Services\CanvasServices;
+use Illuminate\Http\Request;
+
+class StoreUserEnrollmentController extends Controller
+{
+    /**
+     * Handle the incoming request.
+     */
+    public function __invoke(Request $request, int $providerId, string $userId)
```
nokierae, line 18: should userID be an int here?

app/Http/Controllers/v1/StoreUserEnrollmentController.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+
+        $this->canvasService = CanvasServices::byProviderId($providerId);
+        $canvasEnrollments = $this->canvasService->getEnrollmentsByUser($userId);
+        $enrollmentCollection = collect();
+        foreach ($canvasEnrollments as $enrollment) {
+            if ($course = Course::where('provider_resource_id', $enrollment->course_id)->firstOrFail()) {
+                $request->merge([
+                    'user_id' => 1,
```
nokierae, line 27: should this be $userId instead of 1?

app/Http/Controllers/v1/StoreUserCourseController.php
```diff
@@ -0,0 +1,42 @@
...
+            $request->merge(['provider_resource_id' => (string) $course->id, 'provider_course_name' => $course->name]);
+            $validated = $request->validate([
+                'provider_resource_id' => 'required|string|max:255|unique:courses',
+                'provider_course_name' => 'required|string|max:255',
+            ]);
+
+            $courseCollection->push(Course::create($validated));
+        }
+
+        $courseCollection = Course::paginate(10);
```
nokierae, line 38: This should not be paginated. There is no way to request the next page since if you tried it would rerun the action and return nothing.

nokierae: Which also means that we need  a new resource to return.

app/Http/Controllers/v1/StoreUserEnrollmentController.php
```diff
@@ -0,0 +1,56 @@
...
+                    'course_id' => $course->id,
+                    'provider_user_id' => $enrollment->user_id,
+                    'provider_course_id' => $enrollment->course_id,
+                    'provider_id' => $providerId,
+                    // 'provider_course_name' => $enrollment->sis_course_id,
+                    'enrollment_state' => $enrollment->enrollment_state,
```
nokierae, line 33: Should there be a provider_resource_id storing the enrollment id from canvas? (If there is one of course.) And could we then validate for uniqueness? Right now it looks like if I ran this action twice it would create duplicates.

app/Http/Controllers/v1/StoreUserEnrollmentController.php
```diff
@@ -0,0 +1,56 @@
...
+                    'provider_id' => 'required|exists:provider_platforms,id',
+                    'enrollment_state' => 'required',
+                    'links' => 'nullable',
+                ]);
+
+                $enrollmentCollection->push(Enrollment::create($validated));
+            }
+        }
+
+        $enrollmentCollection = Enrollment::paginate(10);
```
nokierae, line 52: This should not be paginated. See above.

app/Http/Controllers/v1/StoreUserCourseController.php
PThorpe92: going with the course resource, where it just calls parent::toArray() either way should be alright?

nokierae: Yes

PR #77, opened 2024-02-02
feat: add controller test generation command rough draft

PThorpe92:
This provides a new command: `artisan app:make-controller-test {NameOfController}`

That will prompt the user for information about the controller, model, and visitbility and generate a class of tests that will test for all the basic CRUD operations and assert that either the route is protected (will fail if non-admin makes requests) or it will assert that the user will be returned only data that belongs to the user.

### DIFF
--- a/tests/Feature/EnrollmentControllerTest.php
+++ b/tests/Feature/EnrollmentControllerTest.php
@@ -6,6 +6,7 @@ use App\Models\Course;
 use App\Models\Enrollment;
 use App\Models\ProviderPlatform;
 use App\Models\User;
+use Database\Seeders\TestSeeder;
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Tests\TestCase;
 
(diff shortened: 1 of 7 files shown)


PThorpe92:
## The CLI will prompt the user for some information about the Controller they are trying to test
image

## and it will generate such a test class:
image
image

PThorpe92:
This has been tested with `enrollments`.. 

upon running the command and providing the model and route, and immediately running `sail artisan test` this program will correctly generate passing tests
image

tests/Feature/EnrollmentControllerTest.php
```diff
@@ -15,40 +16,84 @@ class EnrollmentControllerTest extends TestCase
...
+        $responseUser = $this->actingAs($user)->getJson($this->uri);
+
+        $responseAdmin = $this->actingAs($admin)->getJson($this->uri);
+
+        $responseAdmin->assertJsonCount(10, 'data');
+
+        $responseUser->assertJsonCount(5, 'data');
     }
 
-    // .... i have no words for some of these tests. We can do better
```
chrissantillan, line 31: "We can do better" ... is that a specifically or a general philosophical statement haha?

chrissantillan (approved):

tests/Feature/EnrollmentControllerTest.php
PThorpe92: Lol I guess technically both? I forgot I committed that :rofl:

PR #83, opened 2024-02-12
feat: add user activity model and migrations

coledykstra:


app/Models/UserActivity.php
```diff
@@ -0,0 +1,20 @@
...
+class UserActivity extends Model
+{
+    use HasFactory;
+
+    protected $fillable = [
+        'user_id',
+        'browser_name',
+        'platform',
+        'device',
+        'ip',
```
PThorpe92, line 17: Although this will typically be running on a local network, and no public IP's will be exposed we will probably want to:
```php
 protected $hidden = [
        'ip',
    ];
 ```

File: database/factories/UserActivityFactory.php [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+ * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\UserActivity>
+ */
+class UserActivityFactory extends Factory
+{
+    /**
+     * Define the model's default state.
+     *
+     * @return array<string, mixed>
+     */
+    public function definition(): array
```
PThorpe92, line 17: In order to properly test this, can we add a `forUser()` method on this factory, as demonstrated in https://github.com/UnlockedLabs/UnlockEdv2/blob/5600b52402e9955b7de86b23c7880ec955da768a/database/factories/EnrollmentFactory.php#L38

tests/Feature/UserActivityTest.php
```diff
@@ -0,0 +1,21 @@
...
+
+use App\Models\UserActivity;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Tests\TestCase;
+
+class UserActivityTest extends TestCase
+{
+    use RefreshDatabase;
+
+    public function testUserActivitiesAreCreated(): void
```
PThorpe92, line 13: Because there isn't a controller yet, I would just add your factory to the TestSeeder.php file, so you can use $this->seed(TestSeeder::class) in each test, which can auto create 10 and remove the need for this test.
 
The only thing I would maybe test, would be the `forUser()` method you create on the factory.

```php
// this will seed 10 
$this->seed(TestSeeder::class);
$user = User::factory()->createOne();
// this will add five more
UserActivity::factory(5)->forUser($user->id)->create();
// now you know there should be 10 created from the TestSeeder, and 5 for the User.
$all = UserActivity::all();
$five = [];
foreach($all as $one){
// if the user_id = $user->id, collect it into the $five array, and assert it contains five total.

coledykstra: I will add testing for `forUser` but I plan to add the controller today.

PThorpe92 (approved):

PR #84, opened 2024-02-12
user activity CRUD API

coledykstra:


app/Http/Requests/StoreUserActivityRequest.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public function authorize(): bool
+    {
+        // Check if the authenticated user is an admin
+        if ($this->user() && $this->user()->isAdmin()) {
+            return true; // Admins can create records for any users
+        }
+
+        // Check if the 'user_id' in the request matches the authenticated user's ID
+        return $this->user() && $this->user()->id == $this->input('user_id');
```
PThorpe92, line 20: Interesting.. I didn't know about the this->input() method. This would allow for some refactoring of some of the other requests that create another function to call from the controller because this one doesn't allow parameters (for the user_id) as it is called like middleware essentially. Good stuff

tests/Feature/UserActivityTest.php
```diff
@@ -29,4 +61,122 @@ public function testUserActivityFactoryCanCreateForSpecificUser(): void
         $this->assertEquals($user->id, $userActivity->user_id);
         $this->assertDatabaseCount('user_activities', 1);
     }
+
+    public function testAdminCanCreateUserActivitiesForAnyone()
```
PThorpe92, line 65: Excellent tests :+1:

PThorpe92 (approved):

PR #126, opened 2024-03-07
major fixes

PThorpe92:
Turns out, there was no way that the links and photos could have worked properly during the demo no matter what, just because of how things were set up to begin with. @calisio did an awesome job, everyone killed it just being able to get up what we did in time for the demo :muscle: 

PHOTO of freshly seeded local data that @calisio and I loaded in yesterday.

image

## Description

This PR fixes __lots__ of unnecessary duplication of fields between `courses` and `enrollments` (we were storing `provider_platform_id` on the enrollment but not the course?), and we were storing the `provider_course_id` on both the course and the enrollment, and naming it two different things? (provider_resource_id and provider_course_id?) There was no consistency with types, as `provider_platform_id` was a string and didn't actually reference the table for a relationship. Several other fields were of the wrong type, provider_course_id was an integer in one table and a string in the other.

A lot of these changes were hard to catch until the front-end showed up. Honestly I noticed something at 11:30 after trying to get the photos to show up locally, and ended up going down a rabbit hole of bugs. Here we are 11 hours and many coffees later :P

### Some notes on the assorted changes:
Enrollments/Courses:
I'm not against having a `links` field on the enrollments, but for the current use where its simply being used as the image URL for the course, I think that belongs on the course and not the enrollment (as it can be used for many courses), and in the future if we need many or multiple links for the enrollments, i think they would be better off in their own table, instead of storing JSON which is prone to syntax errors, and causes json parsing/serialization to happen 3 different times, we are casting or asserting JSON in our PHP code, that means its deserializing it, then mysql does a syntax check to assert the json syntax. If this was raw json we were storing that came directly from somewhere like canvas, then I think that would make sense, but i wasn't able see a links field or any corresponding field that's coming directly from [canvas](https://canvas.instructure.com/doc/api/enrollments.html).

Actions/CanvasServices:
As these are RPC's, I felt that they belong in the `/actions` URI, and pulling data out of the path makes less sense for actions, which can have a well-defined JSON body outlining the data/actions being requested. The action to grab the courses required the User ID, which are agnostic to users, unlike enrollments (I left a user specific Courses action in case it's needed, but also added one that grabs all the accounts courses). Also they did not account for the fact that CanvasServices is (was) a direct 1/1 of canvas's API calls, so it was feeding our User ID's to canvas as if they were canvas user ID's. I altered the behavior of canvasServices to work directly with our User ID's from now on.
 
Also, only the root level API key is allowed to query all enrollments for a user. Not even the Site admin key that is currently hooked up to the prod account was able to call that funciton for another user. However, I found a work-around by grabbing the courses and fetching the enrollments one-by-one.

note: All these changes required editing various parts all over the code-base that depended on the existing structures (tests, requests, resources, etc). I'm sure I made a mistake somewhere and I'm sure I'm forgetting something about a change I made. Please let me know if you'd like further explanation. 

Tests all pass, build looks good locally :+1:


PThorpe92:
wtf these tests are absurd i just want to go to sleep. literally haven't failed locally one time. the gh runner must use some ancient libsqlite.so

app/Models/Enrollment.php
```diff
@@ -12,32 +12,23 @@ class Enrollment extends Model
...
-        'provider_enrollment_id',
-        'provider_course_id',
+        'external_enrollment_id',
         'enrollment_state',
-        'links',
-        'provider_start_at',
-        'provider_end_at',
+        'external_start_at',
+        'external_end_at',
+        'link_url',
```
nokierae, line 19: Should this be external_link_url? Are we pulling the url from their data?

app/Http/Resources/EnrollmentCourseJoinResource.php
chrissantillan, line 1: Since this file is identical to `EnrollmentResource.php`, shouldn't we just get rid of one or the other?

app/Http/Controllers/v1/Actions/StoreUserEnrollmentController.php
```diff
@@ -14,41 +14,35 @@ class StoreUserEnrollmentController extends Controller
...
+                    'user_id' => $request->user_id,
                     'course_id' => $course->id,
-                    'provider_user_id' => $enrollment->user_id,
-                    'provider_course_id' => $enrollment->course_id,
-                    'provider_platform_id' => $providerId,
-                    'provider_enrollment_id' => $enrollment->id,
-                    // 'provider_course_name' => $enrollment->sis_course_id,
+                    'external_enrollment_id' => $enrollment->id,
                     'enrollment_state' => $enrollment->enrollment_state,
-                    'links' => [],
```
chrissantillan, line 33: Agreed, `links` was a question mark in my mind too. Vague catchall that didn't correspond to any canvas enrollment data.

app/Http/Resources/EnrollmentCourseJoinResource.php
PThorpe92: oh i see what you mean, If you look closely at the diff, that was actually a mistake from a git rebase that was corrected. now a plain EnrollmentResource will only return the enrollment object without any of the related data.

chrissantillan (approved):

PR #129, opened 2024-03-12
fix: make login button visible in light-mode

Outragedpillow:
Issue Image:
issueProof

After building this project and going to the sign-in page I could not see the login button, so I changed the log-in button color to teal (same as the UnlockEdv2 logo).

Here is what it looks like in light-mode:
issueFixLightMode

And what it looks like in dark-mode:
issueFixDarkMode

### DIFF
--- a/resources/js/Pages/Welcome.tsx
+++ b/resources/js/Pages/Welcome.tsx
@@ -12,7 +12,7 @@ export default function Welcome({ auth }: PageProps) {
                         <Brand />
                     </div>
                     <div className="flex-none">
-                        <ul className="menu menu-horizontal px-1 text-white">
+                        <ul className="menu menu-horizontal px-1 text-teal-500">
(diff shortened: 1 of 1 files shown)


PThorpe92 (approved):

resources/js/Pages/Welcome.tsx
```diff
@@ -12,7 +12,7 @@ export default function Welcome({ auth }: PageProps) {
                         <Brand />
                     </div>
                     <div className="flex-none">
-                        <ul className="menu menu-horizontal px-1 text-white">
+                        <ul className="menu menu-horizontal px-1 text-teal-500">
```
nokierae, line 15: Great catch. There are actually theme specific values that would probably be better to use (found in the tailwind.config.js file.) For this I would probably going with `text-primary` instead of `text-teal-500`.

PR #130, opened 2024-03-13
fix: make login button visible

Outragedpillow:
Issue proof:
issueProof

After I built and ran, I noticed that the login button was not visible in light mode, so I changed the text color of the login button to primary.

Here is what it looks like in light mode:
issueFixLightMode

And how it looks in dark mode:
issueFixDarkMode


Outragedpillow:
Unintentionally was adding commits to last pr. and had made a mess, so closed and made new pr that includes recent commits

nokierae (approved):

PR #133, opened 2024-03-13
Integration tests, user importation, daily user course activity auto (cron job) importation

PThorpe92:
OKAY

Changes in this PR:

1. Implement importing users from canvas.. currently this will import all unique users from canvas from the account that the master key is responsible for. 
2. integration testing. This adds several tests for getting data from canvas. including one that imports all known users, then imports all the courses, and enrollments for each user. 
3. UserCourseActivity.. Currently there are two fields on the Enrollment object from canvas that seem useless on their own. 
   - last_activity: the last date the user signed in to access the course referenced in the enrollment
   - total_activity_time: total number of seconds the user has spent on the specified enrollment
  alone, those are fairly useless, but if we collect them daily.. we can determine each day a user did or did not log in and access the course, and if we keep track daily of the amt of time, we can get a solid number or how many hours a user has spent on all their courses. helping @coledykstra populate his `github contribution` style component.
So this creates a Job that runs every day and introduces the UserCourseActivity table, that will keep these metrics for each student. 
Since we are fetching this data anyway daily, this opens a good opportunity to include other metrics from the users enrollment, (grades, etc), and calendar items (which i have another branch i'm already working on).
image

4. This adds protection for the ProviderPlatform API keys, and alters the front end to not show the key by default. There is now a toggle between showing and hiding the value, and by default the back end will not return the key. it must be requested individually by adding `api/v1/provider-platforms/{id}?show_key=true`. The tests have been adjusted to reflect this behavior.
image

Plus lots more stuff i probably forgot by now


chrissantillan (approved):

chrissantillan (approved):

PThorpe92:
NOTE: @calisio has some UI changes being pushed to this branch before it's merged.

chrissantillan (approved):

PR #134, opened 2024-03-14
fix: change course description and image array key names to match Canvas

chrissantillan:
Corrected the key names for course `description` and course img_url to match those in Canvas and v2 database. Replaced the nullish coalescing operators with else statements to stop description and img_url values from being repeated from previous course during course iteration process.


PThorpe92:
Good catch :+1: . this is a prime example of mind blowing `wtf`ness from canvas..

PR #139, opened 2024-03-16
feat (UN-244): Logs from UnlockED web app

corypride:


nokierae:
@PThorpe92 This looks pretty good to me, but I'm worried it might interfere with deployments. And since we won't have fluentd in our aws deployment yet, the console.logs might throw every time. Thoughts?

PThorpe92:
Yeah this looks good! :+1: 

So I agree we might have to put this on hold until these things are installed and configured outside of our local development environments. Glad we have a good template and most of the configs will already be done for us :rocket:

nokierae:
Converted this to a draft until we can figure out the best way to integrate it.

Issue #127, opened 2024-03-11
Bug: edit user does not set the correct default values after clicking one user, clicking "x", and then editing a different user


calisio:
Tested both locally and the live deployed version.

calisio:
Fixed by @Outragedpillow

Issue #124, opened 2024-03-06
User Modal Remains on Screen After Submitting


chrissantillan:
Edit user modal doesn't disappear after clicking "Submit Query" and the user doesn't show up in user list until after you refresh the screen.

nokierae:
I can't seem to reproduce this locally or on our test deployment. When I edit a user the modal closes and it refreshes in the list automatically. Where are you testing this, and if it is locally, is your branch up to date?

chrissantillan:
@nokierae I just tested it in production instance. Created a new user named John Doe and tried to make them an Admin. The modal still didn't disappear upon submission and when I refreshed the page the role was Student. I was in a private window in firefox on the latest branch.

nokierae:
Interesting. Being unable to set them as an admin is a known bug, but we'll have to look into the other one. I'll try to reproduce on Firefox. What OS are you on?

Outragedpillow:
I am on Fedora Linux and experience the same issue. I am going to put in a PR today that will fix this issue

calisio:
fixed in PR #141

Issue #108, opened 2024-03-05
Bug: Trying to set a user's role through the api doesn't seem to work. Defaults to 'student' even if 'admin' is sent.


nokierae:

calisio:
fixed in PR #141

PR #141, opened 2024-03-18
User Management functionality

Outragedpillow:
Features:
Added functionality to the reset user password button. This gives the user a temporary password to use and will have them reset it at the next login
uiPwResetInit

Success: 
uiResetPwSuccess

If an account is an admin give an error:
uiResetPwFail

Also added functionality to delete user button. For the moment allows deletion of both admin and student user types.
success: 
uiDeleteSuccess

These use a toast to display success or error that auto closes after a set time (5 seconds).
Rather than have a separate toast for different situations, made a dynamic toast that accepts a message, type, and has a bool for whether it should be displayed or not. 

Fixes:
#127 This was also fixed in my last PR and that code is implemented here as well
#124 modal now closes on submit. Response object was being misused.
#108 fixed issue. Enum validation was not working and had to be done manually for now. See Http/Controllers/v1/UserController.php


resources/js/Pages/Users.tsx
```diff
@@ -212,61 +278,93 @@ export default function Users({ auth }: PageProps) {
...
                     </h3>
+                    <h4 className="font-bold text-l text-secondary">
+                        <br />
+                        <p>Note: Only for non-administrator accounts.</p>
+                    </h4>
                     <p className="py-4"></p>
                     <div className="modal-action">
-                        <button className="btn btn-warning">Reset</button>
+                        <button
+                            className="btn btn-warning"
```
calisio, line 290: For consistency sake (until we get a component that is used for all modals), could you modify these modals to follow the same format as left menu management modals? ie. color should be btn-error (or you can change the left menu modals to btn-warning... either way), close button should be left aligned, and all modals should have an "x" on the top right corner. Super small changes (and I know they existed before your PR) but keeps the UI consistent!

resources/js/Components/forms/AddUserForm.tsx
```diff
@@ -25,99 +30,119 @@ export default function AddUserForm({ onSuccess }: { onSuccess: () => void }) {
...
+                    </div>
+                    <input
+                        type="text"
+                        className="input input-bordered w-full"
+                        {...register("username", {
+                            required: "Username is required",
+                            maxLength: {
+                                value: 50,
+                                message:
+                                    "Username shoudl be 50 characters or less",
```
calisio, line 114: small type on "should"

resources/js/Pages/Users.tsx
```diff
@@ -212,61 +278,93 @@ export default function Users({ auth }: PageProps) {
...
-                    <CheckCircleIcon className="h-6" />
-                    <span>User deleted!</span>
+            {toast.isVisible && (
+                <div
+                    className={`toast transition-opacity duration-500 ease-out ${
+                        toast.isVisible ? "opacity-100" : "opacity-0"
+                    }`}
+                >
+                    <div className={`alert alert-${toast.type}`}>
+                        <CheckCircleIcon className="h-6" />
```
calisio, line 363: When its an error, can you render `<ExclamationCircleIcon className="h-6" />` instead?

calisio (changes requested):
Super small changes! @PThorpe92 will review StudentNewPasswordController and UserController.

PThorpe92 (approved):
Other than what @calisio pointed out, looks good to me on the back-end. 

That UserRole enum issue has been driving me insane. I tried __everything__ and the only way to fix it was manually editing the user model like you did. :+1:

Outragedpillow:
Thanks for the feedback @calisio! I will make these changes today. And yeah @PThorpe92 I wasn't sure if that was the best fix but it works for now

calisio (approved):

PR #152, opened 2024-03-22
fix: fix incorrect title and disabled pagination buttons when in an unusable state

techeclecticdesign:


nokierae (approved):
Thanks for the fixes! Looks good to me.

PR #154, opened 2024-03-23
Couple fixes to Welcome Page

techeclecticdesign:
Made the heading text in the timeline readable - it had been set to white on gray-100.
Changed the logo at top left to behave like an SVG image (text not selectable, cursor does not change on hover).
Progressively load the large image on Welcome to boost speed for low bandwidth connections.


resources/js/Pages/Welcome.tsx
```diff
@@ -99,7 +114,7 @@ export default function Welcome({ auth }: PageProps) {
                                     <time className="font-mono italic text-secondary">
                                         1998 - 2017
                                     </time>
-                                    <div className="text-lg font-black text-white">
+                                    <div className="text-lg font-black text-gray-700">
```
PThorpe92, line 117: So daiseyUI allows us to set a theme, and then the colors chosen for the theme are available as a set of options for the tailwind components, such as`text-accent`, `text-error`, `btn-primary`, `btn-secondary`, etc.. 

So can these be changed to `text-neutral` or `text-primary` or whatever is closest to the gray you chose?

techeclecticdesign: Got it.  Thanks for the feedback :thumbsup:

PThorpe92 (approved):

PThorpe92:
Thanks for the PR!

PR #162, opened 2024-03-26
feat: add start_date and end_date parameters for UserActivityMap

coledykstra:


app/Http/Controllers/v1/UserActivityMapController.php
```diff
@@ -3,15 +3,31 @@
...
 use App\Models\UserCourseActivity;
 use Illuminate\Support\Facades\DB;
 
 class UserActivityMapController extends Controller
 {
-    public function show($id)
+    public function show($id, UserActivityMapRequest $request)
     {
-        $aggregatedData = UserCourseActivity::where('user_id', $id)
+        $startDate = $request->input('start_date');
```
PThorpe92, line 15: In order to validate the input from the UserActivityMapRequest, you need to call something like 
```php
$valid = $request->validated();
$start_date = $valid['start_date'];
// ... etc
```

PThorpe92 (approved):

PR #167, opened 2024-03-28
Some improvements on the pagenav dropdown.

techeclecticdesign:
-fixed a bug where clicking on the light/dark theme toggle switch would make the indicator invisible.
-made it so that clicking outside of the dropdown when it is active will cause it to close.


calisio (approved):
Looks great, thanks for fixing these bugs!

PR #181, opened 2024-04-04
fix(176): fix bug in User search

techeclecticdesign:
It had thought it was still on page 2 of the search results, even though the search results were all contained on page 1.


PThorpe92 (approved):
nice, I had assumed it would be a back-end fix but considering how it works I think that should do just fine :+1:

PR #182, opened 2024-04-07
Frontend fixes

techeclecticdesign:
A batch of small and practical UI fixes.

### DIFF
--- a/resources/js/Components/CategoryItem.tsx
+++ b/resources/js/Components/CategoryItem.tsx
@@ -5,6 +5,7 @@ import AddLinkForm from "./forms/AddLinkForm";
 
 import {
     ChevronDownIcon,
+    ChevronRightIcon,
     TrashIcon,
     PlusIcon,
     ChevronUpIcon,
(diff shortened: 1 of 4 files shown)


resources/js/Components/forms/AddUserForm.tsx
calisio, line 1: Currently these edits aren't changing anything on the frontend, because when a new user is created `ShowTempPasswordForm.tsx` should display "You have successfully created a new user." and does not depend on the user's name. If there is a bug I am not seeing that depends on sending the new user information back, let me know how I can reproduce it and I'll check it out!

calisio: All other edits look great though!

techeclecticdesign: To replicate the bug, go to the Users settings page, click on the reset password icon next to a non-admin user, then click on the button Reset Password.

PThorpe92: Is this for creating a new user or for resetting an existing user's password?
This is from our staging deployment after resetting a fake test-student password:
image

techeclecticdesign: I'm very sorry, you are absolutely right that is the behavior.  Not sure what changed, but no matter.  The commit in question is unnecessary now, and i will rebase it out of the pull request in just a moment.

calisio (approved):
Looks great! Thanks for fixing these 🙂

PR #185, opened 2024-04-10
Deletes old frontend files that are no longer relevant

calisio:


PThorpe92 (approved):
always good to see :+1: 
image

PR #195, opened 2024-04-15
feat (183): User Activity Map

techeclecticdesign:
Colors defined in tailwind.config in existing daisyui themes.
Responsiveness by listening to window resize event and adjusting size to 1 of 3 breakpoints.
DaisyUI tooltips.
Large amount of code for a front end component, but nothing too resource hungry that I can tell.
Open to any adjustments to make this best fit with the UnlockEdv2 platform.

image
image

### DIFF
--- a/resources/js/Components/UserActivityMap.tsx
+++ b/resources/js/Components/UserActivityMap.tsx
@@ -1,0 +1,392 @@
+import { useState } from "react";
+import { DropdownControl } from "./inputs";
+import useSWR from "swr";
+import { User } from "@/common";
+
+type ActivityMapData = {
+    user_id: number;
+    date: string;
+    active_course_count: string;
+    total_activity_time: string;
+    total_activity_time_quartile: number;
+};
+
+/* interface for dynamically generated year options */
+interface ValidYears {
+    [year: string]: string;
+}
+
+/* subtract a year from a date */
+const subtractYear = (date: Date) => {
+    const newDate = new Date(date);
+    newDate.setUTCFullYear(date.getUTCFullYear() - 1);
+    newDate.setUTCDate(newDate.getUTCDate() + 1);
+    return newDate;
+};
+
+/* predefining tailwind strings lets us assign them dynamically later */
+const quartileColors: string[] = [
+    "bg-quartile-0",
+    "bg-quartile-1",
+    "bg-quartile-2",
+    "bg-quartile-3",
+    "bg-quartile-4",
+];
+
+/* node sizes for the activity map */
+const nodeSizes: string =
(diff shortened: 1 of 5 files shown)


PThorpe92:
That looks awesome! Will take a look at this tomorrow. Good stuff man

calisio:
This is SO great!! I'm looking at it today too :) Thank you for tackling this!

resources/js/Components/UserActivityMap.tsx
```diff
@@ -0,0 +1,371 @@
...
+            "/api/v1/user-activity-map/" +
+                user.id +
+                "?start_date=" +
+                startDate.toISOString().split("T")[0] +
+                "&end_date=" +
+                yearEnd.toISOString().split("T")[0],
+        );
+        const data = await response.data;
+        setActivityData(data);
+    };
```
calisio, line 88: Rather than use axios.get, we have been using a react hook to fetch data called useSWR (https://swr.vercel.app/). You can see an example in `Users.tsx` (and other files like provider platforms, user activity, etc). With this you can also handle loading & errors and render something accordingly.
This is totally my bad because I completely forgot to add this into the `CONTRIBUTING.md` file. I'll do that soon :)

resources/js/Components/UserActivityMap.tsx
```diff
@@ -0,0 +1,371 @@
...
+    }, [aggregateActivityTime]);
+
+    /* add empty cells for days of week after activity range */
+    for (i = dateCount.getUTCDay(); i < 7; i++) {
+        tableData.push(
+            <td className="block" key={i + 7}>
+                <div className={`${nodeSizes[size]}`}></div>
+            </td>,
+        );
+    }
```
calisio, line 325: do you think for this we could just add grey cells until the end of the year? I just was thinking this because its similar to how github does it. Open to what you think looks best though. @PThorpe92 do you have any thoughts / feelings towards this?
Screenshot 2024-04-17 at 11 20 00 AM
Screenshot 2024-04-17 at 11 21 30 AM

techeclecticdesign:
Thank you for the feedback, if there's anything else let me know. I will have your recommended changes complete by this weekend.

resources/js/Components/UserActivityMap.tsx
PThorpe92: I like that :+1:  I think grey by default, so there is a year's worth at all times is the way to go.

calisio:
That sounds great. Would you also be able to put one more node in the legend that is just a grey square? For a total of 5 squares

techeclecticdesign:
I believe this is done now.  In addition to the recommended fixes, I added months at top, added a  semi-transparent border to the activity cells (similar to how github looks), and in general improved behavior and appearance.  Let me know if there are any other adjustments needed.

calisio (approved):
This looks great! Thanks for making those changes 🙂

PR #254, opened 2024-05-30
feat: add Puppeteer student discussion posts to Canvas

coledykstra:


canvas-discussion-bot/index.js
```diff
@@ -0,0 +1,137 @@
...
+    const iframe = await iframeElement.contentFrame();
+    await iframe.waitForSelector('body');
+    await iframe.type('body', faker.lorem.sentences(8));
+
+    // Click the "Save" button
+    await page.waitForSelector('button.btn.btn-primary.submit_button');
+    await page.click('button.btn.btn-primary.submit_button');
+
+    // Add a 5-second delay
+    await new Promise(resolve => setTimeout(resolve, 5000));
```
PThorpe92, line 82: can probably just do 
```ts
    await Promise.all([
      page.waitForNavigation(),
      page.click('button.btn.btn-primary.submit_button')
    ]);
```
instead of delaying 5 seconds to wait for navigation and submission.

canvas-discussion-bot/index.js
```diff
@@ -0,0 +1,137 @@
...
+const result = dotenv.config({ path: envFilePath });
+
+if (result.error) {
+  console.error('Error loading .env file:', result.error);
+} else {
+  console.log('Successfully loaded .env file');
+}
+
+async function postDiscussion(username, password, courseId, index) {
+  const browser = await puppeteer.launch({ headless: false });
```
PThorpe92, line 38: Is there a reason this can't be run in headless mode?

canvas-discussion-bot/.env.example
```diff
@@ -0,0 +1,3 @@
+API_TOKEN=my_api_token
```
PThorpe92, line 1: can you move this into the canvas-seeder directory so they can share the same .env file?

canvas-discussion-bot/index.js
coledykstra: Yeah, i don't need that all. That was a debugging delay, just so I could see how far it got.

canvas-discussion-bot/index.js
coledykstra: Yes, it can. Again, that was just so I could see what was happening.

canvas-discussion-bot/.env.example
PThorpe92: it's still a canvas seeder, we don't want to pollute the top level directory, plus it's the same exact .env file.

coledykstra: Sorry I'm polluting your code. Can I just assume the other project is there? How do I change `index.js` to get that to work. Do I need something like ```// Specify the path to the .env file for the local directory
const envFilePath = path.resolve(__dirname, '.env');

// Load the .env file
const result = dotenv.config({ path: envFilePath });```

PThorpe92: you can just put all the files into the existing `canvas-seeder`directory

coledykstra: I thought you were saying to just move the single line 
`API_TOKEN=my_api_token` 
I can put the whole node.js project into the same folder as the Go project?

PThorpe92: yeah definitely, just put it all in the 'canvas-seeder' directory

PThorpe92: and the other directory can be removed now that it's merged into canvas-seeder

PR #270, opened 2024-06-23
refactor: Make logging more structured and consistent

techeclecticdesign:
This is a WIP.  I think it could be better, and some feedback would be welcome.
In particular I am wavering on logging errors where ORM calls fail.  Either way the error message generated by GORM would be propagated to the handlers.  I'm not sure if I should log every failure in backend/databases/, or cut back on the ones that are there now.  My initial thought is instead of logging a separate entry, instead returning a modified error object that includes both the GORM error message and some information about which action failed, which then gets presented when an error is thrown in the handler.
I'm also interested in hearing in whether I should include more fields in the handlers.  I went pretty lightweight in terms of what data I included, which is both readable and scannable.  There is a lot of potentially useful information in the request headers, but I don't want to make it too spammy, either.

### DIFF
--- a/backend/src/handlers/actions.go
+++ b/backend/src/handlers/actions.go
@@ -139,11 +139,7 @@ func (srv *Server) HandleGetUsers(w http.ResponseWriter, r *http.Request) {
 				Message: "Successfully fetched users from provider",
 				Meta:    models.NewPaginationInfo(page, perPage, int64(total)),
 			}
-			if err := srv.WriteResponse(w, http.StatusOK, response); err != nil {
-				log.Error("Error writing response:" + err.Error())
-				srv.ErrorResponse(w, http.StatusInternalServerError, err.Error())
-				return
-			}
+			srv.WriteResponse(w, http.StatusOK, response)
 			return
 		} else {
 			delete(cachedProviderUsers, uint(service.ProviderPlatformID))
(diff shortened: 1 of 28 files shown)


backend/src/handlers/actions.go
```diff
@@ -60,19 +68,25 @@ func (srv *Server) HandleImportUsers(w http.ResponseWriter, r *http.Request) {
 }
 
 func (srv *Server) HandleImportPrograms(w http.ResponseWriter, r *http.Request) {
+	logFields := log.Fields{
```
PThorpe92, line 71: I like this method of declaring the log fields, and making sure the handler name is one of them each time :+1:

PThorpe92:
All around looks really good I'm super happy with it :+1: 

I know what you mean as far as the database calls. I recently just started returning the error instead of logging and propogating up
```go
// e.g. instead of
 if err := Db.Conn.Table("users").Where("id = ?", userId).Find(&users).Error; err != nil {
    log.Error("error retrieving users")
   return nil, err
} 

// have been switching to this:
return users, Db.Conn.Table("users").Where("id = ?", userId).Find(&users).Error
}
```
Which will cut down on that.


This is super helpful, we will be in a good place with logging. Much appreciated :+1:

techeclecticdesign:
I redid the database portion, and added a helper function, LogDbError, to reduce boilerplate and standardize error logging on the DB side.  It does propagate calls, but it makes the error messages on the handler side include more context than the very vague GORM errors.  On the database side the errors include context such as what method the error was in, along with an error message describing the query that failed.  Let me know if you like this, or if I maybe overthought things.  Error handling is an important and often overlooked part of a program, and I am happy to put time in to make this as good as possible.

techeclecticdesign:
I only just saw your comment earlier, and reflecting on it, I am going to make one quick set of changes so that there is only one error reported when database has an error.  Be done this weekend.

techeclecticdesign:
Couple last minute changes that make this a little better:
I eliminated 2nd argument of LogDbError because half the time it was repeating itself from the error message in handler.
I added in the method with the failing GORM call as part of the error message formed by LogDbError.

PThorpe92 (approved):
This took a while to fix all the merge conflicts because of how many changes have been made in the past couple weeks.
Gonna see if the tests all pass, but I was pretty careful about it, and I might still pull it down and test again locally.
But thanks for the PR, this greatly improves our logging situation :+1:

PR #293, opened 2024-07-03
Adds admin dashboard UI & connects to backend

calisio:


frontend/src/Pages/AdminDashboard.tsx
```diff
@@ -0,0 +1,83 @@
...
+                            <tr className="flex flex-row justify-between border border-x-0 border-t-0 mt-2">
+                            <th className="body text-grey-4">Course Name</th>
+                            <th className="body text-grey-4">Time Spent</th>
+                            </tr>
+                        </thead>
+                        <tbody className="flex flex-col gap-6 mt-4">
+                            {!error && !isLoading &&
+                            (data.top_program_activity.map((course: any, index: number) => {
+                                var courseTime:string
+                                if (course.hours_engaged < 1) courseTime = course.hours_engaged * 60 + " min"
```
PThorpe92, line 55:
```suggestion
                                if (course.hours_engaged < 1) courseTime = Math.round(course.hours_engaged * 60) + " min"
```

looks like we need one more `Math.round` here. (not that it matters too much cuz it wouldn't be sub 1hr for long)

frontend/src/Components/ConvertSeconds.tsx
```diff
@@ -0,0 +1,17 @@
+const convertSeconds = (secs: number) => {
```
PThorpe92, line 1: Not even a nit: just a thought

Personally, I am partial to terse code so I would do something like
```ts
const convertSeconds = (secs: number) => {
  const hours = Math.floor(secs / 3600);
  const minutes = Math.floor((secs % 3600) / 60);
  const seconds = Math.floor(secs % 60);

  return hours ? { number: hours, label: `hour${hours === 1 ? '' : 's'}` }
       : minutes ? { number: minutes, label: `minute${minutes === 1 ? '' : 's'}` }
       : { number: seconds, label: `second${seconds === 1 ? '' : 's'}` };
}; export default convertSeconds;
```
the only reason I say is because I see that some braces are absent to make it more concise, and I have found leaving them out (usually in c/c++) can be a footgun because it might not be obvious to the next person. That's entirely up to you either way tho. I just gotta comment on something I guess :)

PThorpe92 (approved):

frontend/src/Components/ConvertSeconds.tsx
calisio: No definitely way better do do it this way... I'll get that changed. Thanks!

PR #295, opened 2024-07-06
feat(266): Create a general component for search

techeclecticdesign:
I didn't need to do a component for dropdowns to set order - the DropdownControl component should cover this nicely.


PThorpe92:
Nice, this looks good to me. I'll wait on @calisio's review though.

Search is actually on my list to implement for the backend on a couple endpoints so this works out :+1: 
Thanks for the PR

calisio (approved):

Issue #272, opened 2024-06-24
Add password requirements to UI


calisio:
We need to add password requirements to the password reset UI. There is no design as of now, but we'd love for it to look something along these lines:
<img width="360" alt="Screenshot 2024-06-24 at 4 44 34 PM" src="https://github.com/UnlockedLabs/UnlockEdv2/assets/89868388/84f08e81-9ec4-4354-97ce-6eb1f38585a4">
As of now we only have two requirements: the password must be greater than 8 characters and contain at least 1 number (so you can ignore the first two lines on that example above).

Please use colors that are already included in the theme. Symbols are important in terms of accessibility, so please include those also (as in using a check mark and an x). Let me know if you have any follow up questions and/or if you'd like to pick this up.

jtucholski:
@calisio Should the password requirements list be positioned below the Password or Confirm Password field?

PR #297, opened 2024-07-08
feat: adds password ui validation

jtucholski:
Adds password UI validation and should close https://github.com/UnlockedLabs/UnlockEdv2/issues/272

Implements following UI changes:
* New password messages when the password does/doesn't meet the minimum length or include one number
* New message when the confirm password doesn't match the new password
* Form submission is blocked if any of the validation rules fail

image


calisio (approved):
Looks great!

Issue #267, opened 2024-06-21
My courses search/sort functionality


calisio:
# Overview

The student view of “my courses” does not have a search or sort functionality implemented yet. Below is a view of what the UI should look like.
<img width="1471" alt="my courses search:sort" src="https://github.com/UnlockedLabs/UnlockEdv2/assets/89868388/256fbf69-3e31-4a55-8fb9-35f5836cbeab">

# Relevant files

These should live in the `MyCourses.tsx` file. The handler is in `/backend/src/handlers/dashboard.go` and the call to the database is in `/backend/src/database/user_catalogue.go` . 

# Sorting

Filtering should already partially be available through the different tabs. This needs to be updated to include a few more filters. Users should additionally be able to sort by:

- alphabetically
- progress

# Tasks to complete ticket

- [ ] Update backend api call and table join to query for filtering of courses
- [ ] Update frontend to include search and filter dropdown
- [ ] Connect frontend to backend for search and filter

PR #301, opened 2024-07-11
fix(299): fix search, remove chevrons in activity and users pages

techeclecticdesign:
This PR depends on the feat(266) PR.
Because of this dependency, until that PR is resolved there is an extra commit being shown as a change.
I think that's how it works at least, and I hope that it is ok to do a PR like this before the previous one is accepted.  Let me know if not.

### DIFF
--- a/backend/src/handlers/user_handler.go
+++ b/backend/src/handlers/user_handler.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"slices"
 	"strconv"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 )
@@ -26,7 +27,6 @@ func (srv *Server) registerUserRoutes() {
 **/
 func (srv *Server) HandleIndexUsers(w http.ResponseWriter, r *http.Request) {
 	page, perPage := srv.GetPaginationInfo(r)
-	search := r.URL.Query().Get("search")
 	include := r.URL.Query()["include"]
 	if slices.Contains(include, "logins") {
 		srv.HandleGetUsersWithLogins(w, r.WithContext(r.Context()))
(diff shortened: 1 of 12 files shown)


calisio:
Yes, totally fine! Sorry for the delay on getting #295 back, there are some things with theming I'm still deciding, but in general that component looked good. Hopefully I'll be able to review both of these tomorrow. Thank you for this!

techeclecticdesign:
No problem and no hurry.  #267 is done too.  I'm assuming it would be more convenient for you if I wait to send the PR until review, but if you prefer to do the 3 together I can do that also.

PThorpe92:
Ook my fault I had search implemented locally for a few days and forgot to push it.

I'll just fix the conflicts and make it work :+1:

PThorpe92 (approved):

backend/src/handlers/user_handler.go
```diff
@@ -37,8 +37,10 @@ func (srv *Server) HandleIndexUsers(w http.ResponseWriter, r *http.Request) {
 		srv.HandleGetUnmappedUsers(w, r, providerId)
 		return
 	}
+	order := r.URL.Query().Get("order")
+	search := strings.ToLower(r.URL.Query().Get("search"))
```
PThorpe92, line 41: Actually I forgot this when I pushed the changes:

Can we add a 
```go
search = strings.TrimSpace(search)
```
right below this? without it, it will neglect to return some results when someone pushes the spacebar in an empty search bar.

techeclecticdesign: yes.  I applied the same to user_activity_handler as well.

calisio (approved):

calisio:
Thanks @techeclecticdesign !!

PR #306, opened 2024-07-12
feat: adds resources to right side bar

### LINKED ISSUE #305: Add Resources to right menu for temporary access to open content
We need a place for left menu links to display on the student portal temporarily, and a place to link to our Kolibri integration. This will live in the right menu where notifications were previously hard coded. Again, this is a TEMPORARY solution until we finalize how we want to structure open content/resources (in the future it will probably live in a new page). Below is a preview of what this will look like:

Image

calisio:
Closes #305


PThorpe92 (approved):
:tada:

PR #311, opened 2024-07-16
fix: empty recent programs for user-dashboard

PThorpe92:

PR #315, opened 2024-07-20
feat(268): add sort to MyProgress.tsx

techeclecticdesign:


backend/src/handlers/outcomes_handler.go
```diff
@@ -28,11 +28,13 @@ func (srv *Server) HandleGetOutcomes(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 	}
 
-	// Get type from query param
+	// Get vars from query params
 	typeString := r.URL.Query().Get("type")
+	order := r.URL.Query().Get("order")
+	orderBy := r.URL.Query().Get("orderby")
```
PThorpe92, line 34: I would like to make all these standard across all the endpoints. Can we go with `order_by` ?

backend/src/database/outcomes.go
```diff
@@ -18,6 +18,11 @@ func (db *DB) GetOutcomesForUser(id uint, page, perPage int, outcomeType models.
 	if err := query.Count(&count).Error; err != nil {
 		return 0, nil, err
 	}
+
+	if order != "" && orderBy != "" {
```
PThorpe92, line 22: if you look at the PR i have in right now #311 I created something like this 
```go
func validOrder(str string) string {
	if slices.Contains([]string{"asc", "desc"}, strings.ToLower(str)) {
		return strings.ToLower(str)
	}
	return "desc"
}
```
which you can use `query = query.Order(orderBy + " " + validOrder(order))

but also take a look to see how i sanitized the `order` string, with a map of all the possibilities. can we do something similar?

techeclecticdesign: Absolutely.  That fieldmap with an 'ok' is a nice pattern I like how that looks; glad you pointed it out.

techeclecticdesign: Should be all set.

PThorpe92 (commented):
This looks good, I'll pull it down and test it thoroughly tomorrow. We have a demo coming up very soon so I am being very cautious about what we merge before then, but I do want the feature so hopefully can get this merged in tomorrow :+1: 

thanks for the PR

techeclecticdesign:
Okay, I will double check things just in case tonight.
If there's anything else you want done before demo or I can help in anyway, let me know.  I have a five day weekend from work so I will have a lot of extra time after tomorrow.

PThorpe92 (approved):
This looks good, sorry it has sat around for so long. 

Do you maybe want to add the changes to the new one, or rebase and fix the merge conflicts? It's up to you

PR #316, opened 2024-07-21
fix: Give validation error message when creating preexisting username

techeclecticdesign:
Fixes unresponsive submit click caused by unique constrained fields not being validated in the AddUserForm and EditUserForm

### DIFF
--- a/backend/src/handlers/user_handler.go
+++ b/backend/src/handlers/user_handler.go
@@ -157,6 +157,12 @@ func (srv *Server) HandleCreateUser(w http.ResponseWriter, r *http.Request) {
 	if user.User.FacilityID == 0 {
 		user.User.FacilityID = srv.getFacilityID(r)
 	}
+	userNameExists := srv.Db.IsValidUsername(user.User.Username)
+	if userNameExists {
+		srv.ErrorResponse(w, http.StatusBadRequest, "userexists")
+		return
+	}
+
 	newUser, err := srv.Db.CreateUser(&user.User)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
@@ -231,8 +237,14 @@ func (srv *Server) HandleUpdateUser(w http.ResponseWriter, r *http.Request) {
 		srv.ErrorResponse(w, http.StatusNotFound, "user not found")
 		return
 	}
+	if toUpdate.Username != user.Username && user.Username != "" {
+		userNameExists := srv.Db.IsValidUsername(user.Username)
+		if userNameExists {
+			srv.ErrorResponse(w, http.StatusBadRequest, "userexists")
+			return
+		}
+	}
 	models.UpdateStruct(&toUpdate, &user)
-
 	updatedUser, err := srv.Db.UpdateUser(toUpdate)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/frontend/src/Components/forms/AddUserForm.tsx
+++ b/frontend/src/Components/forms/AddUserForm.tsx
@@ -27,6 +27,7 @@ export default function AddUserForm({
     reset,
     register,
     handleSubmit,
+    setError,
     formState: { errors },
   } = useForm<Inputs>();
 
@@ -48,7 +49,12 @@ export default function AddUserForm({
         ToastState.success,
       );
     } catch (error: any) {
-      setErrorMessage(error.response.data.message);
+      const response = error.response.data.trim()
+      if(response === "userexists") {
+        setError("username", { type: "custom", message: "Username already exists" });
+      } else {
+        setErrorMessage(error.response.data);
+      }
     }
   };
 
@@ -78,7 +84,10 @@ export default function AddUserForm({
 
   return (
     <div>
-      <CloseX close={() => reset()} />
+      <CloseX close={() => {
+        setErrorMessage("");
+        reset()
+      }}/>
       <form onSubmit={handleSubmit(onSubmit)}>
         <TextInput
           label={"First Name"}
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -143,10 +143,26 @@ func (db *DB) DeleteUser(id int) error {
 
 func (db *DB) GetUserByUsername(username string) *models.User {
 	var user models.User
+	var count int64
+
+	_ = db.Conn.Model(models.User{}).Find(&user, "username = ?", username).Count(&count)
 	if err := db.Conn.Model(models.User{}).Find(&user, "username = ?", username).Error; err != nil {
 		return nil
 	}
+	if count == 0 {
+		return nil
+	}
 	return &user
+}
+
+func (db *DB) IsValidUsername(username string) bool {
+	userExists := false
+	err := db.Conn.Raw("SELECT EXISTS(SELECT 1 FROM users WHERE username = ?)", username).
+		Scan(&userExists)
+	if err != nil {
+		log.Error("Error checking if username exists: ", err)
+	}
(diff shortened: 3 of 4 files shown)


frontend/src/Components/forms/AddUserForm.tsx
```diff
@@ -29,23 +31,29 @@ export default function AddUserForm({
     handleSubmit,
     formState: { errors },
   } = useForm<Inputs>();
+  let users = userData ? userData.data : [];
 
   const onSubmit: SubmitHandler<Inputs> = async (data) => {
+    for (let i = 0; i < users.length; i++) {
```
PThorpe92, line 37: due to how pagination is handled on the back-end, we might need to add maybe a query string arg, on the show-user endpoint that returns whether it's a valid/available username?

techeclecticdesign: I see what you mean, I didn't even consider pagination.  Glad you caught that.

techeclecticdesign: Moved logic to test for existing usernames to the backend.
There was an issue with getUserByUsername in users.go, where an empty struct was being returned when no match instead of nil, where nil is the expected no match value throughout other code in this project (ory.go, create user test), so I fixed that as well as added a getUserByEmail function.
Should be all set.

backend/src/handlers/user_handler.go
```diff
@@ -157,6 +157,17 @@ func (srv *Server) HandleCreateUser(w http.ResponseWriter, r *http.Request) {
 	if user.User.FacilityID == 0 {
 		user.User.FacilityID = srv.getFacilityID(r)
 	}
+	userNameExists := srv.Db.GetUserByUsername(user.User.Username)
+	if userNameExists != nil {
+		srv.ErrorResponse(w, http.StatusInternalServerError, "Username already exists")
```
PThorpe92, line 162: Can we return something like BadRequest or Forbidden here instead of a 500?

frontend/src/Components/forms/AddUserForm.tsx
PThorpe92: Ok cool good catch there. i honestly thought i removed all instances of that method in exchange for one that returned an error but i guess i didn't merge it.

backend/src/handlers/user_handler.go
techeclecticdesign: yes, all set now.  Thanks.

calisio (commented):
This looks overall good to me (and was MUCH needed, so thank you!!). Would you be able to put the error underneath that specific input (ie under the username input) rather than at the bottom under the submit button? I think it would be helpful for the future if we are referring to several inputs. Also sorry to get back to you so late on this.

techeclecticdesign:
Added the fields and related state as you asked.  Threw in a bug fix where modals were not resetting validation errors when closing out with the closeX.

calisio:
Ah, my bad.. I wasn't specific in my last comment. I was thinking we should implement this the same way as we set an error if the user doesn't input a first or last name (since those are required). I've input a picture below: 
Screenshot 2024-08-07 at 10 45 42 AM

With your additions, it looks like this, which doesn't look the same as the other errors:
Screenshot 2024-08-07 at 10 46 28 AM

To implement the errors the same way that we do the errors I believe you'd use the setError that is part of react forms - https://react-hook-form.com/docs/useform/seterror
The example under "single field errors" should be a good one. Let me know if you have access to that/if you need any help. Sorry for not being more specific sooner!

techeclecticdesign:
No problem - I had kept the same format as the existing error at bottom of modal.  I will change the format as you described and while I am at it I will double check that all expected validation errors are caught and reported to user in the same way.  Thanks.

techeclecticdesign:
Sorry about the misunderstanding, should be good now.
There is one thing I should mention - there is no email field in the add user modal.  I could see why this may be with an inmate user base, but  I wondered if one should be added?
Since there was not an email field when adding user, but a unique constraint could be violated for email, I put the validation message in the generic error field at bottom in addUserForm.
It would be rare, but this has been tested to occur when a username is changed, and a new user is created with the other users former username, automatically assigning a new email address to one that presently exists.
If you'd like the email added, I would be happy to throw that in to this PR.

calisio:
@PThorpe92 might have more insight to the email field since I think he was the one to implement it. I think the generic error at the bottom would suffice for this since I don't believe we plan on adding a field for email. Do you have any thoughts on this Preston?

backend/src/handlers/user_handler.go
```diff
@@ -157,6 +157,17 @@ func (srv *Server) HandleCreateUser(w http.ResponseWriter, r *http.Request) {
 	if user.User.FacilityID == 0 {
 		user.User.FacilityID = srv.getFacilityID(r)
 	}
+	userNameExists := srv.Db.GetUserByUsername(user.User.Username)
```
PThorpe92, line 160: then in both the handlers we will just need

```go
if !database.IsValidUsername() {
  //...
}
```

This fixes the soft-delete issue, and prevents making several database queries and returning a whole row to make the check.

PThorpe92 (changes requested):

backend/src/handlers/user_handler.go
techeclecticdesign: Made those changes.  Thank you the tips, especially the mention about soft deletions - I learned some new things today :)

PThorpe92: Much better :+1: 

This is good with me if you are all set @calisio || @jtucholski

I will say that rebasing this is going to probably be a pain in the ass :frowning_face: depending on whether you have rebased from main once already to pull the formatting in and then made more changes in the process. Then it will be a complete nightmare and you may want to just scratch the branch and copy the changes over to a new branch you base off main, and then force push over this.

techeclecticdesign:
I can certainly see why that would be easier, and I could do the same thing for the other old PR I have pending if you would like.  More generally, anytime a bunch of commits happen in between the start and finish of a PR I submit, would you like me to do this?

techeclecticdesign:
I am having some trouble with the latest versions of the software where I can't login (releases from 3 weeks ago worked fine).  Is this a known issue, or is it more unique to my setup?

I am on Windows 10 with WSL2+Ubuntu 23.10.  I am trying to launch server on the ubuntu side.

I have tried the following:

Deleting cookies
Running make prod and make frontend-dev + yarn run dev
make migrate-fresh
using docker compose -f docker-compose.yml -f config/docker-compose.prod.yml down --volumes

PThorpe92:
> I am having some trouble with the latest versions of the software where I can't login (releases from 3 weeks ago worked fine). Is this a known issue, or is it more unique to my setup?

This issue has been happening because of how easy it is to get into a bad state when developing. There are cookies being set by two different services, and local development you have 'localhost' and '127.0.0.1' which are two different URL's according to cookies.
I recommend trying by starting over 

`docker compose down --volumes`

then `make prod` (build everything in docker)

ensure in your logs that the server is creating the SuperAdmin account in Kratos.
then go to `127.0.0.1` and try to login with `SuperAdmin` + `ChangeMe!`

Be aware that when you click the login button, it's directing you to a service that appends a `flow_id` to your query string and redirects you to our login page. Previously, when there was any kind of error, we were simply redirecting again to that service, to ensure there was a new flow, and to ensure that the user was on the correct page (127.0.0.1/login?flow=x...) with a fresh session. The service sets a CSRF cookie on the redirect, and the cookie it sets is different every time. 

Because now we removed the automatic refresh to ensure that the user has a fresh session, it is possible to get into a bad state in development that wouldn't be possible in production. Although we will be during some thorough testing to be sure of this.

PR #322, opened 2024-07-26
feat: catalogue search, search fixes, makefile

PThorpe92:


frontend/src/Pages/CourseCatalog.tsx
calisio, line 1: is ascending vs descending supposed to be working? because i dont see a difference when i click between the two in the UI

calisio:
Besides sorting in catalog everything works!

frontend/src/Pages/CourseCatalog.tsx
PThorpe92: Yeah I accidentally left it in the conditional so it only worked when you were searching. Good catch!

PR #332, opened 2024-08-03
Add facility name form for first admin login 

CK-7vn:
## What does this PR do? 

Upon the first login of the SuperAdmin account during prompt of password reset, this PR adds an input box that allows user to set the name of default facility. 

FacilityName

## What is the purpose of this PR?

Currently there is no option or way to set and or change name of the default facility, and "Default" is being used in facility names place. It is currently hard-coded to a value, and this PR allows for the name to be created upon first login so that facility name can be displayed in the admin dashboard. 

> Context: Worked on this PR with @PThorpe92, some of the changes were at his direction. 

## Result: 
DefaultNameAdded

Multiple steps were taken to ensure that this form is only presented upon first login of the default admin, and behavior has been tested by creating a new non-admin, logging in and validating that text-box was not displayed, as well as resetting SuperAdmin's password, and logging in to make sure SuperAdmin was not prompted on second password reset.


frontend/src/Components/forms/ChangePasswordForm.tsx
calisio, line 1: two super small things-
1. would you be able to move the "Valid facility name" underneath the text input- and also make sure its lowercase (not Valid Facility Name but instead Valid facility name)?
2. Would we be able to trim the spaces at the beginning/end of the input before we send it to the DB rather than telling the user it is an invalid facility name in the input? For example, if I put "Potosi Correctional " with a space at the end, instead of showing an invalid facility name, accept the input but make sure you trim it before you send it to the backend.
Screenshot 2024-08-06 at 11 50 13 AM

calisio (changes requested):
Besides those really small changes, this looks great!! Thank you for this addition; we are excited to have another contributor!

frontend/src/Components/forms/ChangePasswordForm.tsx
calisio: also @PThorpe92 feel free to push back on that second comment if you have concerns

calisio: We are still ironing out some things having to do with facility so it might change in the future, but i think for now that should be enough!

CK-7vn: Cool! I'll get those taken care of and posted shortly!

CK-7vn: Got that all taken care of, if it looks weird, or if you were looking for something different, don't hesitate to let me know! Here's a couple pictures for reference. The facility name seen on the dashboard page was submitted with the extensive amount of white space at the beginning of the facility in the change password form picture below, for reference. 
LoweredValidFacilityName
ProfilePageAfterTrimName

calisio (approved):
Looks great! Thank you for making those changes 👍

calisio:
Ah, so the commit message on your last push didn't use a conventional commit. Would you be able to go back and amend that commit message so it passes the check? If you need any help doing that, let us know.

CK-7vn:
> Ah, so the commit message on your last push didn't use a conventional commit. Would you be able to go back and amend that commit message so it passes the check? If you need any help doing that, let us know.

Yes, I know, I totally blanked it and then realized like right as you approved it! SMH. Sorry!

CK-7vn:
I'm still new to conventional commits, but I believe it's all set now. Once again, sorry for that. Still working on getting used to the verbiage, but shouldn't happen again! Appreciate the correction.

PR #337, opened 2024-08-09
Complete Oauth2/OIDC implementation, auto grant scope for session renewal and fix consent page

PThorpe92:
Will add some comments throughout explaining some of what is going on/reasons for changes.

It's on the main branch of my fork, which builds + pushes containers in CI, so this PR can be viewed at staging.

### DIFF
--- a/backend/src/handlers/auth.go
+++ b/backend/src/handlers/auth.go
@@ -3,8 +3,6 @@ package handlers
 import (
 	"UnlockEdv2/src/models"
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"net/http"
@@ -19,17 +17,18 @@ import (
 	"golang.org/x/crypto/bcrypt"
 )
 
-type contextKey string
-
 const ClaimsKey contextKey = "claims"
 
-type Claims struct {
-	UserID        uint            `json:"user_id"`
-	PasswordReset bool            `json:"password_reset"`
-	Role          models.UserRole `json:"role"`
-	FacilityID    uint            `json:"facility_id"`
-	jwt.RegisteredClaims
-}
+type (
+	contextKey string
+	Claims     struct {
+		UserID        uint            `json:"user_id"`
+		PasswordReset bool            `json:"password_reset"`
+		Role          models.UserRole `json:"role"`
+		FacilityID    uint            `json:"facility_id"`
+		jwt.RegisteredClaims
+	}
+)
 
 func (srv *Server) registerAuthRoutes() {
 	srv.Mux.Handle("POST /api/reset-password", srv.applyMiddleware(srv.handleResetPassword))
@@ -55,7 +54,6 @@ func (s *Server) AuthMiddleware(next http.Handler) http.HandlerFunc {
 			return
 		}
 		if claims, ok := token.Claims.(*Claims); ok && token.Valid {
-
 			ctx := context.WithValue(r.Context(), ClaimsKey, claims)
 			if claims.PasswordReset && !isAuthRoute(r) {
 				http.Redirect(w, r.WithContext(ctx), "/reset-password", http.StatusOK)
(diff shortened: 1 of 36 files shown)


backend/src/handlers/auth.go
```diff
@@ -128,21 +126,6 @@ func (srv *Server) adminMiddleware(next func(http.ResponseWriter, *http.Request)
 	})
 }
 
-type OrySession struct {
-	Active    bool
-	TokenHash string
-	Expires   time.Time
-}
-
-func hashCookieValue(cookieValue string) string {
```
PThorpe92, line 137: All of this was removed in favor of pulling the kratos session cookie at the Auth endpoint. Our Identity/Auth services will always be hosted at the same top level domain, so we can take full advantage of all the CSRF + Cookie protections. Generally will be avoiding caching things in memory unless needed. NATS (cache + MQ) should fix that soon

backend/src/models/users.go
```diff
@@ -56,7 +55,7 @@ func (User) TableName() string {
 
 func (usr *User) BeforeCreate(tx *gorm.DB) error {
 	if usr.Email == "" {
-		usr.Email = usr.NameLast + "." + usr.NameFirst + "@unlocked.v2"
+		usr.Email = usr.Username + "@unlocked.v2"
```
PThorpe92, line 58: right now we have to use the `preferred_username` field of the oauth id_token (which is what is sent to the client/canvas to identify the user), because the `sub` field (default) will be set to the users `kratos_id` instead of the username.

It's possible we may have to either use our `{username}@unlocked.v2` as the de-facto external identifier for our users in providers, or duplicate the username field in the email? or use the users kratos_id (uuid) as the SIS id and log them in through their SIS id? not sure yet

backend/tests/test_data/facilities.json
```diff
@@ -1,32 +1,27 @@
...
-        "updated_at": "2024-06-27T19:15:00.123456789Z",
-        "name":      "Farmington Correctional Center"
-    },
-    {
-        "id":        6,
-        "created_at": "2024-06-27T19:15:00.123456789Z",
-        "updated_at": "2024-06-27T19:15:00.123456789Z",
-        "name":      "Algoa Correctional Center"
-    }
+  {
```
PThorpe92, line 2: removed the `id` fields of all the stored `.json` items, as not to screw up adding new items to a database that has been seeded. currently will throw a duplicate primary key error when you create a new program or user in a database that has been seeded with test data

frontend/src/Components/CatalogCourseCard.tsx
```diff
@@ -92,67 +92,71 @@ export default function CatalogCourseCard({
...
-                    <p className="text-xs">{course.provider_name}</p>
-                    <h3 className="card-title text-sm">
-                        {course.program_name}
-                    </h3>
-                    <p className="body-small line-clamp-2">
-                        {course.description}
-                    </p>
-                    <div className="flex flex-wrap py-1 mt-2 space-y-2">
-                        {programPill} {outcomePills}
+        <>
```
PThorpe92, line 96: I got stuck rebasing a massive, massive change (all the formatting) when i rebased from main. So some stuff was confusing. Let me know if this was the wrong change @calisio

frontend/src/Components/forms/LoginForm.tsx
```diff
@@ -1,93 +1,137 @@
...
-            }
-            const cookie = flow_response.data.ui.nodes[0].attributes.value;
-            kratosData['csrf_token'] = cookie;
-            const response = await axios(url, {
-                method: 'POST',
-                data: kratosData,
-                headers: {
-                    'Content-Type': 'application/json',
-                    'X-CSRF-Token': cookie
+            console.log('Submitting login request', reqBody);
```
PThorpe92, line 38: going to leave these console.logs here until I can get the time to walk either carolina or josh through this entire flow. Not important yet, especially as there are a handlful of other console.logs that need to be removed client side

frontend/src/Components/CatalogCourseCard.tsx
calisio: shoot, ya sorry about having to rebase that 😬 This should have kept the incoming change- the one that separates out the returns (ie the one that has the`if (view == ViewType.List) return (...)` rather than the ternary operator. It was a linting error that I fixed.

PThorpe92: Ok I kinda figured that was the case, no worries I will fix it

Issue #323, opened 2024-08-02
Bug?: When Admin attempts to delete a user and try re-adding the SAME user with the same username it is unsuccessful. Add User modal stays up and Submit button doesn’t work.


jtucholski:
There is no visible error message as to why the system cannot create a new user. 

In this case, the username was taken (it was created and then soft-deleted by an admin). The admin can use the front end to create a new user, but the subsequent API request to `/api/users` generates a 500 response with the following error message.

> ERROR: duplicate key value violates unique constraint "uni_users_username" (SQLSTATE 23505)

This scenario will likely occur if we create a new user with the same username as a deleted user. When that happens, we should provide a validation error message informing the user performing the action that the username is unavailable.

techeclecticdesign:
I just noticed you posted this issue.  I believe this has been addressed with PR #316.  You're welcome to improve upon it or create a better version, I just wanted to make sure you knew about the PR if this was something you had started working on.

jtucholski:
Thanks @techeclecticdesign. I appreciate you working on this. We discussed this in the context of some auth changes that @PThorpe92 is making in #337. He might ask you to consider some changes.

PR #342, opened 2024-08-13
fix: Validation error when creating preexisting username revised

techeclecticdesign:
Thank you everyone for your support with this commit.
Also thank you Preston, for showing me that all I had to do was use 127.0.0.1 instead of localhost.  I can't believe how easy that was.

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -167,6 +167,16 @@ func (db *DB) GetUserByUsername(username string) *models.User {
 	return &user
 }
 
+func (db *DB) UsernameExists(username string) bool {
+	userExists := false
+	err := db.Conn.Raw("SELECT EXISTS(SELECT 1 FROM users WHERE username = ?)", username).
+		Scan(&userExists)
+	if err != nil {
+		log.Error("Error checking if username exists: ", err)
+	}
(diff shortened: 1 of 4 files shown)


calisio (approved):
This looks good to me. Will wait for one more approval from @PThorpe92 or @jtucholski. Also, before we merge, would you be able to amend your commit message for it to pass the conventional commit?

backend/src/database/users.go
```diff
@@ -167,6 +167,16 @@ func (db *DB) GetUserByUsername(username string) *models.User {
 	return &user
 }
 
+func (db *DB) IsValidUsername(username string) bool {
```
jtucholski, line 170: Any chance we can rename this? IsValidUsername implies a check to make sure the name doesn't violate any rules (eg. no spaces) whereas this function checks to see if the username has already been taken.

techeclecticdesign: Good catch I wasn't using my head there.

jtucholski:
Great, appreciate all of the changes!

jtucholski (approved):

jtucholski:
This should close Issue #323 when it's merged.

Issue #268, opened 2024-06-21
My progress search/sort functionality


calisio:
# Overview

The student view of “my progress” does not have a sort functionality implemented yet. Below is a view of what the UI should look like. The two sorts that must be added are for courses and for certifications.
<img width="1512" alt="my progress search sort" src="https://github.com/UnlockedLabs/UnlockEdv2/assets/89868388/658fb783-d7b9-487f-8202-88f9b11bf100">

# Relevant files

These should live in the `MyCourses.tsx` file. The handler is in `/backend/src/handlers/dashboard.go` and the call to the database is in `/backend/src/database/dashboard.go` for “all courses” table. The handler for “certificates” is in `/backend/src/handlers/outcomes_handler.go` and call to the database is in `/backend/src/database/outcomes.go`. 

# Sorting

Users should be able to sort/filter by the following in courses:

- alphabetically
- status (completed or in progress)
- grade
- hours spent
- any other you see relevant

Users should be able so sort by the following in certificates:

- date
- alphabetically
- any other you see relevant

# Tasks to complete ticket

- [ ] Update backend api call and table join to query for filtering of courses and certificates
- [ ] Update frontend to include filter dropdown for courses and certificates
- [ ] Connect frontend to backend for filter

jtucholski:
Closed as this was fixed in in PR #348

PR #348, opened 2024-08-13
feat(268): add sort to MyProgress.tsx revised

techeclecticdesign:


frontend/src/Pages/MyProgress.tsx
```diff
@@ -48,9 +75,20 @@ export default function MyProgress() {
                         </div>
                         <div className="flex flex-row gap-12 mt-12">
                             <div className="card bg-base-teal h-[531px] w-[60%] p-4 overflow-y-auto">
-                                <div className="flex flex-row">
-                                    <h2>All Courses</h2>
-                                    {/* dropdown will go here */}
+                                <div className="flex flex-row gap-x-4">
```
calisio, line 78: Instead of using `gap-x-4` for the row with the dropdowns, would you be able to do `justify-between`? It just puts them on opposite sides which I think looks slightly better. Same with the certificates dropdown

calisio (changes requested):
Looks overall good, approved once its fixed!

PThorpe92:
fixed + rebased so you don't have to worry about it. Thanks for the PR! :+1:

PThorpe92 (approved):

jtucholski:
@calisio @PThorpe92 This should be linked to #268, right? I noticed the merge didn't close that issue.

Issue #334, opened 2024-08-07
BUG: Inform user that username cannot have certain characters


calisio:
**Describe the bug**
If a user creates a username that is invalid because it includes certain characters (such as spaces), it accepts it and strips them in the backend. This should not be accepted, but instead should prompt the user to choose a new username without these characters.

**To Reproduce**
1. Go to the Users page.
2. Click on the "Add User" button
3. Type in a username that includes some spaces or any other invalid character (ie "carolina     alisio")
4. the user will successfully be added to the database under the username "carolinaalisio" - the backend strips invalid characters and saves them in the database as such.

**Expected behavior**
Rather than allowing the username to be accepted, we should prompt the user to input a valid username. this error should be set using react hook form's `setError` (see https://github.com/UnlockedLabs/UnlockEdv2/pull/316#issuecomment-2273805693 for more info on what this should look like).

**Screenshots**
Screenshots are attached in [this github comment](https://github.com/UnlockedLabs/UnlockEdv2/pull/316#issuecomment-2273805693).

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser chrome
 - Version 126.0.6478.127

techeclecticdesign:
Since this is similar to what I just did I will happily take this on, thanks.

techeclecticdesign:
@PThorpe92 ,
Did you still want the other validation checks following up on #349?

PThorpe92:
I actually think all of those were added in a very large, breaking change introducing refactor that we did recently.

PR #349, opened 2024-08-14
feat(334): validation support in TextInput

techeclecticdesign:
Instead of using setError I used the built in validation support via register by adding an optional field to TextInput.
I wanted to get opinion on this, I thought it would be more versatile.  I would be happy to set this up with setError instead if that is your preference.
I like that it live auto-updates as you type instead of waiting for submit.

### DIFF
--- a/frontend/src/Components/forms/AddUserForm.tsx
+++ b/frontend/src/Components/forms/AddUserForm.tsx
@@ -105,6 +105,13 @@ export default function AddUserForm({
                     length={50}
                     errors={errors}
                     register={register}
+                    validationRules={{
+                        pattern: {
+                            value: /^[a-zA-Z0-9]+$/,
+                            message:
+                                'Username must only contain letters and numbers'
+                        }
+                    }}
                 />
                 <DropdownInput
                     label={'Role'}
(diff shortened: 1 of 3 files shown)


PThorpe92 (approved):
Looks good to me :+1: 

thanks for this

jtucholski:
Hey @techeclecticdesign

> I like that it live auto-updates as you type instead of waiting for submit.

I'm seeing the new error message, but I'm not seeing any validation as I type. I see it when I press Submit. Is there any chance I'm missing something?

https://github.com/user-attachments/assets/77857c03-0b8d-4178-a885-daee79cbba3f

techeclecticdesign:
Hi @jtucholski,
Playing with it a little I am seeing the live updating only happens after the first bad submission.
It seems this is the default behavior with react-hook-forms.

calisio:
You might be able to use `watch` for real time validation? Not 100% sure but from briefly looking at it it might provide a solution. https://react-hook-form.com/docs/useform/watch

PThorpe92:
Hey so I see that this was meant to take care of #334

I previously thought it was only in the scope of adding the generic form validation for the text input.

If it is going to clear up that ticket, can we add validation on the backend? I have been meaning to add this library to handle all of our forms. https://github.com/go-playground/validator

techeclecticdesign:
Sorry for the delay in reply, I didn't have access to a computer for a few days.
So to clarify, instead of the built-in frontend validation with react-hook-forms, I should handle validation on the backend after the form has been submitted with the validator library.  Sound good?

PR #352, opened 2024-08-15
fix: fixes mapping modal, updates pagination, updates seeder

### LINKED ISSUE #331: modal for mapping user should be updated

calisio:
Fixes the user management mapping modal. This updates the UI for more clarity and uses backend pagination rather than doing it in the frontend. It renders different views for when a user is fuzzy matched vs not, clears up naming conventions, and provides the user with more information on what each page means. Functionality should be the same as it was before. Screenshots below for reference:

### Previous Modal
Screenshot 2024-08-15 at 11 36 50 AM
Screenshot 2024-08-15 at 11 37 01 AM

 ### Updated Modal
If the user has no fuzzy matches, will show all users.
Screenshot 2024-08-15 at 11 17 56 AM

If the user has a fuzzy match, it will show that match.
Screenshot 2024-08-15 at 11 17 39 AM

On selection of a user the map student button is enabled.
Screenshot 2024-08-15 at 11 17 43 AM

If they click see all users, it will show all users along with a link to display the fuzzy users again.
Screenshot 2024-08-15 at 11 17 49 AM

Closes #331.

### DIFF
--- a/backend/src/database/provider_user_mappings.go
+++ b/backend/src/database/provider_user_mappings.go
@@ -106,14 +106,6 @@ func (db *DB) getUnmappedProviderUsersWithSearch(providerID string, userSearch [
 	if err := tx.Find(&users).Error; err != nil {
 		return nil, err
 	}
-	if len(users) == 0 {
-		fmt.Println("couldn't find any good searches, returning all")
-		return users, db.Conn.Table("users u").Select("u.*").
-			Where("u.role = ?", "student").
-			Where("u.id NOT IN (SELECT user_id FROM provider_user_mappings WHERE provider_platform_id = ?)", providerID).
-			Where("facility_id = ?", fmt.Sprintf("%d", facilityId)).
-			Find(&users).Error
-	}
 
 	fmt.Printf("found %d matches", len(users))
 	return users, nil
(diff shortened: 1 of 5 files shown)


backend/src/database/provider_user_mappings.go
```diff
@@ -106,14 +106,6 @@ func (db *DB) getUnmappedProviderUsersWithSearch(providerID string, userSearch [
...
 		return nil, err
 	}
-	if len(users) == 0 {
-		fmt.Println("couldn't find any good searches, returning all")
-		return users, db.Conn.Table("users u").Select("u.*").
-			Where("u.role = ?", "student").
-			Where("u.id NOT IN (SELECT user_id FROM provider_user_mappings WHERE provider_platform_id = ?)", providerID).
-			Where("facility_id = ?", fmt.Sprintf("%d", facilityId)).
-			Find(&users).Error
-	}
```
calisio, line 116: Removed this because if there is no fuzzy match, should return an empty array. All users is already returned in a different call.

backend/tests/test_data/users.json
calisio, line 1: Just updated the seeder so there would be more students in the facility. Facility 1 is the only one relevant for now anyways. Could change this back if needed.

PThorpe92:
looks sooooo much better :grin:

jtucholski (approved):

PR #355, opened 2024-08-19
feat: add unauthorized page and protect routes

### LINKED ISSUE #344: bug: if user goes directly to https://staging.unlockedlabs.xyz/{any path} and they are not logged in, they should be redirected
**Describe the bug**
If the user goes directly to any path in our app, a 401 unauthorized error is returned and the page is blank. This should either redirect them or should show an error page that gives them the opportunity to redirect to the login page.

**To Reproduce**
1. Go to 'https://staging.unlockedlabs.xyz/dashboard' without being logged in to any account.
2. See blank page with no error - can see the 401 error in the console

**Expected behavior**
We need to show some sort of screen or redirect the user to the login page.

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser Chrome

**Additional context**
If we plan to create a new page for this, it should be sketched out on Figma first, or spoken about what the UI should look like. A discussion could be had under this issue.

PThorpe92:
When a user is not logged in and tries to access any page that requires authentication, they will be redirected to the login page.

When a student (non-admin) is logged in, and tries to access a page that requires administrative privileges, they will be presented with the following page:
unauthorized

When any logged in user attempts to access a page that doesn't exist, they will be presented with: 

not_found

fixes #344

### DIFF
--- a/backend/src/handlers/actions.go
+++ b/backend/src/handlers/actions.go
@@ -13,12 +13,12 @@ import (
 
 func (srv *Server) registerActionsRoutes() {
 	// returns the users for mapping on the client
-	srv.Mux.Handle("GET /api/actions/provider-platforms/{id}/get-users", srv.applyMiddleware(srv.HandleGetUsers))
+	srv.Mux.Handle("GET /api/actions/provider-platforms/{id}/get-users", srv.ApplyAdminMiddleware(srv.HandleGetUsers))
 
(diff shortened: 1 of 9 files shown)


jtucholski (approved):
I added two calls to logWithFields in the activity_handler. Looks good.

Front-end was tested with:
- Student auth accessing admin pages (saw the unauthorized screen)
- Logged out user (saw redirect to login screen)
- Logged in user accessing invalid URLs (saw the not found screen)

PR #359, opened 2024-08-23
added responsive design to pills

MillerSenior:
[Screen Shot 2024-08-23 at 2.23.28 PM.pdf](https://github.com/user-attachments/files/16732979/Screen.Shot.2024-08-23.at.2.23.28.PM.pdf)
[Screen Shot 2024-08-23 at 2.24.03 PM.pdf](https://github.com/user-attachments/files/16732982/Screen.Shot.2024-08-23.at.2.24.03.PM.pdf)

### DIFF
--- a/frontend/src/Components/CatalogCourseCard.tsx
+++ b/frontend/src/Components/CatalogCourseCard.tsx
@@ -148,7 +148,10 @@ export default function CatalogCourseCard({
                         <p className="body-small line-clamp-2">
                             {course.description}
                         </p>
-                        <div className="flex flex-wrap py-1 mt-2 space-y-2">
+                        {/*<div className="flex flex-col sm:flex-row flex-wrap py-1 mt-2 gap-2">*/}
+                        {/*    {programPill}  {outcomePills}*/}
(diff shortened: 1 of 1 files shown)


jtucholski:
@MillerSenior Rather than making the pills take up the container width, can you keep their width limited to the text of the pill?

What if you tried getting rid of space-y-2? Or applied it for smaller screen sizes but blocked it from larger screen sizes? Like this (screenshot when it is properly applied and not at the same time, faked using Inspector).

image

frontend/src/Components/CatalogCourseCard.tsx
jtucholski, line 1: Try seeing if you can use `space-y-2` for smaller screens and removing it for larger screens (whenever the pills stack). This might be the only change that's necessary. No need to change all the other pills to `inline-flex` or `items-center`.

jtucholski (changes requested):

MillerSenior:
Updated CatalogCourseCard.tsx, closes #359

jtucholski (approved):

PR #361, opened 2024-08-26
feat(334): backend validation for frontend forms

techeclecticdesign:
## Description of the change

Implement legal character username validation on the backend using validator v10.  

Closes 334

## Additional context

Should I apply this pattern to the rest of the form validation on adduser/edituser?  Eg. max, required, etc?

Also a minor aside, I noticed response to axios put was always 200 even when server should have been returning an error.  So edituserform and adduserform use slightly different logic from each other until I can figure out whats going on.


backend/go.mod
techeclecticdesign, line 1: Go seems to have changed go.mod quite a lot by me adding a new library.  Is this normal, am i doing something wrong here?

PThorpe92: try running `go mod tidy` and see what the output is. I can already see it added some unnecessary libs and bumped `gorm` down a version which we definitely don't want to do if possible.

backend/src/handlers/user_handler.go
```diff
@@ -324,3 +335,14 @@ func (srv *Server) HandleResetStudentPassword(w http.ResponseWriter, r *http.Req
 	}
 	srv.WriteResponse(w, http.StatusOK, response)
 }
+
+func validateUsername(username string) string {
```
PThorpe92, line 339: I was thinking we could tag the structs, and then be able to validate by calling `ValidateStruct()`
So something like
```go
// static variable in database/DB.go
var validate *validator.Validate


// models/users.go
type User struct {
DbFields
Username string `json:"username" validate:"alphanum"`
Email string `json:"email" validate:"email;omitempty"`
/...

err := validator.ValidateStruct(user)
```

I can make a ticket for this if you'd like

techeclecticdesign: I'm happy to do it, so if you think a ticket would be tidier than adding extra to this PR that'd be good.  Otherwise I can just add it to this PR.

backend/go.mod
techeclecticdesign: running go mod tidy did change the go.mod file but it did bring gorm back up to 1.25.11, nor did it get rid of the extra libs.

backend/src/handlers/user_handler.go
PThorpe92: How about add validation to just the user in this PR in that manor, and I'll get a ticket made up for the rest?

techeclecticdesign:
i updated to use struct for just username.
I got an error on test.  Should I fix test so that it accepts a user not being included in response, or should i change the response when invalid input is provided so that a user struct is still included in response.data?

backend/go.mod
PThorpe92: maybe try manually changing back at least the gorm version, everything else looks harmless.

techeclecticdesign: Cool.  I had thought go would change it back but it stuck just fine 👍

PThorpe92 (approved):
I am good with this if you are @jtucholski

PR #401, opened 2024-09-30
Kiwix db schema

### LINKED ISSUE #398: Kiwix Management: Database Schema & Library Population
**Description** 
* Create the `libraries` table in Postgres that will be used to persist Kiwix Open Content Libraries/Books.
* Build out a seed script to populate the `libraries` table with five Kiwix libraries. Use the [library feed](https://library.kiwix.org/catalog/v2/entries?lang=eng) from Kiwix to select five libraries. 

**Tasks**
- [x] Define the database schema for the `libraries` table
- [x] Implement the migrations to create the `libraries` table
- [x] Create the Go models for the `libraries` table
- [x] Write seeder scripts to populate the `libraries` table for development purposes

**Table structure**
Fields: (exact name can change if conventions require it, refer to similar database columns for types)
* `id` -- SERIAL internal identifier
* `open_content_provider_id` -- INT FK field to the `open_content_provider` table
* `external_id` -- VARCHAR? this field holds the external content provider's id
* `title` -- VARCHAR the name of the library
* `language` -- VARCHAR
* `description` -- TEXT description text for the library
* `href/url` -- VARCHAR this field holds the url that the user navigates to view the library
* `imageUrl` -- VARCHAR this field holds the url for the image
* `visibility_status` -- BOOLEAN this field reflects it this library is visible to users
* `created_at`
* `updated_at`
* `deleted_at`

**Acceptance criteria**
- The `libraries` table exists with the schema needed to support libraries.
- A row in the `open_content_provider` table exists for Kiwix.
- The seed script includes 5 libraries that are imported into the `libraries` table during `make seed`

[Link to Design in Notion](https://www.notion.so/unlockedlabs/Open-Content-Management-Design-5dcf69031dde4c909bd6528674346bc9?pvs=4)

calisio:
## Description of the change
- two migrations scripts to create a library table and update open content provider table
- creates library model
- **Related issues**: Closes #398

### DIFF
--- a/backend/migrations/00006_add_libraries_tables.sql
+++ b/backend/migrations/00006_add_libraries_tables.sql
@@ -1,0 +1,26 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE public.libraries (
+    id SERIAL PRIMARY KEY,
+    open_content_provider_id integer NOT NULL,
+    external_id VARCHAR,
+    name VARCHAR(255) NOT NULL,
+    language VARCHAR(255),
+    description TEXT,
+    url VARCHAR NOT NULL,
+    image_url VARCHAR,
+    visibility_status BOOLEAN NOT NULL DEFAULT false,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    deleted_at timestamp with time zone,
+    FOREIGN KEY (open_content_provider_id) REFERENCES open_content_providers(id) ON UPDATE CASCADE ON DELETE CASCADE
+);
+CREATE INDEX idx_libraries_deleted_at ON public.libraries USING btree (id);
+CREATE INDEX idx_libraries_open_content_provider_id ON public.libraries USING btree (open_content_provider_id);
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE public.libraries CASCADE;
+-- +goose StatementEnd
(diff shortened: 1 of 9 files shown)


backend/src/models/model_helpers.go
```diff
@@ -0,0 +1,9 @@
+package models
+
+func StringPtr(s string) *string {
```
PThorpe92, line 3: We have the `UpdateStruct` function in `backend/src/models/resource.go`

Could you move that into this file? that func is also used in many other places.

calisio: Sounds good, just pushed that change

backend/migrations/00006_add_libraries_tables.sql
```diff
@@ -0,0 +1,26 @@
...
+    image_url VARCHAR,
+    visibility_status BOOLEAN NOT NULL DEFAULT false,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    deleted_at timestamp with time zone,
+    FOREIGN KEY (open_content_provider_id) REFERENCES open_content_providers(id) ON UPDATE CASCADE ON DELETE CASCADE
+);
+CREATE INDEX idx_libraries_deleted_at ON public.libraries USING btree (id);
+CREATE INDEX idx_libraries_open_content_provider_id ON public.libraries USING btree (open_content_provider_id);
+
```
PThorpe92, line 20: can we make indexes on the URL and external_id here too? Only cuz its a migration and we cannot edit it later on without making a new one. then this is good to go :+1:

jtucholski: @PThorpe92 These have been added.

PThorpe92 (approved):

PR #402, opened 2024-10-01
HeroIcon refactor - ResourceManagement.tsx

corypride:
## Description of the change

## Screenshot(s)

## Additional context

### DIFF
--- a/frontend/src/Components/ULIComponent.tsx
+++ b/frontend/src/Components/ULIComponent.tsx
@@ -1,0 +1,21 @@
+import * as React from 'react';
+
+type ULIComponentProps = {
+    className: string;
+    tooltip: string;
+    onClick: () => void;
+    icon: React.ForwardRefExoticComponent<
+        React.PropsWithoutRef<React.SVGProps<SVGSVGElement>> & {
+            title?: string;
+            titleId?: string;
+        } & React.RefAttributes<SVGSVGElement>
+    >;
+};
+
+export default function ULIComponent(props: ULIComponentProps) {
+    return (
+        <div className="tooltip" data-tip={props.tooltip}>
+            <props.icon className={props.className} onClick={props.onClick} />
+        </div>
+    );
+}
--- a/frontend/src/Pages/ResourcesManagement.tsx
+++ b/frontend/src/Pages/ResourcesManagement.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import axios from 'axios';
 import useSWR from 'swr';
 import Modal, { ModalType } from '../Components/Modal';
@@ -15,6 +15,7 @@ import { CloudArrowUpIcon } from '@heroicons/react/16/solid';
 import DeleteForm from '../Components/forms/DeleteForm';
 import { useDebounceValue } from 'usehooks-ts';
 import ExternalLink from '@/Components/ExternalLink';
+import ULIComponent from '@/Components/ULIComponent.tsx';
 
 interface ToastProps {
     state: ToastState;
@@ -141,7 +142,7 @@ export default function ResourcesManagement() {
         setCollectionList(newCollections);
     };
 
-    const updateFinalState = async (e: any) => {
+    const updateFinalState = async (e: React.MouseEvent) => {
         setToast({ state: ToastState.null, message: '' });
         e.preventDefault();
         const newCollectionList = collectionList.map((c, i) => {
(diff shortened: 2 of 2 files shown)


corypride:
@jtucholski what about this one.... it does not fit the shape of the rest of the Icon components on the ResourcesManagement.tsx...... 
image

calisio:
Hey @corypride, briefly looking over this before the weekend - do you have any other changes that aren't yet pushed? Just double checking because I believe there are some other files that should also be updated with this PR (like `ProviderPlatformManagement.tsx`, `StudentManagement.tsx`, `AdminManagement.tsx`-- maybe check in with Anthony regarding student and admin management as he is doing a refactor of that page already in PR #404 ). Don't worry about `ProviderUserManagement.tsx` for the moment - that page will have a refactor so will be updated later.

corypride:
Ok. That is a misunderstanding in my part. I did not do the other two file…
student and admin. On it now.

jtucholski:
@corypride Sorry, I missed this notification and didn't reply. The approach you took on Resource Management works well. 

> @jtucholski what about this one.... it does not fit the shape of the rest of the Icon components on the ResourcesManagement.tsx...... image

frontend/src/Components/ULIComponent.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+type ULIComponentProps = {
+    outerDivClassName?: string; //for applying classes other than 'tooltip'
+    innerDivClassName?: string; //for applying classes other than 'w-4 h-4'
```
jtucholski, line 5: Good thinking. Are there examples of other classes we need to apply to the outer div besides tooltip? 

Also, do we have another name besides `outerDiv` and `innerDiv`? I'm stuck on naming it but I think it should convey what it's doing and not what element it applies to (in case the element changes later on).

corypride: This was one example, but  Carolina mentioned in the discussion of the refactor where there would be cases where there would be other classes applied to the component... specfically mentioning a case where there would be a need to override default width on the logo.

image

@jtucholski I am always open for name suggestions.... always have the most difficult time with coming up something meaingful... :)

corypride:
@calisio

frontend/src/Pages/ResourcesManagement.tsx
calisio, line 1: Hey @corypride, I think there might be a formatting thing that was accidentally changed in this file? The first screenshot shows this version, the second shows staging (how its supposed to look). Do you mind checking this out? I think you'll just need to add a div around the two icons so it groups them together and a gap between both of them that is 8px (so space-x-2 should do the trick).
Screenshot 2024-10-08 at 10 30 54 AM
Screenshot 2024-10-08 at 10 31 48 AM

frontend/src/Components/ProviderCard.tsx
```diff
@@ -53,64 +54,50 @@ export default function ProviderCard({
...
+                        <ULIComponent
+                            dataTip={'Edit Provider'}
+                            icon={PencilSquareIcon}
+                            onClick={() => openEditProvider(provider)}
+                        />
+                        <ULIComponent
+                            data-tip={'Disable/Archive Provider'}
+                            icon={XMarkIcon}
                             onClick={() => archiveProvider(provider)}
                         />
```
calisio, line 93: Would you be able to check on your local if this gives a tooltip for you? On my local it doesn't and I'm not sure why.. it could be an issue only I'm having?

frontend/src/Components/ULIComponent.tsx
jtucholski: How about we call it `toolTipClassName` and `iconClassName`?

frontend/src/Components/ProviderCard.tsx
corypride: image

corypride: That is the X on the Canvas card

calisio: No I mean if you hover over the x does it give a tooltip for you?

corypride: Great catch!!! data-tip= instead of dataTip=

frontend/src/Components/ProviderCard.tsx
```diff
@@ -53,64 +54,50 @@ export default function ProviderCard({
...
+                        <ULIComponent
+                            dataTip={'Edit Provider'}
+                            icon={PencilSquareIcon}
+                            onClick={() => openEditProvider(provider)}
+                        />
+                        <ULIComponent
+                            dataTip={'Disable/Archive Provider'}
+                            icon={XMarkIcon}
                             onClick={() => archiveProvider(provider)}
                         />
```
calisio, line 93: sorry last thing- because this cuts off the tooltip, could we add `tooltipClassName="tooltip-left"
` onto here so you can fully read what it says? Once that is complete it should all be good to go & I'll approve and merge

corypride:
Can we shorten the tooltip? Or do u want me to alter the css

calisio (approved):
Looks great, thank you Cory!

PR #437, opened 2024-10-15
Admin headers

### LINKED ISSUE #427: BUG: Update all Admin pages to have header and same padding
Each page on the admin portal has a different set of classes/padding. Some have headers some don't.

**Tasks**
- [ ] Ensure that each page has a header with a padding of `px-8` to align with the top breadcrumb.

**Attached screenshots**
Image
Image

MillerSenior:
Updated all admin pages to have headers and same padding, closes #427
Screenshot 2024-10-15 143936

### DIFF
--- a/frontend/src/Pages/AdminDashboard.tsx
+++ b/frontend/src/Pages/AdminDashboard.tsx
@@ -22,8 +22,10 @@ export default function AdminDashboard() {
     const totalActivity = convertSeconds(activityData.total_weekly_activity);
 
     return (
-        <div className="px-8 py-4">
-            <h1 className="text-5xl">{activityData.facility_name}</h1>
+        <div className="flex flex-col space-y-6 overflow-x-auto rounded-lg p-4">
+            <div className="pt-8">
+                <h1>{activityData.facility_name}</h1>
+            </div>
             <div className="flex flex-row mt-12 gap-12">
                 <div className="flex flex-col gap-6">
                     <div className="card h-[240px]">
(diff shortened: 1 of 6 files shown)


frontend/src/Pages/AdminDashboard.tsx
```diff
@@ -22,8 +22,10 @@ export default function AdminDashboard() {
     const totalActivity = convertSeconds(activityData.total_weekly_activity);
 
     return (
-        <div className="px-8 py-4">
-            <h1 className="text-5xl">{activityData.facility_name}</h1>
+        <div className="flex flex-col space-y-6 overflow-x-auto rounded-lg p-4">
+            <div className="pt-8">
+                <h1>{activityData.facility_name}</h1>
+            </div>
```
calisio, line 28: This is actually the only one that should stay as-is! The facility name is larger on the designs so i think we should just maybe keep it that way?

calisio (changes requested):
You shouldn't need to add these extra divs with `pt-8` on the headers. I believe the ticket asked for making the adjustment to `px-8`, not `pt-8`. px will alter the padding along the x-axis, whereas pt will alter the padding to the top of the header. 

Provider Platforms (in its original state) should be an example of what each page should look like. 
Screenshot 2024-10-16 at 2 56 57 PM
If you can see on this page, the header is aligned with the house icon on the top of the page. On the right hand side, the button and table should be aligned with the name dropdown. It might be easier to just add the padding to the div that wraps both the `<PageNav>` and the element that passes along `{children}` within `AuthenticatedLayout.tsx`, that way we dont need to worry about it in all the pages. If you choose to go this route, make sure to check all pages for the correct padding (for example, in all of the student pages you would have to go back and take away `px-8 py-4` in the first div that is after `AuthenticatedLayout.tsx`. This approach would really help to reduce redundancy and make sure all pages will always have the same padding. 

It's a bit hard to write out, so if you have any questions we can hop on a call and walk through it.

PR #438, opened 2024-10-15
fix: removed grade from ui

### LINKED ISSUE #415: Remove "Grade" column from "All Courses" table in `MyProgress.tsx`
The grade column was added in the original Harvard designs, but because this is not a value that is returned by our database currently. The entire column should be removed. 

An image of what it currently looks like is attached below:

MillerSenior:
Removed grade column from "All Courses" table in MyProgress.tsx closes #415

Screenshot 2024-10-15 150515


calisio (commented):
Looks good! Would you be able to delete them rather than commenting them out? Then we can merge 😎

MillerSenior:
Removed commented code

calisio (approved):

PR #436, opened 2024-10-15
Fixes Build System for the Frontend.

Huijiro:
## Description of the change

This PR fixes the build system for the frontend that was setup improperly leading to typescript to not be evaluated.

This entails:
- Fixing the TSConfig setup to properly handle all files.
- Fixed TSC not compiling with Typescript checks.
- Updated ESLint to version 9, and migrated config files to new version.
- Added new type checks on ESLint for things like, async errors, unused variables, undefined checks and such.

## Additional context

This was done with help of @PThorpe92, if anything needs to be discussed please lead the conversation with him.

PR #448, opened 2024-10-17
Fixes issue with side navigation not coming back after hidden

### LINKED ISSUE #447: BUG: Issue with Collapsible Nav Bar after its hidden
There's an issue where the collapsible navigation bar cannot be reopened after it is closed. This has been reproduced both in staging and in local dev.

jtucholski:
## Description of the change

This change fixes #447. A user could hide the side navigation but once it was hidden, they couldn't click the hamburger bar icon to bring it back.

### DIFF
--- a/frontend/src/Components/PageNav.tsx
+++ b/frontend/src/Components/PageNav.tsx
@@ -46,9 +46,7 @@ export default function PageNav({
                     {showOpenMenu ? (
                         <li>
                             <ULIComponent
-                                onClick={() => {
-                                    onShowNav;
-                                }}
+                                onClick={onShowNav}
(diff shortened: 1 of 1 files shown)


PThorpe92:
This issue was fixed with the large refactor in PR's #436 and #445, and currently works in staging :+1: 

Closing because the fix provided here will no longer work with the new TS strict typing we have enforced, because onShowNav is defined as ? optional, it now has to be done like 
```ts
                            <ULIComponent
                                onClick={() => {
                                    if (onShowNav) onShowNav();
                                }}
```

I think on wednesday I will be going over some of the changes that were made to the linting/build system, and the routing fixes.

Issue #467, opened 2024-10-28
User is unable to delete first link in a collection


calisio:
**Describe the bug**
When a user attempts to delete the first link in a collection, they are unable to do so.

**To Reproduce**

1. Go to Resource Management page on admin view
2. Click on a collection to modify
3. Attempt to delete the first link in the collection
4. See error - cannot successfully do this.

**Expected behavior**
You should be able to delete the first link. Also, if there is a problem deleting a link (or a collection for that matter), a toast should appear telling the user there is an error.

**Screenshots**

https://github.com/user-attachments/assets/9e19531e-318b-496a-8ed0-14b7a49eeed1

**Desktop (please complete the following information):**
Mac OS Version 14.5
Browser Chrome Version 130.0.6723.70

**Additional context**

Arghya721:
@calisio there is an issue with `ResourcesManagement.tsx` where we are setting `activeLinkToDelete` to 0 when trying to delete the first link in a collection. Now in the check at line 686 
```
onSuccess={() =>
                            activeLinkToDelete &&
                            deleteLink(collection, activeLinkToDelete)
                        }
```
`activeLinkToDelete` is treated as false (since it is set to zero) hence it is not being deleted.

jtucholski:
Closing this. I haven't been able to reproduce it and I think that it was addressed in #470 (like mentioned up above).

PR #470, opened 2024-10-29
fix: update condition to check for active link deletion for the first index

Arghya721:
## Description of the change

Updated `activeLinkToDelete` check from a boolean to an explicit undefined check to correctly include the first index.

- **Related issues**: #467 

## Screenshot(s)

https://github.com/user-attachments/assets/2406a467-80a2-4787-a6ad-67adbb0df866

## Additional context

The issue in `ResourcesManagement.tsx` is that `activeLinkToDelete` is set to 0 when attempting to delete the first link in a collection. Since 0 is treated as a falsy value in JavaScript, the condition `activeLinkToDelete && deleteLink(collection, activeLinkToDelete)` fails, preventing the deletion of the first link. In this case, replacing the condition with an explicit check for `undefined` (to align with the author's intent) would resolve the issue and allow deletion of the first link.

### DIFF
--- a/frontend/src/Pages/ResourcesManagement.tsx
+++ b/frontend/src/Pages/ResourcesManagement.tsx
@@ -683,7 +683,7 @@ const ResourceCollectionEditor = ({
                         item="Link"
                         onCancel={() => setActiveLinkToDelete(undefined)}
                         onSuccess={() =>
-                            activeLinkToDelete &&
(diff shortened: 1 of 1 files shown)


PThorpe92 (approved):
This looks good to me :+1: 

Thanks for the contribution! :heart:

Issue #464, opened 2024-10-28
Create Department-wide Program


jtucholski:
# Create New Program & Mark Department-wide

As a Department-wide Program Coordinator (Admin) I want to create new programs that are department-wide so that these programs are available in every facility.

## Description

Create functionality in UnlockEd that allows the creation and management of Programs. Programs will include both educational courses (e.g., Hi-SET, college-level courses via Brightspace or Moodle) and rehabilitative offerings (e.g., therapy sessions, vocational training). These programs are going to be assigned to all facilities upon creation.

I am unsure if the best way to approach this task is to use some sort of department-wide flag so that it is always available to all facilities or to add a "Department-wide" checkbox that automatically checks all of the facilities.

### Task List

- [x] Update the Programs model with a Credit Type field and add a database migration for this new field.
- [x] Update the Programs model with a “Status” and “Program Type” fields and add a database migrations for these new fields.
- [x] Add Programs tab to the Admin Nav (check with #465 to see if this page is done)
  -  **Note** this Programs page (to view a list of programs) doesn't exist and will need to be completed as part of a separate task.
- [x] Create the Create Program form
  - Include fields for name, description, and credit type.
- [ ] Use the `POST api/programs` handler to create new Programs.

### Acceptance Criteria

* A user can create new programs in the system, including basic details (name, description, and credit type).
* A user can make the program department-wide.
* The program exists in the database and all of the data entered was captured.
* The program is assigned to all facilities in the system (upon creation).

### User Flow

* **Program Creation Flow:**
  * Coordinator logs into the system and navigates to the "Programs" tab in the admin dashboard.
  * Clicks “Create New Program.”
  * Enters basic program information (e.g., name, description, type).
  * Select credit type (e.g., Academic Credit, Participation Credit). 
    * For available credit types reach out to @chrissantillan on Slack.  
  * ~~Program is checked or set to be _department-wide._~~
  * Click "Save" to create the program.  

### Sample Program Data

Note this program data includes additional information beyond the scope of this ticket for presentation purposes.

```json
{
  "program_name": "Life Skills Program",
  "description": "A 10-week course focused on improving essential life skills.",
  "credit_type": "Participation Credit",
  "tags": ["Earned-Time Credit", "Certificate"],
  "facilities": ["Maine Correctional", "Mountain View Correctional"],
  "courses": [
    {
      "course_name": "Communication Skills",
      "lms": "Brightspace",
      "course_id": "BSPROG001"
    },
[...]

PR #477, opened 2024-10-31
Create-department-wide-program-464

corypride:
## Description of the change

Create functionality in UnlockEd that allows the creation and management of Programs.

Fixes: #464

### DIFF
--- a/frontend/src/app.tsx
+++ b/frontend/src/app.tsx
@@ -243,15 +243,15 @@ const router = createBrowserRouter([
                             path: ['open-content-management']
                         }
                     },
-                    // {
-                    //     path: 'programs',
-                    //     element: <Programs />,
-                    //     errorElement: <Error />,
-                    //     handle: {
-                    //         title: 'Programs',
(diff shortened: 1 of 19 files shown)


frontend/src/app.tsx
```diff
@@ -242,6 +243,15 @@ const router = createBrowserRouter([
                             path: ['open-content-management']
                         }
                     },
+                    // {
```
PThorpe92, line 246: was this supposed to be commented out?

I see in staging this is now a live route

PThorpe92 (changes requested):
Other than that route, this looks good to me :+1:

PThorpe92:
I looked at the app.tsx file and it looks like we are importing programs twice, once is commented out, but the route is defined three times.

If it's the same route, and the same component, it only needs to be defined once (not in the WithAdmin section, because that route needs to be accessible to everyone, and it contains the logic to handle each type of user)

PThorpe92 (approved):
Looks good man

PR #479, opened 2024-11-04
 #4 Impl ImportCourses for BrightspaceService

carddev81:
@PThorpe92 Initial implementation of ImportCourses for BrightspaceService. Retrieves CSV from Brightspace REST API, parses the CSV into UnlockEd type course, if the Brightspace course does not exist within UnlockEd, makes REST API call to retrieve course images and if one does not exist a default one is used.  Example of implemented courses:

image


provider-middleware/brightspace_data.go
```diff
@@ -49,15 +55,116 @@ type BrightspaceEnrollment struct {
...
+		Message string `json:"message"`
+	}{}
+	err = json.NewDecoder(response.Body).Decode(&urlRes)
+	if err != nil {
+		return "", err
+	}
+	return urlRes.Data.Url, nil
+}
+
+func getDefaultImg() ([]byte, error) {
```
PThorpe92, line 155: In the case that the download fails, can we have this just set the ThumbnailURL to `/brightspace.png` which we can keep in the /frontend/public directory, so we don't have to read and upload this default image each time?

PThorpe92 (approved):
made the changes so keighan could put his PR in that depended on ReadCSV, etc. will discuss with you tomorrow. looks good :+1:

provider-middleware/brightspace_data.go
carddev81: Yeah, much more efficient.  I'll make the change in the upcoming PR.

PR #480, opened 2024-11-05
add youtube/video downloads and viewer

PThorpe92:
This PR adds functionality to download videos from most popular video websites, so they can be viewed by students in the open content page/viewer.

image

The Provider Middleware is now equipped with three additional jobs, which it subscribes to via the NATS message queue.

1. Add videos:
   - This job accepts an array of video URLs input by an administrator, and attempts a download one at a time.
   
image
image

2. Retry failed downloads
  - This job is run automatically every 3 hours, and attempts to re-download any videos that have failed to download in the past. 
  - Each time a download is attempted, an `attempt` is added for that video, and after 5 attempts, the video will no longer retry, and the administrator will be prompted to remove the video, and the relevant error message will be displayed. 
 
3. Retry single video download
  - An administrator can select to attempt to retry the download manually, although the previous rule of 5 attempts maximum still applies. There is an endpoint at `/api/videos/{id}/retry` that will publish the job to NATS.

This feature requires a google data/youtube API key, which needs only be entered once. 
image

The `Add Videos` button will not appear until the user has added the relevant API key. 

Then the video can be viewed upon clicking a valid link
image


calisio:
Does this have other changes besides the youtube additions? It seems like there are a bunch of files that have changes that are tied to linting maybe? Reviewing 75 files and over 2,000 lines of code seems tough

calisio:
Screenshot 2024-11-08 at 3 43 34 PM
This is the student view of videos- you probably shouldnt see "Video management" here, since the student wouldn't be managing the content. I don't even really think you really need that header at all honestly.

calisio:
Screenshot 2024-11-08 at 3 47 02 PM
In terms of real life usability, I think we should consider putting some sort of informational hover instructing the admin what a "youtube key" is, and how they would be able to get one. I only know what that is because you told me that it was in a developer account on google, but I had to look up how to get one for it to work. This can be a future ticket, but I definitely think it should be something to consider.
Something like this:

image

calisio:
After I put in the URL and it told me it was attempting to download, maybe we should refresh the page automatically in order to be able to view the video? In the future we should definitely include something in the UI that illustrates that we received the request for that specific video and display that the content is loading to keep the user up-to-date on the status of their request. If not, the user is left there waiting on an empty screen not knowing if it has gone through or not. In order to see if my video had gone through successfully, I had to refresh the page (where it then appeared).

On that note- if you refresh the page (which you have to) it sets the tab back to the default "Libraries" tab, even though we are on the video page. See the top tab view:
Screenshot 2024-11-08 at 3 55 54 PM

frontend/src/app.tsx
calisio, line 1: I haven't looked too deep into this code, so if I'm completely off just ignore me. It seems like you condensed the admin view of open content management into the same page as the open content for the student. In developing the open content pages, I had left "Open Content" and "Open Content Management" as two separate pages because Chris and Nokie have been talking about being able to view student pages from the admin view (ie you would see what the student sees with open content, without the toggle buttons). I can still see that being a possibility (I think?) if we just change the role on that page to render the correct content, but let me know what you think. Again, I haven't reviewed the code in depth because there is a lot of it & I won't get into the details before the weekend. Just wanted to surface that in case it wasn't brought to your attention before.

calisio:
As a user, if i put an invalid link (such as that of a YT profile), I should probably be notified that it is either an invalid link or that it has failed.
Screenshot 2024-11-08 at 4 23 03 PM
Screenshot 2024-11-08 at 4 23 39 PM
Right now, it shows nothing after the success toast (even after a refresh), since it does not create a card because it is not a video

calisio:
Should you be able to upload the same video twice?
Screenshot 2024-11-08 at 4 31 18 PM

calisio:
Features I have tried that work:

- searching by the title
- adding multiple links WITH commas
- visible/hidden toggle & successfully changes it in the student portal
- viewing the video in the video viewer page

Features that do not work:

- Deleting a video does not work, it rather redirects you to the video viewing page. I assume that is because you have the onclick that redirects to the video viewing page to be on the entire card. You’ll need to increase the z-index of the trash can for it to trigger that button over the redirect.
- Sort does not work. If we dont need it at the moment, you can just take it out for now and we can add a ticket to add it into this page later.
- pagination stops at 10. Make sure to either add in pagination to the bottom of this page so you can move through the pages
- I would recommend returning the list of videos in a specific order (I think that for open content libraries it is sorted automatically by created_at), that way when a user updates the state from visible/hidden it always stays in the same spot. Right now, if you switch it to visible/hidden it jumps around to a different spot on the screen making it hard to keep track where it moved to.

calisio:
I can help clean up some of the other parts of the UI in this page so it stays consistent (sadly not before Monday). Overall this is a SICK feature and I cant wait for it to be shown off!! 😎

frontend/src/app.tsx
PThorpe92: Solution: OpenContentManagement is it's own route, and I added a feature for an Admin to "View as Student", which will just direct the admin to the other route, and in the component: it will check whether it is an administrator viewing on that particular route: causing it to render itself as a student.

image

image

### Libraries

image

image

PThorpe92:
>This is the student view of videos- you probably shouldnt see "Video management" here, since the student wouldn't be managing the content.
 - [x] fixed: 

> I think we should consider putting some sort of informational hover instructing the admin what a "youtube key" is

  - [x] I realized that I can just avoid using youtube's service for fetching the data and can instead rely in using yt-dlp itself, so we can do away with the notion of an api-key.

 >Should you be able to upload the same video twice?

  - [x] Good point, fixed

>(videos not showing up upon submission...) This makes me as the user assume that the YT link failed

 - [x]  Good idea. Added a polling system that will call `mutate()` every N seconds after videos are added, if no additional videos are fetched in N seconds, we wait N*2 seconds then request again until at least 1 of the videos requested shows up, or we reach a 10 second wait time. 
While this polling process continues, users are not allowed to manually request retries for obvious reasons. also changed the middleware to save + download each video in it's own goroutine so will be much faster.
image


>Sort 

  - [x] fixed

>Pagination 
 
  - [x] fixed

> Deletion

  - [x] fixed
>As a user, if i put an invalid link (such as that of a YT profile), I should probably be notified that it is either an invalid link or that it has failed.

- [x] :+1:  It's kinda difficult to tell just from a link whether it will actually work or not, because youtube has a few different formats that act as valid links. Also, since we can accept about any video site, we have to account for that and really can only give warnings that the link might be invalid. For this reason, I have added an ignore button so they have the ability to try it anyway.
image

> would recommend returning the list of videos in a specific order

- [x] defaults to created_at DESC unless they specify otherwise

PThorpe92:
Added favorites 
image

PThorpe92:
Also added:

Videos will be hosted on s3, and many/all instances will share the same s3 bucket. Along with the videos stored as .mp4 files, we will also be storing `.json` metadata for each video, so if one instance downloads a video, it will be made available to all other instances to choose from, preventing any downloads from ever happening twice.

I added a cronjob that will fetch all .json from s3 to update the instances videos table so that they all stay up to date. We offer as many libraries as possible, so it would be ideal to offer as many videos as possible too and administrators can hide/show them as they please

PR #484, opened 2024-11-07
User form fix

### LINKED ISSUE #463: BUG: Validating When Adding Users
Should we allow spaces in the first or last name? This is a valid first name but it is blocked from getting added.

Image

MillerSenior:
Screenshot 2024-11-06 170711
Screenshot 2024-11-06 170933
Screenshot 2024-11-06 171057
Added functionality to allow users to add usernames, first, and last names that contain spaces. Closes #463

### DIFF
--- a/frontend/src/Components/inputs/TextInput.tsx
+++ b/frontend/src/Components/inputs/TextInput.tsx
@@ -33,7 +33,18 @@ export function TextInput({
                 value: length,
                 message: `${label} should be ${length} characters or less`
             }
-        })
+        }),
+        // Adding pattern validation only for fields that should allow spaces (username, name_first, name_last)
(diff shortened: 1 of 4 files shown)


frontend/src/Components/inputs/TextInput.tsx
```diff
@@ -33,7 +33,18 @@ export function TextInput({
                 value: length,
                 message: `${label} should be ${length} characters or less`
             }
-        })
+        }),
```
PThorpe92, line 36: Can you remove the 'username' field from this? Only the first and last name fields are allowed spaces.


Also can you remove all the changes to the backend from this PR? I included a more robust fix for this issue in #480. I should have referenced the issue

Issue #490, opened 2024-11-08
Track open content library activity


jtucholski:
## User Story:

As the **System**, I want to **track viewing activity when residents interact with Open Content**, so that I can build dashboards and offer insights to residents on their dashboard.

## Description:

To provide various Dashboard components to users (such as displaying top N libraries or facility top N libraries), we need to track resident activity when they browse open content. There has been discussion about whether this tracking might be related to "activity data," which has historically been used for capturing course progress. However, at the time of creating this task, this approach hasn't been finalized. It still needs to be determined if there's a more optimal method for tracking this information.

## Tasks:

- [x]  Modify reverse proxy to log requests / capture activity
    - Log/capture requests that residents make when visiting open content pages. Capture, at minimum:
        - User id
        - Library id
        - Content path (if possible)
        - Timestamp

## User Flow:

**Viewing Open Content**

- Resident navigates to an Open Content Library (or Video)
- System records their interaction with the Open Content Library (or page).

## Technical considerations:

Some libraries may serve as inspiration for "page analytics." Currently, we don't have requirements to track time spent on web pages. Avoid overdesigning for this task, but keep in mind that more sophisticated needs may arise in the future.

Future tasks will require querying this data, so consider logging or capturing it in a way that facilitates easy querying.

## Acceptance Criteria:

- The system captures all page requests when a resident interacts with Open Content.

PR #514, opened 2024-11-20
Track open content library activity #490

carddev81:
Added feature to track users (students/admins) viewing activity when interacting with open content (videos and libraries). A couple of details follow:

Two tables were created to track open content library activity.
1. open_content_urls - This table is used as a lookup table for when a user clicks an open content link. If the url that the user requested does not exist then it is created and the ID is used for the saving in the open_content_activities table. This table is used for improving query performance.  
2. open_content_activities - This table is used to track all user open content activity, basically tracks user requests for all open content.

### DIFF
--- a/backend/migrations/00021_add_open_content_activity_tables.sql
+++ b/backend/migrations/00021_add_open_content_activity_tables.sql
@@ -1,0 +1,36 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE open_content_urls (
+        id SERIAL NOT NULL, 
+        content_url CHARACTER VARYING(255) NOT NULL, 
+        PRIMARY KEY (id)
+);
+
+CREATE TABLE open_content_activities (
+        id SERIAL NOT NULL, 
+        request_ts TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT CURRENT_TIMESTAMP, 
+        open_content_provider_id INTEGER NOT NULL, 
+        facility_id INTEGER NOT NULL,
+        user_id INTEGER NOT NULL, 
+        content_id INTEGER NOT NULL, 
+        open_content_url_id INTEGER NOT NULL, 
+        PRIMARY KEY (id), 
+
+        CONSTRAINT open_content_activities_open_content_provider_id_fkey FOREIGN KEY (open_content_provider_id) REFERENCES "open_content_providers" ("id") ON DELETE CASCADE ON UPDATE CASCADE, 
+        CONSTRAINT open_content_activities_facility_id_fkey FOREIGN KEY (facility_id) REFERENCES "facilities" ("id") ON DELETE CASCADE ON UPDATE CASCADE,
+        CONSTRAINT open_content_activities_user_id_fkey FOREIGN KEY (user_id) REFERENCES "users" ("id") ON DELETE CASCADE ON UPDATE CASCADE, 
+        CONSTRAINT open_content_activities_open_content_url_id_fkey FOREIGN KEY (open_content_url_id) REFERENCES "open_content_urls" ("id") ON DELETE CASCADE ON UPDATE CASCADE, 
+        CONSTRAINT unique_user_facility_library_url_timestamp UNIQUE (user_id, facility_id, content_id, open_content_url_id, request_ts)
+);
+
+CREATE INDEX idx_open_content_activities_user_id ON public.open_content_activities USING btree (user_id);
+CREATE INDEX idx_open_content_activities_open_content_url_id ON public.open_content_activities USING btree (open_content_url_id);
+CREATE INDEX idx_open_content_activities_user_id_content_id_facility_id ON public.open_content_activities USING btree (user_id, content_id, facility_id);
+CREATE INDEX idx_open_content_activities_content_id_open_content_provider_id_facility_id ON public.open_content_activities USING btree (content_id, open_content_provider_id, facility_id);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE open_content_activities;
+DROP TABLE open_content_urls;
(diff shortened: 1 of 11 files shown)


backend/migrations/00021_add_open_content_activity_tables.sql
```diff
@@ -0,0 +1,36 @@
...
+
+CREATE INDEX idx_open_content_activities_user_id ON public.open_content_activities USING btree (user_id);
+CREATE INDEX idx_open_content_activities_open_content_url_id ON public.open_content_activities USING btree (open_content_url_id);
+CREATE INDEX idx_open_content_activities_user_id_content_id_facility_id ON public.open_content_activities USING btree (user_id, content_id, facility_id);
+CREATE INDEX idx_open_content_activities_content_id_open_content_provider_id_facility_id ON public.open_content_activities USING btree (content_id, open_content_provider_id, facility_id);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE open_content_activities;
```
PThorpe92, line 34: casdade when dropping to remove indexes

backend/src/handlers/middleware.go
```diff
@@ -62,23 +70,73 @@ func (srv *Server) libraryProxyMiddleware(next http.Handler) http.Handler {
...
+				ContentID:             proxyParams.ID,
+			}
+			srv.createActivity(urlString, activity)
+		}
+		ctx := context.WithValue(r.Context(), libraryKey, &proxyParams)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	}))
 }
 
+func (srv *Server) createActivity(urlString string, activity models.OpenContentActivity) {
```
PThorpe92, line 125: pass by reference models.OpenContentActivity

backend/src/database/libraries.go
```diff
@@ -44,15 +44,15 @@ func (db *DB) GetLibraryByID(id int) (*models.Library, error) {
 	return &library, nil
 }
 
-func (db *DB) ToggleLibraryVisibility(id int) error {
+func (db *DB) ToggleLibraryVisibility(id int) (models.Library, error) {
```
PThorpe92, line 47: can we return by reference here?

backend/src/handlers/libraries_handler.go
```diff
@@ -55,9 +58,43 @@ func (srv *Server) handleToggleLibraryVisibility(w http.ResponseWriter, r *http.
...
 		return newDatabaseServiceError(err)
 	}
+	if srv.buckets != nil { //make sure to update value in bucket if exists
+		libraryBucket := srv.buckets[LibraryPaths]
+		updateLibraryBucket(libraryBucket, r.PathValue("id"), library, log)
+	}
 	return writeJsonResponse(w, http.StatusOK, "Library visibility updated successfully")
 }
+
+func updateLibraryBucket(libraryBucket nats.KeyValue, key string, library models.Library, log sLog) {
```
PThorpe92, line 73: if this is a method: we can avoid passing keyValue, and another pass library by ref

backend/migrations/00021_add_open_content_activity_tables.sql
carddev81: Added cascade command to drop statement.

backend/src/handlers/middleware.go
carddev81: fixed both of these

backend/src/database/libraries.go
carddev81: made change to return by reference, and renamed to ToggleVisibilityAndRetrieveLibrary because the library is needed for  bucket logic.

backend/src/handlers/libraries_handler.go
carddev81: made both changes

backend/src/handlers/middleware.go
```diff
@@ -62,19 +70,54 @@ func (srv *Server) libraryProxyMiddleware(next http.Handler) http.Handler {
...
-		if user.isAdmin() {
-			tx = tx.First(&library)
-		} else {
-			tx = tx.First(&library, "visibility_status = true")
+		if err != nil { //check if there was an error
+			if srv.Db.Debug().Model(&models.Library{}).Preload("OpenContentProvider").Where("id = ?", resourceID).First(&library).RowsAffected == 0 {
+				srv.errorResponse(w, http.StatusNotFound, "Library not found.")
+				return
+			}
+			proxyParams = models.LibraryProxyPO{
```
PThorpe92, line 89: hey can we make a method on `models.Library` called `IntoProxyP0`, just so we can bring this down to 1 line?

backend/src/handlers/proxy_handler.go
```diff
@@ -64,6 +64,11 @@ func (srv *Server) handleForwardKiwixProxy(w http.ResponseWriter, r *http.Reques
 			},
 		},
 		ModifyResponse: func(res *http.Response) error {
+			//MAKE SURE TO NOT CACHE MAIN HTML CONTENT | WE MAY NEED TO ADD OTHER CONTENT TYPE TO NOT CACHE
+			contentType := res.Header.Get("Content-Type")
+			if contentType == "text/html" || contentType == "" {
+				res.Header.Set("Cache-Control", "no-store, no-cache, must-revalidate, private")
```
PThorpe92, line 70: Might want to also set `Pragma: no-cache` as well

backend/src/handlers/middleware.go
carddev81: Good call, made the change, added IntoProxyPO method to Library and made the necessary changes.

backend/src/handlers/proxy_handler.go
carddev81: added http/1.0 Pragma no-cache header

Issue #442, opened 2024-10-16
BUG: Pager Control Admin and Student Management


jtucholski:
The pager control at the bottom of the Admin & Student Management screen indicates more data even though the Admin Management screen has one Admin that gets displayed.

I assume that since this page was recently split out under [Split User Management into Students-only and Admins-only Views #382](https://github.com/UnlockedLabs/UnlockEdv2/issues/382). **The Pager Control at the bottom may be calculating the number of pages based on all users (admin and student) and not just the type of user shown on the screen.**

## Screenshots

![Image](https://github.com/user-attachments/assets/f06c1371-026a-4ce2-8e51-acb01bdb8b27)
![Image](https://github.com/user-attachments/assets/86b17a80-aecd-41a9-9035-54600077def2)

corypride:
The first part of this bug can be fixed by correcting the bug in the function that returns "count" which is used for the total of pages. The following code fixes that, but there is still an issue with the page not rendering the total correctly.... not quite sure what it is....  but here is the code that fixes the GetCurrentUsers func:

func (db *DB) GetCurrentUsers(page, itemsPerPage int, facilityId uint, order string, search string, role string) (int64, []models.User, error) {
	if search != "" {
		return db.SearchCurrentUsers(page, itemsPerPage, facilityId, order, search, role)
	}
	offset := (page - 1) * itemsPerPage
	var count int64
	var users []models.User

	tx := db.Model(&models.User{}).Where("facility_id = ?", facilityId)
	switch role {
	case "admin":
		tx = tx.Where("role = 'admin'")
	case "student":
		tx = tx.Where("role = 'student'")
	}

	if err := tx.Count(&count).Error; err != nil {
		return 0, nil, newGetRecordsDBError(err, "users")
	}

	tx = tx.Order(getValidOrder(order))

	if err := tx.Find(&users).
		Offset(offset).
		Limit(itemsPerPage).
		Error; err != nil {
		log.Printf("Error fetching users: %v", err)
		return 0, nil, newGetRecordsDBError(err, "users")
	}
	log.Tracef("found %d users", count)
	return count, users, nil
}

PR #516, opened 2024-11-20
BUG-FIX: Pager Control Admin and Student Management

corypride:
## Description of the change
Refactored GetCurrentUsers func to account for the count not being correctly calculated and accounted for the Limit(n) not being applied in the proper order in the qry execution.

- **Related issues**: 
Closes: #442 

## Screenshot(s)

https://github.com/user-attachments/assets/a55aa99d-1163-4155-8439-e9b6fb6a528f

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -29,24 +29,29 @@ func (db *DB) GetCurrentUsers(page, itemsPerPage int, facilityId uint, order str
 	offset := (page - 1) * itemsPerPage
 	var count int64
 	var users []models.User
-	if err := db.Model(&models.User{}).Where("facility_id = ?", facilityId).Count(&count).Error; err != nil {
-		return 0, nil, newGetRecordsDBError(err, "users")
-	}
-	tx := db.Model(&models.User{})
+
+	tx := db.Model(&models.User{}).Where("facility_id = ?", facilityId)
 	switch role {
 	case "admin":
-		tx = tx.Where("role IN ('admin', 'system_admin')")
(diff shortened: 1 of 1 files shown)


backend/src/database/users.go
```diff
@@ -29,24 +29,29 @@ func (db *DB) GetCurrentUsers(page, itemsPerPage int, facilityId uint, order str
...
 	var users []models.User
-	if err := db.Model(&models.User{}).Where("facility_id = ?", facilityId).Count(&count).Error; err != nil {
-		return 0, nil, newGetRecordsDBError(err, "users")
-	}
-	tx := db.Model(&models.User{})
+
+	tx := db.Model(&models.User{}).Where("facility_id = ?", facilityId)
 	switch role {
 	case "admin":
-		tx = tx.Where("role IN ('admin', 'system_admin')")
```
PThorpe92, line 38: this needs to stay the way it was, otherwise it will not work for the system_admin user

backend/src/database/users.go
```diff
@@ -29,24 +29,29 @@ func (db *DB) GetCurrentUsers(page, itemsPerPage int, facilityId uint, order str
...
+		tx = tx.Where("role = 'admin'")
 	case "student":
 		tx = tx.Where("role = 'student'")
 	}
+
+	if err := tx.Count(&count).Error; err != nil {
+		return 0, nil, newGetRecordsDBError(err, "users")
+	}
+
 	tx = tx.Order(getValidOrder(order))
```
PThorpe92, line 45: can we combine the Order with the other chain below?

PR #519, opened 2024-11-21
feat: add library favorites

### LINKED ISSUE #493: Allow Residents to Favorite Libraries
## User Story:

As a **Resident**, I want to **favorite libraries**, so that I can easily access them later.

## Description:

This story involves introducing functionality that makes it possible for residents to “favorite” libraries that they may regularly visit or come back to.

All favorite libraries can be viewed on the Open Content page. In another story, [[Dashboard > Display “favorited” libraries](https://www.notion.so/Dashboard-Display-favorited-libraries-13601ac3401f8039a084e466a3c176d3?pvs=21)](https://www.notion.so/Dashboard-Display-favorited-libraries-13601ac3401f8039a084e466a3c176d3?pvs=21) users will be able to see their favorited libraries on the Dashboard page and also have a link that navigates them to the Open Content > Favorites view.

## Tasks:

- [x]  Add a database migration to allow for users to favorite open content.
- [x]  Expose API endpoint (i.e., `PUT /api/open-content/{id}/save`) operations that allow a user to favorite/unfavorite an open content item.
- [x]  Expose API endpoint operations that retrieve a user’s favorited open content items.
- [x]  Add Favorites tab to Open Content page to display favorited libraries.
- [x]  Update the Open Content page to allow Residents ability to “favorite” an open content library.

## User Flow:

**Resident toggles favorite setting for an open content library**

- The resident visits the Open Content page.
- Libraries that are favorited display a filled-in bookmark icon.
- Libraries that aren't favorited show a transparent bookmark outline.
- Residents can click on an Open Content library's bookmark button to favorite or unfavorite it.

**User views open content favorites**

- Resident navigates to the Open Content > Favorites page
- All favorited libraries are displayed on this page
- Resident can unfavorite library from this page and it is removed from their view.

**Favorited content is disabled by administrator**

- Precondition: A user has favorited an open content library that an administrator has since disabled.
- User visits the Open Content > Favorites page.
- Previously visible libraries that are now disabled appear grayed out with a "Disabled" label.
- User can remove these libraries from their favorites.

## Acceptance Criteria:

- The Resident can visit a page under Open Content that shows their favorited open content libraries.
- The Resident can favorite/unfavorite open content libraries.
- Any favorited libraries that have been hidden by administrators still show but are visible marked indicating they are disabled.

CK-7vn:
## Description of the change

This feature adds the ability for users to favorite libraries. A user clicks the star icon in the right hand corner of the LibraryCard component located in the Libraries tab of Open Content, and that Library is added to the library_favorites table. This also opens up two endpoints in handlers/open_content.go, GET /api/open-content/favorites and PUT /api/open-content/{id}/save. The GET requests returns a users favorites in the order of most recently favorited. The PUT handler checks the database and either adds a favorite row in the library_favorites table for that library and that user, and or soft-deletes the favorite from the library_favorites table if an entry for that user, and that library is already present. The favorites tab combined Video favorites and Library favorites into one cumulative page so that a user can view **all** of their open content favorites. 

- **Related issues**: closes #493 

## Screenshot(s)
Screenshot from 2024-11-21 16-17-35
Screenshot from 2024-11-21 16-17-51
Screenshot from 2024-11-21 16-18-06
Screenshot from 2024-11-21 16-18-16
Screenshot from 2024-11-21 16-19-05

## Additional context

A couple unrelated changes were changing the videos favoriting icon from a bookmark to a star, and also adding an explicit favorite route for videos because the prior implementation checked an {action} which required admin privileges with the new feature flagging, this made it impossible for users to favorite a video, so that was changed and added as well.


frontend/src/Pages/Favorites.tsx
```diff
@@ -0,0 +1,150 @@
...
+                setFavorites(mappedFavorites);
+            } catch (err) {
+                setError('Error fetching favorites.');
+                console.error(err);
+            } finally {
+                setIsLoading(false);
+            }
+        };
+        void fetchFavorites();
+    }, [user]);
```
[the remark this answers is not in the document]
CK-7vn: 110% honesty, skill issue. I was having a ton of trouble getting things mapped correctly while also avoiding linting errors, let me try a couple things now that I've had a little more time to look into some things.

frontend/src/Components/LibraryCard.tsx
```diff
@@ -57,6 +88,16 @@ export default function LibraryCard({
...
             </div>
+
+            {role === UserRole.Student && (
+                <div
+                    className="tooltip tooltip-top absolute right-1 top-2 w-6 h-6 z-100"
+                    data-tip="Favorite video"
+                    onClick={(e) => e.stopPropagation()}
+                >
+                    {favoriteIcon}
+                </div>
```
calisio, line 99: Does using <ULIComponent> interfere with this? If not, for consistency could you implement this using that component? It just wraps the icon with the tooltip and you'd have to pass in the extra classes in there. I think it should probably be updated anywhere where we use the favorite icon, so make sure to update the other pages too.

frontend/src/Pages/Favorites.tsx
```diff
@@ -0,0 +1,150 @@
...
+        return <p className="text-red-500">{error}</p>;
+    }
+
+    return (
+        <div className="px-8 pb-4">
+            <h1>My Favorites</h1>
+            {favorites.length === 0 ? (
+                <p>You have no favorites yet.</p>
+            ) : (
+                <div className="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4 xl:grid-cols-6 gap-6 mt-6">
```
calisio, line 133: I am having some odd frontend behavior with the cards, and I think it is probably due to the varying columns. Could you keep it the same as how the library and video cards are with just 4 columns and then we can update it to respond to smaller screens in a different ticket (if needed)
Screenshot 2024-11-22 at 10 02 04 AM

frontend/src/Pages/Favorites.tsx
```diff
@@ -0,0 +1,150 @@
...
+        } finally {
+            setIsLoading(false);
+        }
+    };
+    if (isLoading) {
+        return <Loading />;
+    }
+
+    if (error && favorites.length !== 0) {
+        return <p className="text-red-500">{error}</p>;
```
calisio, line 124: use text-error for this

frontend/src/Components/LibraryCard.tsx
```diff
@@ -40,12 +43,40 @@ export default function LibraryCard({
...
+    ) : (
+        <StarIconOutline
+            className="w-5 cursor-pointer"
+            onClick={toggleFavorite}
+        />
+    );
 
     return (
         <div
-            className="card overflow-hidden cursor-pointer"
```
calisio, line 48: keep `overflow-hidden` in here, unless there is a specific reason you deleted it I'm not aware of.

frontend/src/Components/FavoriteCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+        if (favorite.type === 'video') {
+            navigate(`/viewer/videos/${favorite.content_id}`);
+        } else if (favorite.type === 'library') {
+            navigate(`/viewer/libraries/${favorite.content_id}`, { state });
+        }
+    };
+
+    return (
+        <div
+            className={`card relative rounded-lg p-3 shadow-md ${
```
calisio, line 32: `card` class should handle `rounded-lg shadow-md` so you dont have to repeat that in this class. you can see what is in the card class inside of `app.css`

frontend/src/Components/LibraryCard.tsx
CK-7vn: I deleted it because the tool-tip was being rendered behind, so you only saw a very small piece of the tool-tip.

calisio: oohhhh okay good point! okay to leave it as is then!

frontend/src/Components/LibraryCard.tsx
CK-7vn: Everything else I fixed, but this is confusing me?

frontend/src/Pages/Favorites.tsx
```diff
@@ -0,0 +1,150 @@
...
+                );
+
+                if (type === 'library') {
+                    await mutate('/api/libraries');
+                }
+            } else {
+                setError('Failed to unfavorite item.');
+            }
+        } catch (err) {
+            setError('Error unfavoriting item.');
```
calisio, line 113: could we have a toast that sets this error? Im not sure if a success toast is necessary, but I am feeling like an error toast would be helpful for the user to know that they had an issue favoriting that item.

frontend/src/Pages/Favorites.tsx
calisio: Actually, I would say that this probably shouldn't be rendered here. If there is an error, this would re render the whole contents of this page to show that error, when in reality we'd probably still want to show the user the favorites but instead just show them a toast telling them that we were unable to favorite/unfavorite their library. 
Screenshot 2024-11-22 at 10 17 07 AM

frontend/src/Components/LibraryCard.tsx
calisio: like it could maybe look something like this?
```
                <div onClick={(e: MouseEvent) => toggleFavorite(e)}>
                    <ULIComponent
                        tooltipClassName={'absolute right-2 top-2 z-100'}
                        iconClassName={`w-6 h-6 ${library.is_favorited ? 'text-primary-yellow' : ''}`}
                        icon={library.is_favorited ? StarIcon : StarIconOutline}
                        dataTip="Favorite Library"
                    />
                </div>
```
and inside of togglefavorite you'd just have to call `e.stopPropogation( )` - I don't think you need `e.preventDefault()` since I don't think this element has any default behavior?

Also- this tooltip should be favorite library rather than favorite video :)

calisio: I have the `onClick` on a div outside of it since the `ULIComponent`'s `onClick` doesn't accept a param, although maybe we could update it to optionally accept a mouse event to be able to pass in that function into the `ULIComponent` so we dont have to have a wrapper div?

frontend/src/Components/FavoriteCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+                favorite.visibility_status ? 'bg-grey-2' : 'bg-inner-background'
+            } ${
+                favorite.visibility_status
+                    ? 'cursor-not-allowed'
+                    : 'cursor-pointer '
+            }`}
+            style={{ width: '220px', height: '220px' }}
```
calisio, line 39: also this could be added into the `className` above as `w-[220px] h-[220px]` rather than using the `style` since we don't typically use style.

frontend/src/Pages/OpenContent.tsx
```diff
@@ -41,7 +54,7 @@ export default function OpenContent() {
                         className="button border border-primary bg-transparent text-body-text"
                         onClick={() =>
                             navigate(
-                                `/open-content-management/${activeTab.value}`
+                                `/open-content-management/${String(activeTab.value).toLowerCase()}`
```
calisio, line 57: If the admin is on the student view of the favorites and they click return to admin view, an error will be thrown since there is no such thing as a favorites page for an admin. Here we will probably have to check their location and what role they are trying to switch to, and handle the case that they are on the favorites page trying to switch to admin view and just direct them to the libraries tab.
Screenshot 2024-11-22 at 10 53 08 AM

Screenshot 2024-11-22 at 10 49 56 AM

backend/migrations/00022_add_library_favorites_table.sql
```diff
@@ -0,0 +1,28 @@
...
+        updated_at TIMESTAMP(6) WITH TIME ZONE,
+        deleted_at TIMESTAMP(6) WITH TIME ZONE,
+        user_id INTEGER NOT NULL,
+        name CHARACTER VARYING(255),
+        content_id INTEGER NOT NULL,
+        visibility_status BOOLEAN,
+        open_content_url_id INTEGER NOT NULL,
+        open_content_provider_id INTEGER NOT NULL,
+        PRIMARY KEY (id),
+        CONSTRAINT library_favorites_user_id_fkey FOREIGN KEY (user_id) REFERENCES "users" ("id") ON DELETE CASCADE ON UPDATE CASCADE,
```
PThorpe92, line 15: RE: migration conventions discussed earlier

backend/src/database/open_content.go
```diff
@@ -68,3 +69,88 @@ func (db *DB) CreateContentActivity(urlString string, activity *models.OpenConte
...
+            fav.id,
+            fav.name,
+            'library' as type,
+            fav.content_id,
+            lib.image_url as thumbnail_url,
+            ocp.description,
+            NOT lib.visibility_status AS visibility_status,
+            fav.open_content_provider_id,
+            ocp.name AS provider_name,
+            fav.created_at AS created_at
```
PThorpe92, line 115: `Debug()` can be removed, and here you can just select `fav.created_at` w/o the alias

frontend/src/Components/FavoriteCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <figure className=" w-full mx-auto object-contain bg-contain">
+                <img
+                    src={favorite.thumbnail_url}
+                    alt={favorite.name}
+                    className="w-full h-28 object-contain rounded-md mb-3"
+                />
+            </figure>
+            <h3 className="text-lg font-bold text-header-text text-center mb-1 line-clamp-1">
+                {favorite.name}
+            </h3>
```
calisio, line 61: `h3` should already have most necessary classes applied, such as text size, color, and weight. there are a few other things that i changed just to match the cards on the other pages in terms of spacing, etc
```
<div
            className={`card relative p-4 space-y-2 ${
                favorite.visibility_status
                    ? 'bg-grey-2 cursor-not-allowed'
                    : 'bg-inner-background cursor-pointer'
            }`}
            onClick={favorite.visibility_status ? undefined : handleCardClick}
        >
            <div
                className="absolute top-2 right-2 cursor-pointer"
                onClick={(e) => {
                    e.stopPropagation();
                    onUnfavorite();
                }}
            >
                <SolidStar className="w-5 text-primary-yellow" />
            </div>
            <img
                src={favorite.thumbnail_url}
                alt={favorite.name}
                className="h-16 mx-auto object-contain"
            />
            <h3 className="body text-center line-clamp-2">{favorite.name}</h3>
            <p className="body-small text-center">
                {favorite.type === 'video'
                    ? favorite.channel_title
                    : favorite.provider_name}
            </p>
        </div>
```

backend/src/handlers/open_content_handler.go
```diff
@@ -76,3 +78,31 @@ func (srv *Server) handleCreateOpenContent(w http.ResponseWriter, r *http.Reques
...
+	}
+	reqBody.UserID = uint(userID)
+	reqBody.ContentID = uint(contentID)
+	if _, err := srv.Db.ToggleLibraryFavorite(&reqBody); err != nil {
+		return newDatabaseServiceError(err)
+	}
+	return writeJsonResponse(w, http.StatusOK, "Favorite toggled successfully")
+}
+
+func (srv *Server) handleGetUserFavoriteOpenContent(w http.ResponseWriter, r *http.Request, log sLog) error {
```
PThorpe92, line 101: Will this handler will need pagination?

backend/src/models/open_content.go
```diff
@@ -42,6 +42,28 @@ type OpenContentUrl struct {
 
 func (OpenContentUrl) TableName() string { return "open_content_urls" }
 
+type CombinedFavorite struct {
```
PThorpe92, line 45: Can we call this `OpenContentFavorites`? 

We have Programs being favorited as well so that is a bit more explicit

frontend/src/Pages/Favorites.tsx
```diff
@@ -0,0 +1,150 @@
...
+            }
+        };
+        void fetchFavorites();
+    }, [user]);
+
+    const handleUnfavorite = async (
+        contentId: number,
+        type: 'video' | 'library'
+    ) => {
+        try {
```
PThorpe92, line 66: line 66) calls from the API class don't need to be wrapped in try/catch block :)

(line 62) `type` is a keyword in TS, so we should avoid using it as a variable name.

lines 68-74 could be written more concisely as:

```ts
const endpoint = favType === 'video' ? `videos/${contentId}/favorite` : `open-content/${contentId}/save`;
```

frontend/src/Pages/Favorites.tsx
calisio: It also actually has to do with the fixed width of the favorites card-see comment below

frontend/src/Components/LibraryCard.tsx
CK-7vn: Cool, thanks a ton for the clarification, still trying to figure out all of this typescript and react stuff...

frontend/src/Pages/Favorites.tsx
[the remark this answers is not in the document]
CK-7vn: Think I got it. Thanks again.

frontend/src/Pages/Favorites.tsx
calisio: Looking at this code, you could actually probably do this unfavoriting within the favorite card component itself rather than in this higher level element. That would eliminate the lines of code of you checking to see if the favorite is in the favorites array since you can directly access that particular favorites information from that card. This, paired with the implementation of useSWR, would significantly reduce the amount of code that is in the `Favorites.tsx` file to around 30 total lines of code. Hard to explain without actually showing the code: 

Inside of `Favorites.tsx`:
```
export default function FavoritesPage() {
    const {
        data: favorites,
        isLoading,
        error,
        mutate
    } = useSWR<ServerResponseMany<CombinedFavorite>, AxiosError>(
        `/api/open-content/favorites`
    );

    if (isLoading) {
        return <Loading />;
    }

    return (
        <div className="px-8 pb-4">
            <h1>My Favorites</h1>
            {error || favorites == undefined ? (
                <div className="text-error">Error loading favorites</div>
            ) : favorites.data.length === 0 ? (
                <p>You have no favorites yet.</p>
            ) : (
                <div className="grid grid-cols-4 gap-6 mt-6">
                    {favorites.data.map((favorite) => (
                        <FavoriteCard
                            key={favorite.content_id}
                            favorite={favorite}
                            mutate={mutate}
                        />
                    ))}
                </div>
            )}
        </div>
    );
```

Inside of `FavoriteCard.tsx` 
```
interface FavoriteCardProps {
    favorite: CombinedFavorite;
    mutate: KeyedMutator<ServerResponseMany<CombinedFavorite>>;
}

const FavoriteCard: React.FC<FavoriteCardProps> = ({ favorite, mutate }) => {
    const navigate = useNavigate();
    const { toaster } = useToast();

    const handleUnfavorite = async () => {
        const { content_id, name, open_content_provider_id } = favorite;
        const endpoint =
            favorite.type === 'video'
                ? `videos/${content_id}/favorite`
                : `open-content/${content_id}/save`;

        const payload = {
            name,
            content_id,
            open_content_provider_id,
            content_url: `/api/proxy/libraries/${content_id}/`
        };

        const response = await API.put(endpoint, payload);

        if (response.success) {
            await mutate();
        } else {
            toaster('Failed to unfavorite item.', ToastState.error);
        }
    };

// REST OF CODE
```

follow up: do you even need to pass a payload into this api put? In libraries i just call the put and it should do the toggling for me in the backend. might be different- but check out `LibraryCard.tsx`

calisio: Also, not sure if this was implemented the way it was because of the nested favoriting- and in that case let me know. Since i am only looking at the code that is in this branch, I'm not sure what you need/don't need in the case of favoriting nested pages.

frontend/src/Pages/Favorites.tsx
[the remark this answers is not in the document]
calisio: oops didnt see this comment- i also played around with it and attached it in a [comment below](https://github.com/UnlockedLabs/UnlockEdv2/pull/519/files#r1854561457)

CK-7vn:
Screenshot from 2024-11-22 17-31-54
Screenshot from 2024-11-22 17-32-14
Screenshot from 2024-11-22 17-32-30

**ALSO**
I fixed the routing when an Admin switches from student view back to admin view while on the favorites page.

CK-7vn:
Screenshot from 2024-11-23 10-24-41
Screenshot from 2024-11-23 10-25-05
Screenshot from 2024-11-23 10-32-52
Screenshot from 2024-11-23 10-34-59

Ok Ok @calisio <- I switched the unfavoriting into the card, sorry, it should have been that way in the first place. The payload is necessary because of the way that the unfavoriting works on the backend...but I did switch change as much as I could...the bug with the Admin switching back to student view was fixed, double checked and the favorites are still being returned to you in order of most recently favorited, but that handler did change slightly to add the pagination.

frontend/src/Components/FavoriteCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+            if (favorite.type === 'library') {
+                await mutate('/api/libraries');
+            }
+        } else {
+            toaster('Failed to unfavorite', ToastState.error);
+        }
+    };
+    return (
+        <div
+            className={`card w-60 h-40  p-4 space-y-2 ${
```
calisio, line 91: dont restrict height and width here, it won't allow the cards to grow/shrink per screen. This change will also now allow for the grid to take up the entire width of the screen

frontend/src/Components/FavoriteCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+                        ...cachedData,
+                        data: newFavorites
+                    };
+                },
+                false
+            );
+
+            if (favorite.type === 'library') {
+                await mutate('/api/libraries');
+            }
```
calisio, line 84: there are two ways to do this which would condense this code a bunch:
1. I listed this in my previous comment, but you could have the `mutate` attached to the initial useSWR call that calls the favorites api and pass that function into this favorite card, and then simply just call `await mutate()` on success. Because it is established on the original useSWR call, it knows automatically which api call it has to revalidate.
2. You could skip passing in the mutate into this favorites card and instead you could import the global mutate from SWR and then give the function the api that you want to revalidate. In this case, you would also have to pass in page and per page, which would need to be passed into the card (since i guess it is required in the backend? i haven't checked the backend out, but it seems like it will not return anything if those variables are not defined. although they probably should by default be something i think...). So, the call would be one simple line, ``await mutate(`/api/open-content/favorites?page=${pageQuery}&per_page=${perPage}`)``, which will revalidate the data on this favorites page.

The reason you do not need to call mutate on `api/libraries` or videos is because those pages will be revalidated when you switch to those tabs anyways, since they are separate useSWR calls. So, when the user changes from favorites -> libraries tab, that component, while rendering, will take the useSWR hook on that page and call the `api/libraries` handler and will fetch the new, updated data.

calisio: Also if im completely off base on this and not understanding why you have to do this and it IS a requirement, please let me know!

frontend/src/Pages/Favorites.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const meta = data?.meta;
+
+    const handleSetPerPage = (val: number) => {
+        setPerPage(val);
+        setPageQuery(1);
+    };
+
+    return (
+        <div className="  p-8 ">
+            <h1>My Favorites</h1>
```
calisio, line 34: this is a super small knit picky thing, but could you remove this h1 from here? i saw it was also in the videos page and must have slipped in there while i was reviewing preston's code, but its not super necessary and looks a little weird spacing wise lol

calisio: also, in the div that has `p-8` could we add `w-full`? It just makes it so that that block will fill the whole screen rather than just a part of it
Screenshot 2024-11-25 at 11 25 05 AM

vs with `w-full`
Screenshot 2024-11-25 at 11 25 33 AM

calisio:
I think besides those comments, everything else is looking good (as long as @PThorpe92 thinks the backend looks good!)

PThorpe92 (approved):
backend looks good :+1:

CK-7vn:
Sweeet! I just force pushed those updates @calisio

calisio (approved):

Issue #506, opened 2024-11-18
Administrator > Curate Open Content Libraries


jtucholski:
## User Story:

As an **Administrator**, I want to **curate specific approved open content libraries**, so that I can make sure some of the better resources in the large collection are seen by the residents.

## Description:

Given the volume and significant size of Open Content Libraries, this story focuses on giving Administrators the ability to identify and *star* specific open content libraries to curate the list for Residents.

It's similar to a bookmark or favorite feature, but when an Administrator stars a library, it becomes visible to all Residents.

## Tasks:

- [x]  Update the Library model/table with a boolean flag to identify it as *curated*.
- [x]  Update the Open Content Library API with an endpoint for Administrators to star/unstar a library.
- [x]  Update the libraries' Open Content Management page to allow Administrators the ability to star/unstar a library.
- [x]  Add a filter setting to the Open Content page to view starred libraries.

## User Flow:

- **Starring a library**
    - Administrator accesses the Open Content Management page.
    - Administrator clicks the star icon on the library card to star or unstar it.
    - System marks the library as curated, increasing its visibility to residents.
- **Filtering by starred libraries**
    - Administrator navigates to the Open Content Management page.
    - Administrator selects "Starred" from the filter options.
    - System displays only the curated libraries.

## Acceptance Criteria:

- An Administrator can filter libraries on the Open Content Management page by "Starred" status.
- An Administrator can star a library by clicking on it.
- An Administrator can unstar a previously starred library by clicking on it again.

PR #538, opened 2024-11-27
feat: add featuring libraries

CK-7vn:
## Description of the change

This feature adds the ability for admins to feature libraries. It exposes 2 backend handlers a PUT "/api/open-content/{id}/feature" that will "feature" a library, the admins facility ID is pulled from the context and added in the table in-so that residents in the admins facility can view said featured content. The second backend handler is a GET "/api/open-content/featured" that returns a paginated response of featured content for the current users facility. 

- **Related issues**: #506 

## Screenshot(s)

Screenshot from 2024-11-25 14-48-01
Screenshot from 2024-11-25 21-19-54
Screenshot from 2024-11-25 21-21-53
Screenshot from 2024-11-25 21-22-06


backend/src/handlers/open_content_handler.go
```diff
@@ -109,3 +111,36 @@ func (srv *Server) handleGetUserFavoriteOpenContent(w http.ResponseWriter, r *ht
...
+	var reqBody models.OpenContentParams
+	if err := json.NewDecoder(r.Body).Decode(&reqBody); err != nil {
+		return newJSONReqBodyServiceError(err)
+	}
+	defer r.Body.Close()
+
+	userID := r.Context().Value(ClaimsKey).(*Claims).UserID
+	facilityID := r.Context().Value(ClaimsKey).(*Claims).FacilityID
+	contentID, err := strconv.Atoi(r.PathValue("id"))
+	if err != nil {
```
PThorpe92, line 125: if we have this check at the very beginning (since the rest of the handler depends on it), it can fail first and we can return error before opening the request body, etc

PThorpe92 (approved):

PR #542, opened 2024-12-02
feat: add handshake process between Brightspace and Unlocked - Resident-to-Student Mapping #522

### LINKED ISSUE #522: Brightspace > Resident-to-Student Mapping
## User Story:

As a **System Administrator**, I want **the ability to map resident users in UnlockEd to corresponding student accounts in Brightspace**, so that Residents see and access their Brightspace courses from UnlockEd.

## Description:

This task ensures System Administrators can integrate UnlockEd with Brightspace LMS and map user accounts between the systems. While most of this work is likely complete, the focus is on identifying and addressing any gaps between the current and desired functionality.

## Tasks:

- [x]  Add Brightspace as a provider in the Add Provider screen
    - Capture required fields like base url, account id, or access key
- [x]  Implement functionality to map users (inserting users into `provider_user_mappings`)
- [x]  Implement functionality to remove mapping from an UnlockEd user to Brightspace Student account.

## User Flow:

- Administrator clicks **Add Provider** to add a new Brightspace integration
    - The `type` dropdown includes Brightspace LMS
    - The Administrator fills in required fields relevant to integrating w/Brightspace LMS.
- Administrator clicks **Map User** on an UnlockEd user
    - The Administrator can search for and map the UnlockEd user to a Brightspace user

## Acceptance Criteria:

- Administrator can map an UnlockEd user to a Brightspace Student account
- Administrator can remove a mapping from an UnlockEd user.
- The provider_platform_mapping table is updated with the new mapping.

## Additional Notes:

- If there is development work associated with this task then convert this to a GitHub issue and link it in the PR.

carddev81:
## Description of the change

Added logic for initial handshake between UnlockEd and Brightspace for secure data exchange and integration. 

Upon submission of the form data for the Brightspace platform provider, UnlockEd will be redirected to Brightspace web site and will need Brightspace User Credentials entered to complete the OAuth2.0 authentication flow. Once logged in, Brightspace will redirect user back to the UnlockEd API, sending an access token and a refresh token that will be used by the provider middleware. The refresh token is saved into the database along side the client secret separated by a semi-colon. 

The process of logging into Brightspace should only have to occur one time. There is a scenario that we may need to refresh the token, this would only occur if there is some sort of data corruption in the database caused by the provider middleware. If this scenario were to arise, there is a 'Refresh Token' icon that was added to the Provider Platform listing screen for Brightspace LMS systems. When pressing the 'Refresh Token' icon, it will initiate the same OAuth2.0 process explained above.

**NOTE** Before the Brightspace platform provider is added to UnlockEd (see https://github.com/UnlockedLabs/UnlockEdv2/wiki/Registering-UnlockEd-with-Brightspace-LMS-%E2%80%90-(Bulk-Data-Scope)) we need the "Client ID" and "Client Secret" to be provided to us by the Brightspace IT administrator. On the Platform Provider form, the Client ID will go into the Account ID textfield and the Client Secret will go into the Access Key textfield. Also, the Brightspace registration requires the use of the https scheme, therefore the local development version of UnlockEd will not register with Brightspace. 

@PThorpe92 Added functionality for allowing user first/last names to contain spaces. Also, added functionality for allowing usernames to contain numbers. 

- **Related issues**: Closes #522"

## Screenshot(s)
PR-Pic


frontend/src/Pages/ProviderPlatformManagement.tsx
```diff
@@ -116,6 +140,24 @@ export default function ProviderPlatformManagement() {
                 toaster('Unable to toggle provider state', ToastState.error);
             });
     };
+    const refreshToken = (provider: ProviderPlatform) => {
+        const errorMsg = 'Unable to refresh token for provider for ' + provider.name;
+        API.get<ProviderResponse>(`/provider-platforms/${provider.id}/refresh`)
```
PThorpe92, line 145: the `API` class automatically prepends `/api/` to the request, so this can be 
```ts
API.get<ProviderResponse>(`provider-platforms/${provider.id}/refresh`)
```

provider-middleware/brightspace.go
```diff
@@ -17,12 +17,13 @@ import (
...
 const (
-	CsvDownloadPath     = "csvs"
 	TokenEndpoint       = "https://auth.brightspace.com/core/connect/token"
-	DataSetsEndpoint    = "https://unlocked.brightspacedemo.com/d2l/api/lp/1.28/dataExport/bds/list"
-	DataDownloadEnpoint = "https://unlocked.brightspacedemo.com/d2l/api/lp/1.28/dataExport/bds/download/%s"
+	DataSetsEndpoint    = "https://unlocked.brightspacedemo.com/d2l/api/lp/%s/dataExport/bds/list"
+	DataDownloadEnpoint = "https://unlocked.brightspacedemo.com/d2l/api/lp/%s/dataExport/bds/download/%s"
 )
 
+var CsvDownloadPath = os.Getenv("BRIGHTSPACE_TEMP_DIR")
```
PThorpe92, line 25: Can this get a default value in the `docker compose` file? also I will need to remember to set this value in the helm chart

PThorpe92 (commented):
Looks really good man, just those minor changes and this should be all set

frontend/src/Pages/ProviderPlatformManagement.tsx
carddev81: made change to path to account for /api/

provider-middleware/brightspace.go
carddev81: I included this in the docker-compose.yml file, unless I put in the wrong location? I placed this value within the provider-service image configuration.

image

PThorpe92: I must have missed that i swore i looked

carddev81: i was thinking I may have put in the wrong place. :)

PR #546, opened 2024-12-03
feat: add layer 0 operational insights page w/ charts for user data

PThorpe92:
This adds an `Operational Insights` page visible by administrators. This also creates the necessary table to track logins/counts/and times across facilities and displays the data to admins.

image

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/stretchr/testify v1.9.0
 	github.com/teambition/rrule-go v1.8.2
 	golang.org/x/exp v0.0.0-20240613232115-7f521ea00fb8
+	golang.org/x/oauth2 v0.24.0
 	gorm.io/datatypes v1.2.0
 	gorm.io/driver/postgres v1.5.9
 	gorm.io/gorm v1.25.11
@@ -74,7 +75,6 @@ require (
 	go.uber.org/multierr v1.11.0 // indirect
 	golang.org/x/crypto v0.26.0 // indirect
 	golang.org/x/net v0.27.0 // indirect
-	golang.org/x/oauth2 v0.24.0 // indirect
 	golang.org/x/sync v0.8.0 // indirect
 	golang.org/x/sys v0.25.0 // indirect
 	golang.org/x/text v0.18.0 // indirect
--- a/backend/migrations/00025_add_login_count_table.sql
+++ b/backend/migrations/00025_add_login_count_table.sql
@@ -1,0 +1,28 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE public.login_metrics (
+     user_id INTEGER NOT NULL PRIMARY KEY,
+	 total BIGINT NOT NULL DEFAULT 1,
+	 last_login timestamp with time zone DEFAULT now(),
+	 FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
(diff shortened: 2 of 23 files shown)


frontend/src/Components/StudentMetrics.tsx
```diff
@@ -0,0 +1,197 @@
...
+            {error && <div>Error loading data</div>}
+            {!data || (isLoading && <div>Loading...</div>)}
+            {data && metrics && (
+                <>
+                    <PrimaryButton
+                        className="w-30"
+                        onClick={() => setResetCache(!resetCache)}
+                    >
+                        Refresh Data
+                    </PrimaryButton>
```
calisio, line 55: primary button will soon disappear so try to use a just regular button with the class "button" on there

frontend/src/Components/StudentMetrics.tsx
```diff
@@ -0,0 +1,197 @@
...
+                    <div className="grid gap-4 mb-6">
+                        <div className="stats shadow mb-6">
```
calisio, line 57: I dont love the use of stat here, especially since we are trying to move away from daisy and more into using our own classes. there is a class called "card" which should have a correct layout (i think) similar to these, but puts them into their own individual cards. "grid" isn't doing anything to the stats cards here because that is there is only one div as a child of this div, but if you made each of these stats cards children of this larger grid div, then it would behave as you likely intended it to

frontend/src/Components/StudentMetrics.tsx
```diff
@@ -0,0 +1,197 @@
...
+    CartesianGrid,
+    XAxis,
+    YAxis,
+    Tooltip,
+    Legend,
+    Bar
+} from 'recharts';
+import PrimaryButton from './PrimaryButton';
+
+const OperationalInsights = () => {
```
calisio, line 17: if the page is called student metrics it should probably align with the exported component in this page to clarify confusion

frontend/src/Components/StudentMetrics.tsx
```diff
@@ -0,0 +1,197 @@
...
+                                        })
+                                    }
+                                />
+                                <YAxis allowDecimals={false} />
+                                <Tooltip />
+                                <Legend />
+                                <Bar dataKey="total_logins" fill="#8884d8" />
+                            </BarChart>
+                        </div>
+                    </div>
```
calisio, line 147: typically i have been extracting charts into their own component that way we can reuse them in other pages if we need to in the future

calisio: Also, try to use colors we already use in the application rather than this purple `#8884d8`

frontend/src/Components/StudentMetrics.tsx
calisio: Sorry, actually there is a component called `StatsCard.tsx` that you should use instead of these.

frontend/src/Components/StudentMetrics.tsx
```diff
@@ -0,0 +1,197 @@
...
+                                <span className="label-text">Days</span>
+                            </label>
+                            <select
+                                id="days"
+                                className="select select-bordered w-full max-w-xs"
+                                value={days}
+                                onChange={(e) =>
+                                    setDays(parseInt(e.target.value))
+                                }
+                            >
```
calisio, line 185: I think visually you could move this up to the same row where the refresh button is- kinda similar to our search/sort and add rows in other pages. so it would look something like this
Screenshot 2024-12-06 at 12 55 12 PM

calisio: with the labels too obviously

PR #552, opened 2024-12-06
fix: path for csv volume was invalid

jtucholski:
The path for the CSV volume in the Dockerfile wasn't recognized as a valid path. Updated it to include the `provider-middleware` folder.


jtucholski:
Closed. The CSV folder was moved to the root path in #553

PR #575, opened 2024-12-10
fix: Default to Kiwix thumbnail for small image libraries

### LINKED ISSUE #567: UX FIX: Add default image for libraries if provided image is too small or undetectable
Some of the thumbnails that are provided w/Kiwix are valid URLs but they are bad images. See the follow screenshot. 

Image

**Learning Statistics - CogStat** has a valid thumbnail URL (https://library.kiwix.org/catalog/v2/illustration/b25482c4-115c-42d7-9eec-8bc2342f5c29/?size=48) but the image is a transparent PNG (430 B). Images like the one with freeCodeCamp are 2.0 KB.

Let's introduce a check on the image file size. If it is below a threshold (e.g. 1.0 KB) then default to using the Kiwix logo instead. If it's above the threshold use the provided image.

CK-7vn:
## Description of the change
This checks the byte size of the image being downloaded for the library thumbnail, if it is below the constant of 2048 bytes, it sets the filename and therefor the default thumbnail URL for the library to the kiwix thumbnail helping to avoid blank images when displaying libraries in the Dashboard or the Knowledge center. 

- **Related issues**: Closes #567 

## Additional context

Also adds a check in ParseUrls in which if there is an error in downloading the thumbnail it will default to the Kiwix thumbnail.

### DIFF
--- a/provider-middleware/kiwix_data.go
+++ b/provider-middleware/kiwix_data.go
@@ -211,6 +211,8 @@ func (ks *KiwixService) ParseUrls(externId string, links []Link) (string, string
 			if !ks.thumbnailExists(externId) {
 				if thumbnail, err := ks.downloadAndHostThumbnailImg(externId, link.Href); err == nil {
 					thumbnailURL = thumbnail
+				} else {
+					thumbnailURL = "/kiwix.jpg"
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


jtucholski (approved):
After reducing the file sizing and adjusting the image path to go from the root, looks good to go!

PR #591, opened 2024-12-16
fix: change Raw to Exec, add update to total progress milestones when…

carddev81:
## Description of the change
Modified seeder to delete activities, milestones, outcomes when reran. I added logic to update total progress milestones on courses that have a 0 set as its value. Added the outcome record when the course is complete. These changes should correct any course progress issues.


PThorpe92 (approved):
thanks :+1:

PR #594, opened 2024-12-17
revert course catalog changes, fix oauth callback

PThorpe92:
This renames all the improper spelling of `catalog`, and reverts the changes to UserCourses that were made because of a seeder XY problem.

Also fixes the issue where users are not automatically logged into canvas upon clicking on a course

https://gist.github.com/PThorpe92/c693588b16452a6cfb488cc71a91a27f
Schema changes documented here


backend/src/database/activity.go
carddev81, line 1: The query that was modified needs to be reworked, the addition of the join of the milestones will cause the query to not return any results for activity, this query should change to:

	err = db.Table("courses c").
		Select(`c.id as course_id,
				c.alt_name,
				c.name,
				pp.name as provider_platform_name,
				c.external_url,
				a.total_time as time_delta`).
		Joins("JOIN provider_platforms pp ON c.provider_platform_id = pp.id").
		Joins("JOIN milestones m ON m.course_id = c.id and m.user_id=?", userID).
		Joins(`LEFT JOIN (select id, created_at,course_id, user_id, total_time, row_number() over (PARTITION BY course_id, user_id ORDER BY created_at DESC) AS RN 
				from activities
				) a on a.course_id = c.id
				and a.user_id = m.user_id
				and a.created_at >= ?
				and a.RN = 1`, time.Now().AddDate(0, 0, -7)).
		Joins("LEFT JOIN outcomes o ON o.course_id = c.id AND o.user_id = ?", userID).
		Where("o.type IS NULL").
		Group("c.id, c.alt_name, c.name, pp.name, c.external_url, a.total_time").
		Find(&results).Error

carddev81: need to discuss the logic when this check is performed:  if len(enrollments) == 0 {

backend/src/database/user_catalog.go
```diff
@@ -0,0 +1,164 @@
...
+		case "in_progress":
+			query = "o.type IS NULL"
+		}
+		if i == 0 {
+			tx.Where(query)
+		} else {
+			tx.Or(query)
+		}
+	}
+	tx.Group("c.id, c.name, a.total_time, c.thumbnail_url, pp.name, c.external_url, progress.course_progress")
```
carddev81, line 141: add c.start_dt, c.end_dt to group by

PR #596, opened 2024-12-17
fix: correct logic for descending date

carddev81:
## Description of the change

Corrected the logic to account for descending dates.  Also changed random time to be 5000 rather than 50.


PThorpe92:
changes are already in the other PR, but we can merge this anyway ig

PThorpe92 (approved):

PR #600, opened 2024-12-20
simplify the cron scheduling, remove blocking for first run to get courses

PThorpe92:
This improves the flow of the scheduled jobs, prevents blocking all other tasks if a provider is found with no courses, allowing the OCP tasks to run on first go. Simplifies the general flow.

### DIFF
--- a/backend/tasks/scheduler.go
+++ b/backend/tasks/scheduler.go
@@ -4,6 +4,7 @@ import (
 	"UnlockEdv2/src/models"
 	"encoding/json"
 	"os"
+	"slices"
 
 	"github.com/nats-io/nats.go"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


backend/tasks/scheduler.go
```diff
@@ -52,32 +53,36 @@ func (jr *JobRunner) generateOpenContentProviderTasks() ([]models.RunnableTask,
...
 			}
+			job := models.CronJob{Name: string(jobType)}
+			task, err := jr.handleCreateOCProviderTask(&job, providers[libProvider].ID)
+			if err != nil {
+				log.Errorf("failed to create task: %v", err)
+				continue
+			}
+			otherTasks = append(otherTasks, *task)
+		}
+		if jobType.IsVideoJob() {
```
carddev81, line 75: Should this be pulled up and made into an else if statement? or be changed to a switch block? This if check will always be checked if the first condition is true.

PThorpe92: yeah true either that or a continue after appending, but i like the else if better :+1:

Issue #530, opened 2024-11-23
Admin Dashboard > Student Activity


jtucholski:
Update the Admin Layer 2 Dashboard for “Student Activity” so that it shows the following data

Top of dashboard

- [ ]  Shows number of total courses offered in the system (can be filtered by facility)
- [ ]  Shows number of total students enrolled in at least one course (can be filtered by facility)
- [ ]  Shows total activity time in hours (can be filtered by facility)

Next part of dashboard

- [ ]  Shows a table with a list of the courses offered (can be filtered by facility)
    - Includes # of students in the course
    - Includes the completion rate for the course
    - Includes activity time for the course

corypride:
Questions: 
* Can there be more than one record in the milestone table, with a user_id = 5 and course_id of 5 (how are user_course milestones tracked
* How are courses linked to facilities

corypride:
1 . assumption  student has a course completed they are guaranteed to have a row in the outcomes table. 
- completion rate was determined to be: (total_learners / total_progress_milestones) * 100
- I feel that an individuals completion rate should be based off of the ratio between 2 numbers that have a relationship. Question: explain to me the relationship learner_completion and a course's total_milestones?

corypride:
https://www.notion.so/unlockedlabs/UnlockEd-Glossary-of-Terms-12801ac3401f8000b71cf16c01c37a0e?pvs=6&utm_content=12801ac3-401f-8000-b71c-f16c01c37a0e&utm_campaign=T02TK77NETS&n=slack&n=slack_link_unfurl#13501ac3401f801ab512e7d6651c2611

PR #624, opened 2025-01-07
Learning insights dash

corypride:
## Description of the change

Update the Admin Layer 2 Dashboard for student activity (named Learning Insights) so that it shows the data relative to courses, student enrollments and activity time, all which have the option to be filtered by facility

- **Related issues**: #530

### DIFF
--- a/backend/src/database/activity.go
+++ b/backend/src/database/activity.go
@@ -411,7 +411,7 @@ func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, erro
 	// learning_insights
 	// TODO: add the completion percentage
 	err = db.Table("courses c").
-		Select(`
+		Select(`ROW_NUMBER() over (order by c."name") as idx,
 			c.name AS course_name,
 			COUNT(DISTINCT u.id) AS total_students_enrolled,
 			COALESCE(ROUND(SUM(a.total_time)/3600, 0), 0) AS activity_hours`).
--- a/backend/src/database/activity.go
+++ b/backend/src/database/activity.go
@@ -356,11 +356,8 @@ func (db *DB) GetAdminDashboardInfo(facilityID uint) (models.AdminDashboardJoin,
 
 	return dashboard, nil
 }
-
-func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, error) {
-	var adminLayer2 models.AdminLayer2Join
-
-	// total courses
+func (db *DB) GetTotalCoursesOffered(facilityID *uint) (int, error) {
+	var totalCourses int
 	subQry := db.Table("courses c").
 		Select("COUNT(DISTINCT c.id) as total_courses_offered").
 		Joins("INNER JOIN user_enrollments ue on c.id = ue.course_id").
@@ -368,56 +365,57 @@ func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, erro
 
 	if facilityID != nil {
 		subQry = subQry.Where("u.facility_id = ?", facilityID)
-		subQry = subQry.Group("u.facility_id")
 	}
 
-	err := db.Table("(?) as sub", subQry).
-		Find(&adminLayer2.TotalCoursesOffered).Error
-
+	err := db.Table("(?) as sub", subQry).Find(&totalCourses).Error
 	if err != nil {
-		return adminLayer2, NewDBError(err, "error getting admin dashboard info")
+		return 0, NewDBError(err, "error getting total courses offered")
 	}
+	return totalCourses, nil
+}
 
-	// total_students_enrolled
-	subQry = db.Table("courses c").
-		Select("COUNT(DISTINCT m.user_id) AS students_enrolled, c.name").
+func (db *DB) GetTotalStudentsEnrolled(facilityID *uint) (int, error) {
+	var totalStudents int
+	subQry := db.Table("courses c").
+		Select("COUNT(DISTINCT m.user_id) AS students_enrolled").
 		Joins("LEFT JOIN milestones m ON m.course_id = c.id").
 		Joins("INNER JOIN users u on m.user_id = u.id").
 		Where("u.role = ?", "student")
 
 	if facilityID != nil {
-		subQry = subQry.Where(" u.facility_id = ?", facilityID)
+		subQry = subQry.Where("u.facility_id = ?", facilityID)
 	}
 
-	subQry = subQry.Group("c.name")
-
-	err = db.Table("(?) as sub", subQry).
-		Select("CASE WHEN SUM(students_enrolled) IS NULL THEN 0 ELSE SUM(students_enrolled) end AS students_enrolled").
-		Scan(&adminLayer2.TotalStudentsEnrolled).Error
-
+	err := db.Table("(?) as sub", subQry).
+		Select("CASE WHEN SUM(students_enrolled) IS NULL THEN 0 ELSE SUM(students_enrolled) END AS students_enrolled").
+		Scan(&totalStudents).Error
 	if err != nil {
-		return adminLayer2, NewDBError(err, "error getting total_students_enrolled dashboard info")
+		return 0, NewDBError(err, "error getting total students enrolled")
 	}
+	return totalStudents, nil
+}
 
-	// total_hourly_activity
-	subQry = db.Table("users u").
+func (db *DB) GetTotalHourlyActivity(facilityID *uint) (int, error) {
+	var totalActivity int
+	subQry := db.Table("users u").
 		Select("CASE WHEN SUM(a.total_time) IS NULL THEN 0 ELSE ROUND(SUM(a.total_time)/3600, 0) END AS total_time").
 		Joins("LEFT JOIN activities a ON u.id = a.user_id").
 		Where("u.role = ?", "student")
 
 	if facilityID != nil {
-		subQry = subQry.Where(" u.facility_id = ?", facilityID)
+		subQry = subQry.Where("u.facility_id = ?", facilityID)
 	}
-	subQry = subQry.Group("u.id")
 
-	err = subQry.Find(&adminLayer2.TotalHourlyActivity).Error
-
+	err := subQry.Scan(&totalActivity).Error
 	if err != nil {
-		return adminLayer2, NewDBError(err, "error getting admin dashboard info")
+		return 0, NewDBError(err, "error getting total hourly activity")
 	}
+	return totalActivity, nil
+}
 
-	// learning_insights
-	subQry = db.Table("outcomes o").
+func (db *DB) GetLearningInsights(facilityID *uint) ([]models.LearningInsight, error) {
+	var insights []models.LearningInsight
+	subQry := db.Table("outcomes o").
 		Select("o.course_id, COUNT(o.id) AS outcome_count").
 		Group("o.course_id")
 
@@ -442,12 +440,9 @@ func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, erro
 		subQry2 = subQry2.Where("u.facility_id = ?", facilityID)
 	}
 
-	err = subQry2.Group("c.name, c.total_progress_milestones").
-		Find(&adminLayer2.LearningInsights).Error
-
+	err := subQry2.Group("c.name, c.total_progress_milestones").Find(&insights).Error
 	if err != nil {
-		return adminLayer2, NewDBError(err, "error getting learning insight table info")
+		return nil, NewDBError(err, "error getting learning insights")
 	}
-
-	return adminLayer2, nil
+	return insights, nil
 }
--- a/backend/src/database/activity.go
+++ b/backend/src/database/activity.go
@@ -393,7 +393,7 @@ func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, erro
 
 	err = db.Table("(?) as sub", subQry).
 		Select("CASE WHEN SUM(students_enrolled) IS NULL THEN 0 ELSE SUM(students_enrolled) end AS students_enrolled").
-		Scan(&adminLayer2.TotalStudentsEnrolled).Debug().Error
+		Scan(&adminLayer2.TotalStudentsEnrolled).Error
 
 	if err != nil {
 		return adminLayer2, NewDBError(err, "error getting total_students_enrolled dashboard info")
--- a/backend/src/database/activity.go
+++ b/backend/src/database/activity.go
@@ -361,17 +361,25 @@ func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, erro
 	var adminLayer2 models.AdminLayer2Join
 
 	// total courses
-	// TODO: find the link between coures and facility
-	err := db.Table("courses c").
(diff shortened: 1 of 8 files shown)


frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,197 @@
...
+                                            Name: 'order=asc&order_by=course_name',
+                                            'Completed Only': 'completed',
+                                            'In Progress Only': 'in_progress',
+                                            'Total time':
+                                                'order=desc&order_by=total_time'
+                                            //"Grade": "order=asc&order_by=grade",
+                                        }}
+                                    /> */}
+                                </div>
+                                <table className="w-full mt-4">
```
calisio, line 138: I think you use the class "card" twice in here, which is why there is one card within another. I would remove one of them. I would also remove the div `<div className="flex flex-row gap-12">` since this was there because there were two divs within this larger card. You should be able to condense this all down into 
```
                   <div className="card card-row-padding mb-30">
                        <h2 className="mt-2">Course Metrics</h2>
```
UNLESS you plan on adding in that dropdown control back in, then it would look like 
```
                   <div className="card card-row-padding mb-30">
                       <div className="flex flex-row justify-between"> 
                            <h2 className="mt-2">Course Metrics</h2>
                            //dropdown control component
```

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,197 @@
...
+                                                    <tr
+                                                        className="flex flex-row justify-between body-small items-center"
+                                                        key={index}
+                                                    >
+                                                        <td className="w-1/2">
+                                                            {
+                                                                insight.course_name
+                                                            }
+                                                        </td>
+                                                        <td className="w-1/5">
```
calisio, line 171: I would modify this to use `table-2`, and also would remove the widths that you set and instead use `grid-cols-4` and `justify-self-start` & `justify-self-end` like we do in other pages. For an example you can look to `AdminManagement.tsx` on line 188

frontend/src/Pages/AdminLayer2.tsx
calisio, line 1: Not sure what the stemming issue is, but I was messing with the frontend to see if the stats match up, and when I switched to my third facility (Potosi Correctional), it said 0 courses offered and 0 enrollments but then the course metrics table was filled out with the same 10 courses as the previous facility. It could just be my seeded data acting weird, but worth it to check out. See images below
Screenshot 2025-01-07 at 11 51 26 AM

Screenshot 2025-01-07 at 11 49 10 AM

Could it be that the "Course Metrics" table is listing all courses in all facilities with their enrollments? Do we want that table to be updated according to the selected facility, or do we want it to display all courses across all facilities? If we would like it to consistently display all courses in all facilities (which i believe is the behavior it is doing right now), I think it would be worth having a conversation about how we can display the facility filter in a less confusing spot, since right now the user probably assumes that that filter will update the entire page including the course metrics table. Let me know your thoughts

corypride:
Table is not filtered by facility yet

backend/src/database/activity.go
```diff
@@ -356,3 +356,98 @@ func (db *DB) GetAdminDashboardInfo(facilityID uint) (models.AdminDashboardJoin,
 
 	return dashboard, nil
 }
+
+func (db *DB) GetAdminLayer2Info(facilityID *uint) (models.AdminLayer2Join, error) {
```
calisio, line 360: @PThorpe92 should facilityID be a pointer or just a regular uint?

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
                     </div>
-                    <div className="grid grid-cols-3 gap-6">
+                    <div className="grid grid-cols-3 gap-4 mb-6">
                         <StatsCard
-                            title={'ACTIVE USERS'}
-                            number={`${activityData.weekly_active_users}`}
-                            label={'students'}
+                            title="Total Courses Offered"
+                            number={layer2_metrics.total_courses_offered.toString()}
+                            label=""
```
calisio, line 107: add in a label "courses"

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
                         />
                         <StatsCard
-                            title={'AVG DAILY ACTIVITY'}
-                            number={avgActivity.number.toString()}
-                            label={avgActivity.label}
+                            title="Total Students Enrolled"
+                            number={layer2_metrics.total_students_enrolled.toString()}
+                            label={''}
```
calisio, line 112: add in label "students"

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
-                                            );
-                                        }
-                                    )}
-                                </tbody>
-                            </table>
-                        </div>
+                    <div className="card card-row-padding mb-30">
+                        <table className="table-2 mb-4">
+                            <thead>
+                                <tr className="grid-col-4 px-4">
```
calisio, line 123: remove px-4 (just looks a little better so it lines up with the sides)

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
+                                    ) => {
+                                        return (
+                                            <tr
+                                                className="grid-cols-4 justify-items-center"
+                                                key={index}
+                                            >
+                                                <td className="justify-self-start">
+                                                    {insight.course_name}
+                                                </td>
+                                                <td className="w-1/5">
```
calisio, line 148: remove w-1/5, and same with the next <td>

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
+        void mutate();
+    }, [facility, resetCache]);
+
+    useEffect(() => {
+        const fetchFacilities = async () => {
+            const response = await API.get<Facility>('facilities');
+            setFacilities(response.data as Facility[]);
+        };
+        void fetchFacilities();
+    }, []);
```
calisio, line 41: rather than fetching the facilities on this use effect, i would just use SWR:
`const { data: facilities } = useSWR<Facility[]>('/api/facilities');`
You can then delete `const [facilities, setFacilities] = useState<Facility[]>();` on line 19. You will need to update line 91, where you map through the facilities, to now map through `facilities.data` (that is just the way that SWR returns the data fetched)
You can also probably remove the `useEffect()` on line 31-33 since SWR will handle this. Instead, move `void mutate()` directly on the "Refresh Data" button, so it would look like this on line 71 
`onClick={() => void mutate()}`. You will also be able to delete line 21. 
This will help condense this page down a lot. I would also delete any commented out code (we can add it back in later if we need it but makes the code feel messy)

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
-                    <div className="card h-[240px]">
-                        <h2 className="card-h-padding">
-                            Overall Platform Engagement
-                        </h2>
-                        <ActivityChart data={activityData.monthly_activity} />
+        <div className="p-8">
+            {error && <div>Error loading data</div>}
+            {!data || (isLoading && <div>Loading...</div>)}
+            {data && layer2_metrics && (
+                <>
```
calisio, line 67: also, could we add in a h1 title to this page like the other pages that say "Learning Insights"

backend/src/database/activity.go
calisio, line 1: I will let @PThorpe92 review these queries because he is better at best practices

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -1,141 +1,173 @@
...
+                                        >
+                                            {facility.name}
+                                        </option>
+                                    ))}
+                                </select>
+                            </div>
+                        </div>
                     </div>
-                    <div className="grid grid-cols-3 gap-6">
+                    <div className="grid grid-cols-3 gap-4 mb-6">
```
calisio, line 103: lets change this to be on one row and alter the default selection so it is more clear:
```
<>
                    <div className="flex flex-row justify-between mb-6">
                        <select
                            id="facility"
                            className="select select-bordered w-full max-w-xs"
                            value={facility}
                            onChange={(e) => setFacility(e.target.value)}
                        >
                            <option key={'all'} value={'all'}>
                                All Facilities
                            </option>
                            {errorFacilitiesFetch ? (
                                <div>Error fetching facilities</div>
                            ) : (
                                facilities?.data?.map((facility: Facility) => (
                                    <option
                                        key={facility.id}
                                        value={facility.id}
                                    >
                                        {facility.name}
                                    </option>
                                ))
                            )}
                        </select>
                        <button
                            className="button"
                            onClick={() => void mutate()}
                        >
                            Refresh Data
                        </button>
                    </div>
                    <div className="grid grid-cols-3 gap-4 mb-6">
                        <StatsCard
```

backend/src/database/activity.go
calisio: Ohhh good to know. I will turn to Preston to review the queries and leave comments, since I think he probably knows best.

jtucholski: Can we break each of the calls (for total courses, total students enrolled, etc.) into separate functions and have `dashboard.go` call each of them to build up the response? I'm looking at `dashboard.handleLoginMetrics` as a reference since that has to aggregate a number of different metrics together.

jtucholski (commented):
I added some comments about splitting up the GetAdminLayer2Info function call and leaving the logic to aggregate it in the handler. This is similar to what we also arrived at for the Operational Insights.

backend/src/database/activity.go
```diff
@@ -356,3 +356,98 @@ func (db *DB) GetAdminDashboardInfo(facilityID uint) (models.AdminDashboardJoin,
...
+	var adminLayer2 models.AdminLayer2Join
+
+	// total courses
+	subQry := db.Table("courses c").
+		Select("COUNT(DISTINCT c.id) as total_courses_offered").
+		Joins("INNER JOIN user_enrollments ue on c.id = ue.course_id").
+		Joins("INNER JOIN users u on ue.user_id = u.id")
```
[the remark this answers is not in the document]
corypride: Yes and Yes

backend/src/database/activity.go
```diff
@@ -356,3 +356,93 @@ func (db *DB) GetAdminDashboardInfo(facilityID uint) (models.AdminDashboardJoin,
...
+
+	if facilityID != nil {
+		subQry = subQry.Where("u.facility_id = ?", facilityID)
+	}
+
+	err := db.Table("(?) as sub", subQry).Find(&totalCourses).Error
```
PThorpe92, line 370: this doesn't need to be a subquery as nothing else is being passed in or added

backend/src/handlers/dashboard.go
```diff
@@ -48,6 +49,60 @@ func (srv *Server) handleAdminDashboard(w http.ResponseWriter, r *http.Request,
 	return writeJsonResponse(w, http.StatusOK, adminDashboard)
 }
 
+func (srv *Server) handleAdminLayer2(w http.ResponseWriter, r *http.Request, log sLog) error {
+	facility := r.URL.Query().Get("facility")
+	claims := r.Context().Value(ClaimsKey).(*Claims)
+	var facilityId *uint
+	// Logic goes here for facility
```
PThorpe92, line 56: -_-

backend/src/database/activity.go
corypride:
```func (db *DB) GetTotalCoursesOffered(facilityID *uint) (int, error) {
	var totalCourses int
	subQry := db.Table("courses c").
		Select("COUNT(DISTINCT c.id) as total_courses_offered").
		Joins("INNER JOIN user_enrollments ue on c.id = ue.course_id").
		Joins("INNER JOIN users u on ue.user_id = u.id")

	if facilityID != nil {
		subQry = subQry.Where("u.facility_id = ?", facilityID)
	}

	err := subQry.Find(&totalCourses).Error
	if err != nil {
		return 0, NewDBError(err, "error getting total courses offered")
	}
	return totalCourses, nil
}
```
Is this what you mean?

backend/src/handlers/dashboard.go
corypride: I interpreted this as git rid of  the comment

backend/src/database/activity.go
```diff
@@ -356,3 +356,93 @@ func (db *DB) GetAdminDashboardInfo(facilityID uint) (models.AdminDashboardJoin,
...
+		Select("COUNT(DISTINCT m.user_id) AS students_enrolled").
+		Joins("LEFT JOIN milestones m ON m.course_id = c.id").
+		Joins("INNER JOIN users u on m.user_id = u.id").
+		Where("u.role = ?", "student")
+
+	if facilityID != nil {
+		subQry = subQry.Where("u.facility_id = ?", facilityID)
+	}
+
+	err := db.Table("(?) as sub", subQry).
```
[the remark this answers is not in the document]
corypride: Since I made this change for student_enrollments in this query, should the total_students_enrolled metric, in the learning_insights query be determined in the same way?

func (db *DB) GetLearningInsights(facilityID *uint) ([]models.LearningInsight, error) {
	var insights []models.LearningInsight
	subQry := db.Table("outcomes o").
		Select("o.course_id, COUNT(o.id) AS outcome_count").
		Group("o.course_id")

	subQry2 := db.Table("courses c").
		Select(`
			c.name AS course_name,
			COUNT(DISTINCT u.id) AS total_students_enrolled,
			CASE 
				WHEN MAX(subqry.outcome_count) > 0 THEN  
					COUNT(DISTINCT u.id) / NULLIF(CAST(MAX(c.total_progress_milestones) AS float), 0) * 100.0
				ELSE 0
			END AS completion_rate,
			COALESCE(ROUND(SUM(a.total_time) / 3600, 0), 0) AS activity_hours
		`).
		Joins("LEFT JOIN milestones m ON m.course_id = c.id").
		Joins("LEFT JOIN users u ON m.user_id = u.id").
		Joins("LEFT JOIN activities a ON u.id = a.user_id").
		Joins("INNER JOIN (?) AS subqry ON m.course_id = subqry.course_id", subQry).
		Where("u.role = ?", "student")

	if facilityID != nil {
		subQry2 = subQry2.Where("u.facility_id = ?", facilityID)
	}

	err := subQry2.Group("c.name, c.total_progress_milestones").Find(&insights).Error
	if err != nil {
		return nil, NewDBError(err, "error getting learning insights")
	}
	return insights, nil
}

backend/src/database/activity.go
```diff
@@ -356,3 +356,90 @@ func (db *DB) GetAdminDashboardInfo(facilityID uint) (models.AdminDashboardJoin,
...
+	if err != nil {
+		return 0, NewDBError(err, "error getting total courses offered")
+	}
+	return totalCourses, nil
+}
+
+func (db *DB) GetTotalStudentsEnrolled(facilityID *uint) (int, error) {
+	var totalStudents int
+	query := db.Table("user_enrollments ue").
+		Select("COUNT(DISTINCT ue.user_id) AS students_enrolled").
```
PThorpe92, line 380: Sorry i got distracted i should have checked this earlier. We could emit this join because students would be the only ones in that table/enrolled. other than that this is ready to merge :+1:

PThorpe92 (approved):

PR #630, opened 2025-01-07
fix: add tooltips and line charts to operational insight

### LINKED ISSUE #616: UI: Improvements to the Operational Insights page
The UI and responsiveness of the Operational Insights page need improvement.

**1. Pie Chart Issue:**
- On smaller laptop screens, the pie chart is pushed out of the container (see image below)
- The pie chart contains redundant data that can be determined by Total Users, Active Users, and Inactive Users. It may not be necessary.

**2. Peak Login Times Chart:**
- Consider replacing the pie chart with a line chart to display Peak Login Times.
- The x-axis should represent the hours of the day, and the y-axis should represent the number of logins per hour.

**3. Tooltip Readability:**
- In dark mode, the tooltip popup is difficult to read due to a white background with gray text.
- In light mode, the tooltip is easier to read with a white background and black text.
- Remove the timestamp from the tooltip since the data represents a period and only show the number of logins.

**4. Terminology Consistency:**
- Update the side nav from "Students" to "Residents"

**5. Update to Report Data:**
- Remove the logins per day metric and add a "New Admins Added" value
- The top data points (number of users, active, inactive) should include all users.

## Pie Chart Image

Image

CK-7vn:
## Description of the change

This fixes the wordage on the operational insights dashboard for the cards, removed logins per day, and both the pie chart, and bar graph in the peak login times. Total users was changed to include admins, so now users include the totality of the users. Users = total people using the platform, Admins = everyone except residents (students per user roles), Residents = anyone in student role. 

The tooltip was changed to reflect the theme so there is no longer issues with the coloring as well. 

In the StatsCard's ternary's were added to check the count of the data being returned/rendered so that the pluralism was correct as well.  

Changes "Students" in navbar to "Residents" as well. 
- **Related issues**: Closes #616 

## Screenshot(s)
image

image

When hovering the tooltip displays the time in 24hr time
image


backend/src/database/users.go
```diff
@@ -239,17 +239,25 @@ func (db *DB) GetNumberOfActiveUsersForTimePeriod(active bool, days int, facilit
...
 }
 
-func (db *DB) NewUsersInTimePeriod(days int, facilityId *uint) (int64, error) {
-	var count int64
+func (db *DB) NewUsersInTimePeriod(days int, facilityId *uint) (int64, int64, error) {
+	var admin_count int64
+	var user_count int64
 	daysAgo := time.Now().AddDate(0, 0, -days)
-	tx := db.Model(&models.User{}).Where("created_at >= ? AND role NOT IN ('system_admin', 'admin')", daysAgo)
+	tx := db.Model(&models.User{}).Debug().Where("created_at >= ? AND role NOT IN ('system_admin', 'admin')", daysAgo)
```
jtucholski, line 246: Did you mean to leave `.Debug()` in here?

backend/src/database/users.go
```diff
@@ -239,17 +239,25 @@ func (db *DB) GetNumberOfActiveUsersForTimePeriod(active bool, days int, facilit
...
 	if facilityId != nil {
 		tx = tx.Where("facility_id = ?", *facilityId)
 	}
-	if err := tx.Count(&count).Error; err != nil {
-		return 0, newGetRecordsDBError(err, "users")
+	if err := tx.Count(&user_count).Error; err != nil {
+		return 0, 0, newGetRecordsDBError(err, "users")
 	}
-	return count, nil
+	tx = db.Model(&models.User{}).Debug().Where("created_at >= ? AND role NOT IN ('student', 'system_admin')", daysAgo)
```
jtucholski, line 253: Same as the previous comment. Did you mean to leave `.Debug()` here?

backend/src/handlers/dashboard.go
```diff
@@ -97,40 +97,44 @@ func (srv *Server) handleLoginMetrics(w http.ResponseWriter, r *http.Request, lo
 		if err != nil {
 			return newDatabaseServiceError(err)
 		}
-		totalUsers, err := srv.Db.GetTotalUsers(facilityId)
+		totalStudents, totalAdmins, err := srv.Db.GetTotalUsers(facilityId)
```
jtucholski, line 100: This is going to be challenging since we're changing vernacular but we should probably use `totalResidents` instead of `totalStudents`.

backend/src/handlers/dashboard.go
```diff
@@ -97,40 +97,44 @@ func (srv *Server) handleLoginMetrics(w http.ResponseWriter, r *http.Request, lo
...
 			return newDatabaseServiceError(err)
 		}
 		loginTimes, err := srv.Db.GetLoginActivity(days, facilityId)
 		if err != nil {
 			return newDatabaseServiceError(err)
 		}
-		if totalUsers == 0 {
-			totalUsers = 1
+		if totalStudents+totalAdmins == 0 {
+			totalStudents = 1
```
jtucholski, line 113: Is this to account for SuperAdmin?

backend/src/handlers/dashboard.go
```diff
@@ -97,40 +97,44 @@ func (srv *Server) handleLoginMetrics(w http.ResponseWriter, r *http.Request, lo
...
+			NewAdminsAdded   int64                  `json:"new_admins_added"`
 			PeakLoginTimes   []models.LoginActivity `json:"peak_login_times"`
 		}{
 			ActiveUsers:      acitveUsers,
 			TotalLogins:      totalLogins,
 			LoginsPerDay:     totalLogins / int64(days),
-			PercentActive:    acitveUsers / totalUsers * 100,
-			PercentInactive:  100 - (acitveUsers / totalUsers * 100),
-			TotalUsers:       totalUsers,
+			PercentActive:    acitveUsers / totalStudents * 100,
```
jtucholski, line 131: Can you fix the spelling for `acitveUsers` with this PR too?

frontend/src/Components/PeakLoginTimesChart.tsx
jtucholski, line 1: Are we still using this chart anymore? We can probably remove it from the codebase if not.

frontend/src/Components/OperationalInsightsCharts.tsx
jtucholski, line 1: I know it's not in the original issue but while you are in here can you update the call to fetch facilities to `useSWR`?

frontend/src/Components/OperationalInsightsCharts.tsx
```diff
@@ -35,6 +34,9 @@ const OperationalInsights = () => {
 
     const metrics = data?.data;
 
+    const totalUsers =
```
jtucholski, line 37: I'm getting a typescript error here because metrics is either `LoginMetrics | undefined`. Is there a way to be clear this isn't undefined?

frontend/src/Components/OperationalInsightsCharts.tsx
```diff
@@ -95,58 +97,68 @@ const OperationalInsights = () => {
...
+                            label={
+                                metrics.total_logins === 1 ? 'Login' : 'Logins'
+                            }
                         />
                         <StatsCard
-                            title="Logins per Day"
-                            number={metrics.logins_per_day.toString()}
-                            label="logins/day"
+                            title="New Admins Added"
+                            number={metrics.new_admins_added.toString()}
```
jtucholski, line 131: Same here, I get another TS error but not with `new_residents_added`.

backend/src/handlers/dashboard.go
```diff
@@ -97,40 +97,44 @@ func (srv *Server) handleLoginMetrics(w http.ResponseWriter, r *http.Request, lo
...
+			totalStudents = 1
 		}
 		activityMetrics := struct {
 			ActiveUsers      int64                  `json:"active_users"`
 			TotalLogins      int64                  `json:"total_logins"`
 			LoginsPerDay     int64                  `json:"logins_per_day"`
 			PercentActive    int64                  `json:"percent_active"`
 			PercentInactive  int64                  `json:"percent_inactive"`
-			TotalUsers       int64                  `json:"total_users"`
+			TotalStudents    int64                  `json:"total_students"`
```
jtucholski, line 121: Can we change this to be `TotalResidents` and `total_residents` now? It'll need an update in `common.ts` if we do but it appears the only areas it affects is the OperationalInsightsCharts.

frontend/src/common.ts
```diff
@@ -366,9 +366,11 @@ export interface LoginMetrics {
     logins_per_day: number;
     percent_active: number;
     percent_inactive: number;
-    total_users: number;
+    total_students: number;
```
jtucholski, line 369: I added a comment on `dashboard.go` about updating this to be `total_residents` while the thought is fresh. If you can update it, that'd be great.

jtucholski (changes requested):

backend/src/handlers/dashboard.go
CK-7vn: Now that I think of it, yes, and no, in theory, A SuperAdmin will be the only person that will ever see that data at potentially 0, because in all actuality, given that totalAdmins is not accounting for SuperAdmins, if a facility admin is looking at that data on the frontend, then their account IS explicitly being accounted for. But, I left that in there, just updated it because it was there prior.

CK-7vn:
A lot of updates here.

-- Updated the facilities API call to use SWR.
-- Added labels on the x and y axis for the Peak login times graph
-- Changed it so that based on screen size the ticks on the x axis will change so that the ticks don't bombard the x axis, and makes it easier to see. 
-- Changed the labels in the StatsCard with a slice to uppercase first letter
-- Changed some "students" to residents in backend, and fixed a typo.
-- Cleaned out a couple left over Debug's in queries
-- Added an information tool tip for the StatsCards
-- Moved the refresh button

(Small, key in on the x axis labels)
image

3 hr x axis 
image

2 hr x axis 
image

image

image

image

image

jtucholski (changes requested):
Two small changes
- Try changing some of the step values for the chart to have even greater intervals. There's a point still where the values in there are squeezed together. 
- Look into using the DaisyUI or Tailwind Tooltips that we use throughout the app so you don't need to fight the DOM in the ToolTipIcon component.

jtucholski (approved):

jtucholski (approved):
Awesome work! This looks really good. There may be a few followup enhancements for the messaging but if that's the case, those will be logged separately.

PR #632, opened 2025-01-07
fix: add titles for pages

### LINKED ISSUE #610: UX: Update Page Titles so that it includes the name of the page.
The `<title />`  tag for every page in the application is **UnlockEd**.

This issue involves updating the tag so that it includes the name of the page.

Examples

| Page Name          | Page Title                    |
|--------------------|-------------------------------|
| Learning Insights  | Learning Insights • UnlockEd  |
| Knowledge Insights | Knowledge Insights • UnlockEd |

CK-7vn:
## Description of the change
This PR adds dynamic title changes for pages. In the library viewer the TitleManager checks to see if we are in the library viewer, if we are it gives an explicit 5 second timeout to wait for the iFrame to load the content, after that pulls the title tag from inside of the document and then renders it as the Title for the parent DOM. 

- **Related issues**: Closes #610 

## Screenshot(s)
1'st layer 
image

2'nd nested layer inside of the library viewer 
image

5 layers deep
image

image

## Additional context
There is an issue in which if the user clicks on a link inside of the iFrame that is cross-site, the title will not change, and there will be errors in the console, but, this does not effect any of the functionality outside of that.

### DIFF
--- a/frontend/src/Components/TitleManager.tsx
+++ b/frontend/src/Components/TitleManager.tsx
@@ -1,0 +1,48 @@
+import { useMatches, useLocation } from 'react-router-dom';
+import { useEffect } from 'react';
+import { TitleHandler } from '@/common';
+
+export function TitleManager() {
+    const matches = useMatches();
+    const location = useLocation();
+
+    useEffect(() => {
+        const updateTitle = () => {
+            const isLibraryViewer =
+                location.pathname.includes('/viewer/libraries/');
+
+            if (isLibraryViewer) {
+                setTimeout(() => {
+                    const titleInterval = setInterval(() => {
+                        const libraryViewer = document.getElementById(
+                            'library-viewer-iframe'
+                        ) as HTMLIFrameElement;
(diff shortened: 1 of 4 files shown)


calisio (approved):
This PR looks good to me, but I am a little confused on the "additional context" comment. Is there anything we need to do to fix these errors?
Also, just wanted to confirm we are just updating the page titles and not adding in the extra page titles in the top breadcrumbs nav. If we are just sticking to the tab page titles, I can go ahead and merge!

CK-7vn:
> This PR looks good to me, but I am a little confused on the "additional context" comment. Is there anything we need to do to fix these errors? Also, just wanted to confirm we are just updating the page titles and not adding in the extra page titles in the top breadcrumbs nav. If we are just sticking to the tab page titles, I can go ahead and merge!

Not updating breadcrumbs as to my knowledge, Preston said to ignore the errors for now, they don't cause any issues, and it's only once the links lead somewhere external, only thing that will happen, is at that point the title will stop updating, that's the worst thing that happens!

Issue #613, opened 2025-01-06
UX: Improve Sorting Options and Clarity in Knowledge Center


jtucholski:
Each of the tabs in the Knowledge Center has a different way of sorting. 

**Libraries** uses a drop-down control and sorts by: Most Favorited, Least favorited, Newest, and Oldest
**Videos** uses a drop-down control and sorts by: Date added (Desc), Date added (Asc), Title (Asc), Title (Desc)
**Helpful** Links uses these buttons that toggle the sort and sort by: Date added (Desc), Date added (Asc), Title (Asc), Title (Desc).

**Expected:**

All three tabs should use the drop-down control and include the following values for sorting:
* Most Popular: This shows libraries, videos, and helpful links in descending order by number of favorites. 
* Title (A to Z)
* Title (Z to A)
* Date Added (Newest First)
* Date Added (Oldest First)

PR #649, opened 2025-01-13
Sorting options and clarity in the Knowledge Center

corypride:
## Description of the change
Make uniform the sorting options across the Knowledge center. 
All three tabs should use the drop-down control and include the following values for sorting:

Most Popular
Title (A to Z)
Title (Z to A)
Date Added (Newest First)
Date Added (Oldest First)

- **Related issues**: #613 

## Screenshot(s)

https://github.com/user-attachments/assets/50c36c06-4125-4bd7-bf72-8c8f61f88539

https://github.com/user-attachments/assets/afecf069-23cd-4839-9e50-afa0807597c6

### DIFF
--- a/backend/src/database/helpful_links.go
+++ b/backend/src/database/helpful_links.go
@@ -3,6 +3,7 @@ package database
 import (
 	"UnlockEdv2/src/models"
 	"fmt"
+	"strings"
 )
 
 type HelpfulLinkResp struct {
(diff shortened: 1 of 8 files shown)


carddev81 (changes requested):
When navigating to the Libraries tab I noticed that by default the sort option selected is 'Most Popular' but the default order of the libraries listed is not ordered by most popular. The same goes for the other tabs within the Knowledge Center Management.

carddev81 (approved):

PR #676, opened 2025-01-18
Kiwix search implementation

### LINKED ISSUE #639: Task: Implement Kiwix Search
This task list is incomplete and will be updated to reflect work done on implementing the search feature into UnlockEdV2 for Kiwix

# Steps for Implementing Kiwix Search
- [ ] Setup staging for testing Kiwix search
    - [ ] Add domain name for kiwix in AWS (bind name to load balancer)
    - [ ] Add listener rule (condition) to load balancer for kiwix 
    - [ ] Add helm configuration for deploying kiwix server within staging
    - [ ] Create kubernetes deployment file for kiwix server

carddev81:
## Description of the change

Initial design and implementation of doing full text searching across kiwix libraries.  This change required deploying our own managed kiwix server within the 'staging' development environment so that we could manage Kiwix's API to perform full text searching.

### Things to note
- A Kiwix server was set up to provide control and customization over our search functionalities. 
- Libraries will need to be manually downloaded to the Kiwix server as part of the setup and maintenance process. 
- We need to ensure sufficient disk space is allocated on the server for storing library zim files--could grow significantly depending on the number of zim files downloaded.

- **Related issues**: "Closes #639 ".

## Screenshot(s)
SearchBoxAndDropdown
SearchSingleLibrary
SearchAcrossLibraries
ColorTheme

## Additional context
No additional context.

### DIFF
--- a/backend/seeder/main.go
+++ b/backend/seeder/main.go
@@ -91,7 +91,7 @@ func seedTestData(db *gorm.DB) {
 		}
 	}
 	kiwix := models.OpenContentProvider{
-		Url:              "https://library.kiwix.org",
+		Url:              "https://kiwix.staging.unlockedlabs.xyz",
 		Title:            models.Kiwix,
 		ThumbnailUrl:     "https://images.fineartamerica.com/images/artworkimages/mediumlarge/3/llamas-wearing-party-hats-in-a-circle-looking-down-john-daniels.jpg",
 		CurrentlyEnabled: true,
@@ -109,17 +109,17 @@ func seedTestData(db *gorm.DB) {
 			Language:              models.StringPtr("eng,spa,ara"),
 			Description:           models.StringPtr("A collection of TED videos about ted connects"),
 			Url:                   "/content/ted_mul_ted-connects_2024-08",
-			ThumbnailUrl:          models.StringPtr("/catalog/v2/illustration/67440563-a62b-fabe-415c-4c3ee4546f78/?size=48"),
+			ThumbnailUrl:          models.StringPtr("/kiwix.jpg"),
 			VisibilityStatus:      true,
 		},
 		{
 			OpenContentProviderID: kiwix.ID,
-			ExternalID:            models.StringPtr("urn:uuid:84812c13-fa65-feb7-c206-4f22cc2e0f9a"),
+			ExternalID:            models.StringPtr("urn:uuid:93321718-5228-676d-7e95-14bbe88fa38c"),
 			Title:                 "Python Documentation",
 			Language:              models.StringPtr("eng"),
 			Description:           models.StringPtr("All documentation for Python"),
-			Url:                   "/content/docs.python.org_en_2024-09",
-			ThumbnailUrl:          models.StringPtr("/catalog/v2/illustration/84812c13-fa65-feb7-c206-4f22cc2e0f9a/?size=48"),
+			Url:                   "/content/docs.python.org_en_2025-01",
+			ThumbnailUrl:          models.StringPtr("/kiwix.jpg"),
 			VisibilityStatus:      true,
 		},
 		{
@@ -129,7 +129,7 @@ func seedTestData(db *gorm.DB) {
 			Language:              models.StringPtr("eng"),
 			Description:           models.StringPtr("The Canadian financial wiki"),
 			Url:                   "/content/finiki_en_all_maxi_2024-06",
-			ThumbnailUrl:          models.StringPtr("/catalog/v2/illustration/19e6fe12-09a9-0a38-5be4-71c0eba0a72d/?size=48"),
+			ThumbnailUrl:          models.StringPtr("/kiwix.jpg"),
 			VisibilityStatus:      true,
 		}}
 	for idx := range kiwixLibraries {
--- a/backend/src/database/libraries.go
+++ b/backend/src/database/libraries.go
@@ -13,7 +13,16 @@ type LibraryResponse struct {
 	IsFavorited bool `json:"is_favorited"`
 }
 
-func (db *DB) GetAllLibraries(page, perPage, days int, userId, facilityId uint, visibility, orderBy, search string, isAdmin bool) (int64, []LibraryResponse, error) {
+// Retrieves either a paginated list of libraries or all libraries based upon the given parameters.
+// page - the page number for pagination
+// perPage - the number of libraries to display on page
+// userId - the userId for which libraries to display for
+// facilityId - the facility id of where the libraries were favorited
+// visibility - can either be featured, visible, hidden, or all
+// orderBy - the order in which the results are returned
+// isAdmin - true or false on whether the user is an administrator used to determine how to retrieve featured libraries
+// all - true or false on whether or not to return all libraries without pagination
+func (db *DB) GetAllLibraries(page, perPage, days int, userId, facilityId uint, visibility, orderBy, search string, isAdmin, all bool) (int64, []LibraryResponse, error) {
 	var (
 		total    int64
 		criteria string
@@ -81,14 +90,15 @@ func (db *DB) GetAllLibraries(page, perPage, days int, userId, facilityId uint, 
 				AND f.open_content_provider_id = libraries.open_content_provider_id`)
 		}
 		tx = tx.Group("libraries.id").Order("favorite_count DESC")
-
 	default:
 		tx = tx.Order(orderBy)
 	}
-	if err := tx.Limit(perPage).Offset(calcOffset(page, perPage)).Find(&libraries).Error; err != nil {
+	if !all {
+		tx = tx.Limit(perPage).Offset(calcOffset(page, perPage))
+	}
+	if err := tx.Find(&libraries).Error; err != nil {
 		return 0, nil, newGetRecordsDBError(err, "libraries")
 	}
-
 	return total, libraries, nil
 }
 
@@ -99,6 +109,19 @@ func (db *DB) GetLibraryByID(id int) (*models.Library, error) {
 		return nil, newNotFoundDBError(err, "libraries")
 	}
 	return &library, nil
+}
+
+func (db *DB) GetLibrariesByIDs(ids []int) ([]models.Library, error) {
+	var libraries []models.Library
+	tx := db.Preload("OpenContentProvider").Where("id in ?", ids)
+	if len(ids) > 1 {
+		tx.Where("language = 'eng'")
+	}
+	if err := tx.Find(&libraries).Error; err != nil {
+		log.Errorln("unable to find libraries with these IDs")
+		return nil, newNotFoundDBError(err, "libraries")
+	}
+	return libraries, nil
 }
 
 func (db *DB) ToggleVisibilityAndRetrieveLibrary(id int) (*models.Library, error) {
--- a/backend/src/handlers/libraries_handler.go
+++ b/backend/src/handlers/libraries_handler.go
@@ -3,24 +3,41 @@ package handlers
 import (
 	"UnlockEdv2/src/models"
 	"encoding/json"
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"io"
 	"net/http"
+	"net/url"
+	"path"
 	"strconv"
+	"strings"
 )
 
 func (srv *Server) registerLibraryRoutes() []routeDef {
 	axx := models.Feature(models.OpenContentAccess)
 	return []routeDef{
 		{"GET /api/libraries", srv.handleIndexLibraries, false, axx},
+		{"GET /api/libraries/search", srv.handleSearchLibraries, false, axx},
 		{"GET /api/libraries/{id}", srv.handleGetLibrary, false, axx},
 		{"PUT /api/libraries/{id}/toggle", srv.handleToggleLibraryVisibility, true, axx},
 		{"PUT /api/libraries/{id}/favorite", srv.handleToggleFavoriteLibrary, false, axx},
 	}
 }
 
+// Retrieves either a paginated list of libraries or all libraries based upon the HTTP request parameters.
+// Query Parameters:
+// page - the page number for pagination
+// perPage - the number of libraries to display on page
+// search - the title or patial title of the libraries to search for
+// order_by - (title|created_at|most_popular) the order in which the results are returned
+// visibility - can either be featured, visible, hidden, or all
+// all - true or false on whether or not to return all libraries without pagination
 func (srv *Server) handleIndexLibraries(w http.ResponseWriter, r *http.Request, log sLog) error {
 	page, perPage := srv.getPaginationInfo(r)
 	search := r.URL.Query().Get("search")
 	orderBy := r.URL.Query().Get("order_by")
+	all := r.URL.Query().Get("all") == "true"
 	days, err := strconv.Atoi(r.URL.Query().Get("days"))
 	if err != nil {
 		days = -1
@@ -34,7 +51,7 @@ func (srv *Server) handleIndexLibraries(w http.ResponseWriter, r *http.Request, 
 		showHidden = r.URL.Query().Get("visibility")
 	}
 	claims := r.Context().Value(ClaimsKey).(*Claims)
-	total, libraries, err := srv.Db.GetAllLibraries(page, perPage, days, claims.UserID, claims.FacilityID, showHidden, orderBy, search, claims.isAdmin())
+	total, libraries, err := srv.Db.GetAllLibraries(page, perPage, days, claims.UserID, claims.FacilityID, showHidden, orderBy, search, claims.isAdmin(), all)
 	if err != nil {
 		return newDatabaseServiceError(err)
 	}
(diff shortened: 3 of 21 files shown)


backend/src/models/library.go
```diff
@@ -35,3 +41,129 @@ type LibraryProxyPO struct {
...
+	Items        []KiwixItem `json:"items"`
+}
+
+type KiwixItem struct {
+	Title        string `json:"title"`
+	Link         string `json:"link"`
+	Description  string `json:"description"`
+	Book         string `json:"book"`
+	ThumbnailUrl string `json:"thumbnail_url"`
+	WordCount    string `json:"word_count"`
```
calisio, line 96: can you alter this so this struct is deleted and anywhere you use this struct instead modify to return a Library?

backend/src/handlers/libraries_handler.go
```diff
@@ -55,6 +64,81 @@ func (srv *Server) handleGetLibrary(w http.ResponseWriter, r *http.Request, log
 	return writeJsonResponse(w, http.StatusOK, library)
 }
 
+func (srv *Server) handleSearchLibraries(w http.ResponseWriter, r *http.Request, log sLog) error {
+	page, perPage := srv.getPaginationInfo(r)
+	pattern := r.URL.Query().Get("pattern")
```
calisio, line 69: is this search? to make this consistent with the terms used in the backend, could we make this query param be "search"

backend/src/database/libraries.go
```diff
@@ -101,6 +101,33 @@ func (db *DB) GetLibraryByID(id int) (*models.Library, error) {
...
+		Joins(`JOIN open_content_providers p on p.id = open_content_provider_id
+		AND p.currently_enabled = 'true'`)
+	if viewScope != "all" {
+		tx = tx.Where("visibility_status = true")
+	}
+	if err := tx.Order("libraries.title").Find(&options).Error; err != nil {
+		return options, newGetRecordsDBError(err, "libraries")
+	}
+	return options, nil
+}
```
calisio, line 129: rather than create a new function for this, wouldnt we be able to just call `GetAllLibraries`? Its a messy function that probably (definitely) needs to be cleaned up, but should give you all the libraries you are searching for now.

calisio: This would allow us to remove this function, and also the api `"/api/libraries/options"`, and just directly call `api/libraries` for all libraries

backend/src/handlers/libraries_handler.go
```diff
@@ -55,6 +64,81 @@ func (srv *Server) handleGetLibrary(w http.ResponseWriter, r *http.Request, log
...
+	selectedLibraries := r.URL.Query().Get("libraries")
+	splitedIDs := strings.Split(selectedLibraries, ",")
+	libraryIDs := make([]int, 0, len(splitedIDs))
+	for _, id := range splitedIDs {
+		if libID, err := strconv.Atoi(id); err == nil {
+			libraryIDs = append(libraryIDs, libID)
+		}
+	}
```
calisio, line 77: instead of splitting this this way, you can pass in the library ids by mapping them into the query as separate calls (so, for example you would do `api/..../...libraryId=1&libraryId=2&libraryId=3` etc) and then back here you would do `r.URL.Query()["libraries"]` and then it will return it as an array, and you will have to map through that and convert to an array of integers. see example on `user_handler.go` inside of `handleIndexUsers`

backend/src/database/libraries.go
carddev81: I remember that initially I was going to use the GetAllLibraries call, but because of the pagination logic (page, perPage) and the scalar sub select for is_favorites I shied away from it. I just want to confirm that I should move forward with making the GetAllLibraries work?

backend/src/handlers/libraries_handler.go
carddev81: I like this functionality! :)

backend/src/handlers/libraries_handler.go
carddev81: Changed this to use search instead of pattern

backend/src/database/libraries.go
calisio: Lets get a second opinion.. @PThorpe92 what do you think?

PThorpe92: I think we can add a `&all=true` param that can bypass pagination logic.

Also we need to start documenting above the handlers which parameters they accept, and any behavior associated with them

carddev81: Modified to use GetAllLibraries and added initial documentation for function and params.

backend/seeder/main.go
```diff
@@ -91,7 +91,7 @@ func seedTestData(db *gorm.DB) {
 		}
 	}
 	kiwix := models.OpenContentProvider{
-		Url:              "https://library.kiwix.org",
+		Url:              "https://kiwix.staging.unlockedlabs.xyz",
```
PThorpe92, line 94: I think I am going to launch a dedicated instance we can use for a kiwix server instead of running one on each node. That doesn't have to hold up this PR necessarily but just making a note

backend/src/database/libraries.go
```diff
@@ -13,7 +13,16 @@ type LibraryResponse struct {
...
+// Retrieves either a paginated list of libraries or all libraries based upon the given parameters.
+// page - the page number for pagination
+// perPage - the number of libraries to display on page
+// userId - the userId for which libraries to display for
+// facilityId - the facility id of where the libraries were favorited
+// visibility - can either be featured, visible, hidden, or all
+// orderBy - the order in which the results are returned
+// all - true or false on whether or not to return all libraries without pagination
+// isAdmin - true or false on whether the user is an administrator used to determine how to retrieve featured libraries
+func (db *DB) GetAllLibraries(page, perPage, days int, userId, facilityId uint, visibility, orderBy, search, all string, isAdmin bool) (int64, []LibraryResponse, error) {
```
PThorpe92, line 25: can probably pass `all` as a bool, and do
```go
showAll := r.URL.Query().Get("all") == "true"
```
in the handler

backend/src/database/libraries.go
```diff
@@ -81,14 +90,15 @@ func (db *DB) GetAllLibraries(page, perPage, days int, userId, facilityId uint,
...
 		tx = tx.Group("libraries.id").Order("favorite_count DESC")
-
 	default:
 		tx = tx.Order(orderBy)
 	}
-	if err := tx.Limit(perPage).Offset(calcOffset(page, perPage)).Find(&libraries).Error; err != nil {
+	if !all {
+		tx = tx.Limit(perPage).Offset(calcOffset(page, perPage))
+	}
+	if err := tx.Debug().Find(&libraries).Error; err != nil {
```
calisio, line 99: Is there a reason this has `tx.Debug()` ? Unfamiliar with that

frontend/src/Components/LibraryLayout.tsx
```diff
@@ -25,6 +26,35 @@ export default function LibaryLayout({
     if (!user) {
         return null;
     }
+    const navigate = useNavigate();
+    const [searchModalLibrary, setSearchModalLibrary] = useState<Library | null>(null);
```
calisio, line 30: i am wondering if rather than have this code duplicated on this page, `LibraryLayout.tsx` and `FeaturedContent.tsx`, would we be able to just move this code and the modal into `LibraryCard.tsx`? I am attempting to think about a way where we can reduce this code duplication. Because the search modal needs to be present anywhere a library card is used, then I feel like we should include it with that component? What are your thoughts on this?
The code that is duplicate incudes this usestate, the modal ref, and the functionality (opening and closing the modal). That way you wouldnt need to pass in `onSearchClick` and you could access it directly in the component? 
I am not 100% on this solution, so I am open to pushbacks. Just trying to figure out how to reduce duplicate code & make them more of a package

calisio: especially because these are very specific to libraries and the way that that search is done. seems like it ties in pretty closely with the library card itself. This modal would still have to be present in the LibraryViewer page though, since that does not include the library card.

frontend/src/Components/inputs/LibrarySearchBar.tsx
calisio, line 1: can you add these to `index.ts` and then everywhere you import them they can all be done on one line. ie on `LibrarySearchResultsModal.tsx` you can do 
`import { CloseX, MultiSelectDropdown, LibrarySearchBar} from '@/Components/inputs`

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                    'executeHandleSearch',
+                    executeSearchListener as EventListener
+                );
+            };
+        }
+    }, [ref, handleSearch]);
+    const handleCloseModal = () => {
+        setDefaultOption();
+        onModalClose();
+        setTimeout(() => setSearchResults(getKiwixChannelInstance()), 200);
```
calisio, line 191: just out of curiosity (it seems intentional), but why do you have to set a timeout for this? can you just call set search results?

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+            link: '',
+            book: '',
+            thumbnail_url: '',
+            description: '',
+            total_results: '0',
+            start_index: '0',
+            items_per_page: '10',
+            items: []
+        };
+    };
```
calisio, line 67: i dont think this has to be a function- can it just be a variable?
```
    const BlankChannel = {
        title: 'Search',
        link: '',
        book: '',
        thumbnail_url: '',
        description: '',
        total_results: '0',
        start_index: '0',
        items_per_page: '10',
        items: []
    };
```

backend/src/handlers/libraries_handler.go
calisio, line 1: question to consider- right now, we are only able to search through the libraries that are visible due to incorporating with this api. this should be the case for students, but should admin be able to search through all libraries, even those that are hidden? I would think yes (it might help them be able to search to find the library they want to show)
Screenshot 2025-01-27 at 4 11 43 PM
Screenshot 2025-01-27 at 4 11 58 PM

Searching inside of ubuntu, i can do this if i click directly on the library or on the search on that library, but it does not appear in the dropdown.
Screenshot 2025-01-27 at 4 12 22 PM

if you pass in all, you'll have to bypass `showHidden` also (might happen in database file, not handler file)

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+    }, [ref, handleSearch]);
+    const handleCloseModal = () => {
+        setDefaultOption();
+        onModalClose();
+        setTimeout(() => setSearchResults(getKiwixChannelInstance()), 200);
+    };
+    return (
+        <dialog
+            ref={ref}
+            className="modal rounded-lg bg-opacity-50 bg-black flex items-center justify-center"
```
calisio, line 196: remove everything except for "modal"- should take care of the class for you

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+            style={{ visibility: 'hidden' }}
+            onClose={handleCloseModal}
+        >
+            <div className="w-[1200px] h-[700px] bg-base-100 rounded-lg shadow-lg flex flex-col">
+                <div tabIndex={-1}></div>
```
calisio, line 201: can you move this tab index up to the above element or does it have to be a separate div?

calisio: or i guess what is the purpose of this tab index?

calisio: Also, try not to use `base-100`. it should be `bg-background`

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                <div className="sticky top-0 bg-base-100 z-50 p-4 border-b border-grey-2 rounded-t-lg">
```
calisio, line 202: same, should be `bg-background`

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                    <div className="flex items-center gap-4">
+                        <div className="flex flex-col">
+                            <h2 className="text-2xl font-bold header-text">
+                                {searchResults.title}
+                            </h2>
+                            <p className="text-sm text-grey-3">
+                                {currentPageDisplay}
+                            </p>
```
calisio, line 210: just apply className="body" here

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+            onClose={handleCloseModal}
+        >
+            <div className="w-[1200px] h-[700px] bg-base-100 rounded-lg shadow-lg flex flex-col">
+                <div tabIndex={-1}></div>
+                <div className="sticky top-0 bg-base-100 z-50 p-4 border-b border-grey-2 rounded-t-lg">
+                    <div className="flex items-center gap-4">
+                        <div className="flex flex-col">
+                            <h2 className="text-2xl font-bold header-text">
+                                {searchResults.title}
+                            </h2>
```
calisio, line 207: You only need `className="text-2xl"` here

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                                    </div>
+                                </div>
+                                <div className="mt-2">
+                                    <p className="text-sm text-grey-5 mb-2">
+                                            {item.description ? boldKeywords(item.description) : null}
+                                    </p>
+                                </div>
+                            </div>
+                        ))}
+                    </div>
```
calisio, line 277: can you extract this into its own component? like `SearchResultRow.tsx` it would clean this page up a bit. i would just pass in the whole {item} into the component so you can refer to everything the same

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                </div>
+                <div
+                    ref={scrollContainerRef}
+                    className="flex-1 overflow-y-auto p-4 scrollbar rounded-md"
+                >
+                    <div className="flex flex-col gap-4">
+                        {searchResults.items?.map((item, index) => (
+                            <div
+                                key={index}
+                                className="card bg-base-100 shadow-lg p-4 cursor-pointer bg-inner-background hover:shadow-xl transition-shadow rounded-lg"
```
calisio, line 243: just use "card card-row-padding cursor-pointer" here for consistency with other pages

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                            onBlurSearch={handleOnBlurSearch}
+                        />
+                        <CloseX close={onModalClose} />
+                    </div>
+                </div>
+                <div
+                    ref={scrollContainerRef}
+                    className="flex-1 overflow-y-auto p-4 scrollbar rounded-md"
+                >
+                    <div className="flex flex-col gap-4">
```
calisio, line 239: this div can be deleted, just move this to the div above it, ie
`className="flex flex-col gap-4 overflow-y-auto p-4 scrollbar"` (we dont need a style div inside of a style div)

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                                            />
+                                        </figure>
+                                    </div>
+                                    <div>
+                                        <h3 className="text-lg font-bold">
+                                            {item.page_title}
+                                        </h3>
+                                        <p className="text-sm text-grey-4">
+                                            {item.title}
+                                        </p>
```
calisio, line 267: change to just "body"

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                                    </div>
+                                </div>
+                                <div className="mt-2">
+                                    <p className="text-sm text-grey-5 mb-2">
+                                            {item.description ? boldKeywords(item.description) : null}
+                                    </p>
```
calisio, line 273: change to "body-small"

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                    </div>
+                </div>
+                {isLoading ? (
+                    <div className="flex h-screen gap-4 justify-center content-center">
+                        <span className="my-auto loading loading-spinner loading-lg"></span>
+                        <p className="my-auto text-lg">Loading...</p>
+                    </div>
+                ) : searchError ? (
+                    <div className="flex h-screen gap-4 justify-center content-center">
+                        <p className="my-auto text-lg text-red-500">
```
calisio, line 286: do `text-error`

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -0,0 +1,312 @@
...
+                        <p className="my-auto text-lg">Execute a search</p>
+                    </div>
+                ) : searchResults.items && searchResults.items.length === 0 ? (
+                    <div className="flex h-screen gap-4 justify-center content-center">
+                        <p className="my-auto text-lg">No Results Found</p>
+                    </div>
+                ) : (
+                    ' '
+                )}
+                <div className="border-t border-grey-2 px-6 py-4 flex justify-center rounded-b-lg">
```
calisio, line 301: i think you can remove this entire div?

carddev81:
this was something I did previously when the modal was tightly coupled with
the LibraryViewer page, it can be removed.  I will commit change.

carddev81:
made this change, instead of function created a variable.

carddev81:
modified routeLoader.ts to handle this issue, basically checking whether
the user is a Student and if so then set visibility=visible otherwise
visibility=all.

carddev81:
wow, i was worried about this one since I was having such a tough time
getting the modal to look the way it did...worked perfectly! 🙌

carddev81:
This was another weird one that was needed when coupled with the
LibraryViewer, I was able to remove it. The issue was that the Close icon
would be focused when the Esc key was pressed to close the modal.

carddev81:
made change to all base-100's on this page

carddev81:
made change

carddev81:
made change

carddev81:
moved logic within it's own card, I named it LibrarySearchResultCard, is
this good? or should I name it SearchResultCard?

carddev81:
made change

carddev81:
made change

carddev81:
made change

carddev81:
made this change. Note that the text is much smaller, just making sure this
is what is wanted:

[image: image.png]

carddev81:
i replaced text-red-500 to text-error

carddev81:
the div <div className="border-t border-grey-2 px-6 py-4 flex
justify-center rounded-b-lg"> is for the bottom border with padding the
pagination control in the center, do you still want me to remove it?

[image: image.png]

[image: image.png]

calisio:
Next time, it might be easier if you go into github to reply to the comments- not sure why its being so weird with the replies. I will respond to all in here:
1. when you changed visibility to all if you are not a student, I want to make sure that an admin on student view only sees visible libraries with this update.
2. I also am unable to see the pictures you have attached on the comments. I believe that the smaller text is what I wanted. The pagination control comment i since deleted, because i realized i was looking at the wrong pagination, so ignore that!

carddev81:
all changes have been made.

calisio (approved):
Looks good to me! Thanks for making those changes

carddev81:
thanks for lining me out.

PR #689, opened 2025-01-24
fix: modify css and add conditional for left-aligning page titles with content on page per #663

### LINKED ISSUE #663: UI: Left-align page titles with content on the page
After reviewing PR #657 I noticed that there were discrepancies in the alignment of the page title with the page content. This primarily has to do with the x-padding on the container. 

There are times when:
* the page title aligns with the containers on the page
* the page title is further left than the containers on the page
* the page title is further right than the containers on the page

In all cases, **all of the page titles and the left-most edge of the containers on the page should line up**. When the navigation bar is hidden (using the `<<` icon) then the hamburger icon (three bars) should be visible and what is lined up with the content on the page.

### Example where the page content is probably pushed in too far

image

### Example where the padding on the title doesn't line up with the page content.

image

### Example where the hamburger menu isn't aligned with the page content.

image

carddev81:
## Description of the change

Modified CSS padding styles on several pages for aligning Page titles with the content found on the pages. 

- **Related issues**: Closes #663.

## Screenshot(s)
image
image

### DIFF
--- a/frontend/src/Components/OperationalInsightsCharts.tsx
+++ b/frontend/src/Components/OperationalInsightsCharts.tsx
@@ -33,7 +33,7 @@ const OperationalInsights = () => {
         (metrics?.total_residents ?? 0) + (metrics?.total_admins ?? 0);
 
     return (
-        <div className="p-6 max-w-full overflow-x-hidden">
+        <div className="overflow-x-hidden">
             {error && <div>Error loading data</div>}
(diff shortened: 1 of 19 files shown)


PThorpe92:
@jtucholski I can't even tell the difference in the photos from the issue, so I'm pretty useless in reviewing this :sweat_smile:

jtucholski (approved):

PR #695, opened 2025-01-28
fix: add NATS cache to AdminLayer2

### LINKED ISSUE #685: Layer 2 'learning insights' admin dashboard needs cache
Image

This dash has a `Refresh Data` button, similar to the `Operational Insights` dash. The point of this is to tell the back-end to perform a fresh query, because the data in these insights doesn't change much and perform multiple, sometimes relatively expensive queries.

The Login metrics handler does this by creating a key out of the query parameters, and stores the value of the query in NATS kv bucket.

```go
	key := fmt.Sprintf("%s-%d", facility, days)
	cached, err := srv.buckets[LoginMetrics].Get(key)
```

A bucket would need to be created for this dash and the logic would need to be added to the handler to cache the result

CK-7vn:
## Description of the change
This pull request adds a cache for the admin layer 2 dashboard/learning insights.  

- **Related issues**: Closes #685 

## Additional context
Had to add a little bit of redundant code to test if we are testing because NATS isn't running during testing. 

## Here is the timestamps!
image

image

### DIFF
--- a/backend/src/handlers/dashboard.go
+++ b/backend/src/handlers/dashboard.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/nats-io/nats.go"
 )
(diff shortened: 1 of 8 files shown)


PThorpe92:
Josh had a good point, maybe we can just add it in this PR? if not, I can make a new issue but he said maybe we expose a timestamp on the frontend to show how stale the data is, in case they might want to refresh it? or at least it makes it obvious why they would want to refresh it.

CK-7vn:
@PThorpe92 yes, we've discussed this, I will get this done and push. Good point, good point.

backend/src/handlers/dashboard.go
```diff
@@ -40,39 +41,87 @@ func (srv *Server) handleAdminLayer2(w http.ResponseWriter, r *http.Request, log
 		ref := uint(facilityIdInt)
 		facilityId = &ref
 	}
+	if srv.isTesting(r) {
```
PThorpe92, line 44: can we pull this out to a function and call it for both circumstances when the calls are needed

carddev81 (changes requested):
Noticed that the CSS changed on the learning insights page.  This will need to change back to how it was, see screeshots below.  Staging is correct. 
image

image

backend/src/handlers/server.go
```diff
@@ -236,6 +237,8 @@ func (srv *Server) setupNatsKvBuckets() error {
 				cfg.TTL = time.Hour * 24
 			case OAuthState:
 				cfg.TTL = time.Minute * 10
+			case AdminLayer2:
```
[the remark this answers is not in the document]
carddev81: @CK-7vn sorry for creating confusion here, but you can remove the following lines since you added the default case

```
			case LoginMetrics:
				cfg.TTL = time.Hour * 24
			case AdminLayer2:
				cfg.TTL = time.Hour * 24
```

frontend/src/Pages/AdminLayer2.tsx
```diff
@@ -10,81 +10,108 @@ import {
...
+        layer2_metrics &&
+        new Date(layer2_metrics?.last_cache).toLocaleString('en-US', {});
 
     if (error || isLoading || !user) return <div></div>;
     if (!isAdministrator(user)) {
         return <UnauthorizedNotFound which="unauthorized" />;
     }
     return (
-        <div className="w-full flex flex-col gap-2 px-5 pb-4">
+        <div className="w-full flex flex-col gap-2 pb-4 px-">
```
carddev81, line 43: missing the 5 after px- here---   px-5

carddev81 (changes requested):

PR #698, opened 2025-01-29
fix: add existing helpful link toast

### LINKED ISSUE #643: UX Enhancement: Improve Error Message for Duplicate Helpful Links
**Description:**  
Admins encounter an error when trying to create a new Helpful Link with the same URL as an existing one.  

Once PR #635 is merged, users will see an error toast message that states:  
**Error adding helpful link.**  

This occurs because `helpful_links.go` checks for an existing Helpful Link with the same URL and returns a "Link already exists" error to the client.

**Proposed Solution:**  
Since there may be many Helpful Links, we should provide a clearer error message to users. Instead of the generic message, display **"Link already exists"**. This will help users understand what went wrong and avoid repeated attempts to add the same link.

calisio:
## Description of the change

Adds toaster message for "Link already exists". 

- **Related issues**: closes #643 

## Screenshot(s)

Screenshot 2025-01-29 at 12 34 55 PM

### DIFF
--- a/backend/src/database/helpful_links.go
+++ b/backend/src/database/helpful_links.go
@@ -46,7 +46,7 @@ func (db *DB) GetHelpfulLinks(page, perPage int, search, orderBy string, onlyVis
 
 func (db *DB) AddHelpfulLink(link *models.HelpfulLink) error {
 	if db.Where("url = ?", link.Url).First(&models.HelpfulLink{}).RowsAffected > 0 {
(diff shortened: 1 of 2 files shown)


jtucholski (approved):

carddev81 (approved):

carddev81:
If you want to remove the String() call here.  Looks good and ran tests to make sure everything worked as expected.  Ready for merge when you are.

PR #699, opened 2025-01-29
fix: wrong col count for helpful links

jtucholski:
## Description of the change

When introducing the `ExpandableCardGrid` I made a mistake and didn't apply the four-column layout for the helpful links. This fix adds a cols prop to the grid and sets the Featured content at 3 cols and Helpful links at 4 cols.

## Screenshot(s)

image


CK-7vn (commented):
LGTM.

CK-7vn (approved):

PR #727, opened 2025-02-05
feat: refactor modals

### LINKED ISSUE #652: Refactor Modal Component and Form Handling
**Description:**  
During the review of PR #645, we identified opportunities to refactor the `Modal` component and the Add, Edit, and Delete forms to address some recurring problems and improve maintainability.

**Existing Problems:**  
1. **CloseX Component**:  
   - Each form currently implements its own `CloseX` component and a `closeAndReset` function to reset the form.  
2. **Form Reset on Modal Close (ESC):**  
   - Closing a modal using the `ESC` key happens at the `Modal/HTMLDialogElement` level. However, form values inside the modal were not reset when the modal was closed.  
   - To address this, a `useEffect` hook was added to each form to wipe the values when the dependent object changes. While this works as a temporary fix, it’s not a long-term solution.

**Proposed Solution:**  
- Move the responsibility for resetting the form to the `Modal` component.  
- Refactor `FormProps` so they can be passed into the `Modal` along with the form itself, enabling the `Modal` to invoke `reset()` on any form.  

This refactor would simplify form handling, reduce redundant code, and improve the maintainability of the `Modal` and related components.

calisio:
## Description of the change

Refactors modal into a new modal component. This will standardize the way that we implement form modals by passing an array of inputs, which will then create the form automatically. This will eliminate the need to create a new file included in the `/forms` folder.

NOTE: will be moving the interfaces/enums to `common.ts`, but just for simplicity now i am keeping them in the new modal. I also will be renaming at the very end once all is transferred to new modal.

**Related issues**: closes #652 

## Additional context

Still a work in progress. Leaving as a draft for people to look at it first before implementing across most pages that include modals.

## Update 2/17/2025
Mostly all modals have been updated. Some forms still exist in the `/forms` folder, but these are typically forms that are free standing (not inside modals). Exceptions are:
- `MapUserForm.tsx` because of the complexity of this form which does not follow the same pattern.
- Anything related to programs is not updated. (ie add program modal, create program form)

## Update 2/18/2025
- added a new hook called `useCheckResponse` (open to name change for specificity) to check responses and call correct functions

---
  - https://app.asana.com/0/0/1209460076710897


frontend/src/Components/Modaltest.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const {
+        register,
+        reset,
+        handleSubmit,
+        formState: { errors }
+    } = useForm<T>({ defaultValues: defaultValues });
+
+    useEffect(() => {
+        reset(defaultValues);
+    }, [defaultValues, reset]);
```
jtucholski, line 51: Does this handle the case where the user presses Esc and the form resets back to it's default values?

frontend/src/Pages/FacilityManagement.tsx
```diff
@@ -40,73 +46,104 @@ export default function FacilityManagement() {
...
+            deleteFacilityModal.current?.showModal();
+        }
+    }, [targetFacility]);
+
+    const openDeleteFacility = (facility: Facility) => {
+        setTargetFacility({
+            action: CRUDActions.Delete,
+            facility: facility
+        });
+    };
```
jtucholski, line 69: It feels like these should have the word handle in the name. My initial reaction was that it's doing the same as what is in the `useEffect`

frontend/src/Pages/FacilityManagement.tsx
```diff
@@ -40,73 +46,104 @@ export default function FacilityManagement() {
...
+            void mutate();
+        } else {
+            toaster('Error deleting facility', ToastState.error);
+        }
         deleteFacilityModal.current?.close();
-        resetModal();
-        void mutate();
+        setTargetFacility(null);
         return;
     };
```
jtucholski, line 145: What do you think about something like putting the form inputs and the handlers into their own file so that we have a reusable form module for managing facilities? 

It cleans up the FacilityManagement page a bit more and adds an element of reusability (not that we have a use case for it yet).

I could be overthinking it and I am open to you pushing back on this.

jtucholski (commented):
Looks good. I like the direction this is going and added a few comments for consideration.

frontend/src/Components/Modaltest.tsx
calisio: It does

frontend/src/Pages/FacilityManagement.tsx
```diff
@@ -40,73 +37,67 @@ export default function FacilityManagement() {
...
-        deleteFacilityModal.current?.showModal();
-        setTargetFacility(facility);
-    };
-    const handleDeleteFacility = async () => {
-        if (targetFacility?.id == 1) {
+
+    const deleteFacility = async () => {
+        if (targetFacility?.target.id == 1) {
             toaster('Cannot delete default facility', ToastState.error);
+            deleteFacilityModal.current?.close();
```
jtucholski, line 80: @calisio Should this be `closeModal(defaultFacilityModal)`?

calisio: yes, thank you for the catch!

frontend/src/Pages/HelpfulLinksManagement.tsx
```diff
@@ -48,20 +52,20 @@ export default function HelpfulLinksManagement() {
...
+        const response = await API.delete(`helpful-links/${currentLink?.id}`);
+        if (!response.success) {
             toaster('Error deleting helpful link', ToastState.error);
         }
-        deleteLinkModal.current?.close();
+        toaster('Helpful link deleted successfully', ToastState.success);
+        await mutate();
+        closeModal(deleteLinkModal);
+        setCurrentLink(null);
+    }
```
jtucholski, line 64: Do you want to update this one to use the new checkResponse hook that you created?

frontend/src/Components/modals/EditUserModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+        mutate: mutate,
+        refModal: editUserModal
+    });
+    const editUser: SubmitHandler<FieldValues> = async (data) => {
+        const resp = await API.patch(`users/${target?.id}`, data);
+        checkResponse(
+            resp.success,
+            'Unable to modify user',
+            'User updated successfully'
+        );
```
jtucholski, line 22: Checking to make sure we can't add similar message logic like we do in add user because the backend doesn't give a detailed enough error message if the edit user fails. This is what I get when I try to update a user to take an existing user name.

```
{"message":"username","data":null}
```

frontend/src/Pages/StudentManagement.tsx
```diff
@@ -302,72 +314,51 @@ export default function StudentManagement() {
...
-                                : undefined
-                        }
-                        onClose={handleShowPasswordClose}
-                    />
-                }
-            />
+                type={TextModalType.Information}
+                title={'New Password'}
+                text={`Copy this password now. If you lose it, you'll need to
+                        regenerate it to get a new one.`}
```
jtucholski, line 351: Can you update this text to

"Copy your password now. If you lose it, you'll need to generate a new one."

frontend/src/Pages/AdminManagement.tsx
```diff
@@ -318,72 +281,43 @@ export default function AdminManagement() {
...
-                                : undefined
-                        }
-                        onClose={handleShowPasswordClose}
-                    />
-                }
-            />
+                type={TextModalType.Information}
+                title={'New Password'}
+                text={`Copy this password now. If you lose it, you'll need to
+                        regenerate it to get a new one.`}
```
jtucholski, line 310: Can you update this text to

"Copy your password now. If you lose it, you'll need to generate a new one."

frontend/src/Pages/VideoManagement.tsx
jtucholski, line 1: I noticed `videoErrorModal.current?.showModal();`. Can we use `showModal(videoErrorModal)` here?

frontend/src/Components/modals/index.ts
```diff
@@ -0,0 +1,295 @@
...
+    return true;
+};
+
+export const videoInputs: Input[] = [
+    {
+        type: FormInputTypes.TextArea,
+        label: `This will begin to attempt to download videos. You can enter
+        links from youtube or any other popular video hosting site.
+        If there is an error, downloads will be attempted up to 5
+        times, after which time you will need to re-add the video.`,
```
jtucholski, line 288: Can you shorten this message to

"Enter video URLs from YouTube and similar sites to download. If a download fails, it will be retried up to 5 times before the video needs to be added again."

jtucholski (commented):
Looks great! I added a few comments to consider. Otherwise it really helps clean up the codebase and brings some consistency to how we deal with modals!

jtucholski (approved):

calisio:
Waiting to resolve merge conflicts because I am going to squash all commits into one so rebasing is easier. I'll have to force push but don't want to lose the comments that already exist on the PR.

PThorpe92 (approved):

calisio:
Seems like once you push a new commit it dismisses the old reviews automatically...promise I didn't do that :)
@PThorpe92 and @jtucholski do you mind looking over/approving again if everything looks good?

PR #728, opened 2025-02-06
add feature to enable email to admins to whitelist helpful links

PThorpe92:
image

with the following tooltip:
image

This Pr adds the ability for an admin to request that an email be sent to the network admin

It also adds the `email` form back to the New User form, but only when its an admin user being created. this is so we are able to use their email for reference in the request sent to the domain admin for whitelisting.


frontend/src/Pages/HelpfulLinksManagement.tsx
```diff
@@ -53,6 +53,19 @@ export default function HelpfulLinksManagement() {
         editLinkModal.current?.close();
         void mutate();
     }
+
+    async function reqWhitelist(id: number) {
+        const response = await API.post(`helpful-links/${id}/whitelist`, {});
+        if (response.success) {
+            toaster(
+                'Request to Whitelist link request sent successfully',
```
carddev81, line 61: change W to lowercase for consistency with toaster messages

frontend/src/Pages/HelpfulLinksManagement.tsx
```diff
@@ -53,6 +53,19 @@ export default function HelpfulLinksManagement() {
         editLinkModal.current?.close();
         void mutate();
     }
+
+    async function reqWhitelist(id: number) {
```
carddev81, line 57: should this entire function be moved into HelpfulLinkCard.tsx? unless there is a reason for it being here, you could remove some lines of code.

frontend/src/Components/forms/AddUserForm.tsx
```diff
@@ -153,6 +156,20 @@ export default function AddUserForm({ onSuccess, userRole }: AddUserFormProps) {
                     }}
                 />
 
+                {isAdministrator(user) && userRole === UserRole.Admin && (
+                    <TextInput
+                        label={'Email'}
+                        interfaceRef={'email'}
+                        length={60}
```
carddev81, line 163: making sure that max length should be 60 here due to database column able to hold 255.

backend/src/handlers/server.go
```diff
@@ -167,6 +172,7 @@ func (srv *Server) initAwsConfig(ctx context.Context) {
 		if err != nil {
 			log.Fatal(err)
 		}
+		srv.sesClient = sesv2.NewFromConfig(cfg)
```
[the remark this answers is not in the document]
PThorpe92: NewFromConfig will automatically check for the AWS_DEFAULT_REGION env var, in fact because the nodes are on AWS, you can set the instance metadata so that you never even have to authenticate, it will just work, with the proper IAM role you assign to the instance :)

backend/src/handlers/server.go
```diff
@@ -485,3 +491,79 @@ func (srv *Server) calculateLast(total int64, perPage int) int {
...
+		bodyHTML: bodyHTML,
+	}
+}
+
+func newWhitelistRequest(user *models.User, link *models.HelpfulLink) *emailReq {
+	reqTime := time.Now().Format("2006-01-02 15:04:05")
+	emailStr := `A new Helpful Link has been added to the UnlockEd system by an Administrator.
+	In order for it to successfully become available to users, it must be whitelisted globally after your pending review of the site. 
+	It may already be available, but the Administrator has chosen to submit it for whitelisting`
+	emailTo := os.Getenv("DOMAIN_ADMIN_EMAIL")
```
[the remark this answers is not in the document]
PThorpe92: Yeah even better maybe we make a deployment page in the wiki that lists all the necessary env vars for a production deployment.

frontend/src/Pages/HelpfulLinksManagement.tsx
PThorpe92: typically functions like this in card elements tend to be passed in because different views/pages the card may be used on, may or may not need them

backend/src/handlers/server.go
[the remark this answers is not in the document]
PThorpe92: https://github.com/UnlockedLabs/UnlockEdv2/wiki/Deployment-environment

backend/src/handlers/helpful_links_handler.go
```diff
@@ -81,6 +82,30 @@ func (srv *Server) handleAddHelpfulLink(w http.ResponseWriter, r *http.Request,
...
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	log.add("link", link.Url)
+	cliams := r.Context().Value(ClaimsKey).(*Claims)
+	user, err := srv.Db.GetUserByID(cliams.UserID)
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	emailReq := newAllowlistRequest(user, link)
```
jtucholski, line 100: Is this sending the email to the user (the Admin requesting access) or is it sending it to a different user?

jtucholski: nvm.

backend/src/handlers/utils.go
```diff
@@ -92,3 +120,80 @@ func (srv *Server) getQueryContext(r *http.Request) models.QueryContext {
...
+		from:     from,
+		subject:  subject,
+		bodyText: bodyText,
+		bodyHTML: bodyHTML,
+	}
+}
+
+func newAllowlistRequest(user *models.User, link *models.HelpfulLink) *emailReq {
+	reqTime := time.Now().Format("2006-01-02 15:04:05")
+	emailStr := `A new Helpful Link has been added to the UnlockEd system by an Administrator.
```
jtucholski, line 144: @PThorpe92 Should the message in the emailStr match the html version?

I modified the message for grammar. Can you have it say?

> An Administrator has added a new Helpful Link to the UnlockEd system and requests your review to confirm it is allow-listed for all users.

jtucholski (commented):
Left one comment on the message body and pushed one change for the helpful link layout.

PThorpe92:
@jtucholski In Maine, the same email address is used for all facilities, but this is not likely to be the case everywhere. Perhaps this needs some more thought. Maybe a UI to add the email for each facility?

PR #757, opened 2025-02-14
fix: use google cloud api key to fetch youtube metadata

PThorpe92:
the error with fetching the thumbnails during the pilot was due to spawning child processes when we just need to fetch the thumbnail. 

we can get the info with the google api instead with one request, as we were before.

### DIFF
--- a/provider-middleware/video_dl.go
+++ b/provider-middleware/video_dl.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"mime/multipart"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -16,6 +17,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/sirupsen/logrus"
 	"github.com/wader/goutubedl"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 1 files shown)


provider-middleware/video_dl.go
```diff
@@ -378,6 +411,54 @@ func (yt *VideoService) addVideos(ctx context.Context) error {
...
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		logger().Errorf("Error decoding video info: %v", err)
+		return nil, err
+	}
+	if len(data.Items) == 0 {
+		logger().Errorf("Video not found in response from youtube API")
+		return nil, fmt.Errorf("Video not found")
+	}
+	logger().Infof("Adding videos: %v", data)
+	thumbnailUrl, err := vs.downloadAndHostThumbnail(data.Items[0].ID, data.Items[0].Snippet.Thumbnails.High.Url)
```
carddev81, line 439: Is it possible that the value here `data.Items[0].Snippet.Thumbnails.High.Url` doesn't exist as in `High == nil` ? Maybe create a conditional block to capture the Url of the best resolution available?

provider-middleware/video_dl.go
```diff
@@ -378,6 +411,54 @@ func (yt *VideoService) addVideos(ctx context.Context) error {
...
+	}
+	defer resp.Body.Close()
+	var data YoutubeDataResponse
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		logger().Errorf("Error decoding video info: %v", err)
+		return nil, err
+	}
+	if len(data.Items) == 0 {
+		logger().Errorf("Video not found in response from youtube API")
+		return nil, fmt.Errorf("Video not found")
```
carddev81, line 436: should `errors.New` be used here since there is no error to pass into `fmt.Errorf`?  Also, the logger method called above it can be changed to just `Error`.

carddev81 (dismissed):
Looks good, tested new method and all was good.  Had a couple of comments.

jtucholski (approved):

PR #772, opened 2025-02-21
feat: add bookmarking

### LINKED ISSUE #760: Support Bookmarking of Individual Library Pages
## Description

Feedback during the pilot indicated that users would find it valuable to include bookmarking (or saving) specific pages inside of a library so that they don't need to navigate through the library to find those pages again.

This task brings that feature back into the application and focuses on making it easy for the user to bookmark one or more pages inside of a library and provides them with a way to see those bookmarks.

## User Stories

As a user who is browsing through a large open content library
I want to bookmark individual pages that I come across
so that I can easily find them again.

As a user who has bookmarked individual library pages
I want to see a list of my bookmarks
so that I can quickly locate what I am looking for.

## Acceptance criteria

* When browsing any library, the user has a way to bookmark an individual page.
* The user can see whether or not the page they are on has been bookmarked.
* The user can remove the bookmark from a bookmarked page if they have it open.
* The user has a way to see all of their bookmarks alongside their other favorites.

CK-7vn:
## Description of the change
This pull requests gives users the ability to "bookmark" nested content in the library viewer so that they can easily return back to said nested content down the road. 

- **Related issues**: Closes #760 

## Screenshot(s)
https://www.loom.com/share/35e0b30b8a1d4f4ebc5817f8f56778d9?sid=0e4bb0da-75f9-44ed-88d6-dcf462aa1fb6

**??**
1. The thumbnails of the libraries, are actually kiwix.jpg, in the database, so I'm not sure why the legitimate thumbnails aren't in there ATM.

---
  - https://app.asana.com/0/0/1209465983032135


backend/src/handlers/websockets.go
```diff
@@ -0,0 +1,151 @@
...
+	log.Infof("Sending message to user_id %d, message: %s", client.UserID, message)
+	client.sendChan <- message
+}
+
+func (client *WsClient) writePump() {
+	for {
+		select {
+		case message := <-client.sendChan:
+			log.Infof("Writing message to user_id %d", client.UserID)
+			err := client.Conn.Write(context.Background(), websocket.MessageText, message)
```
PThorpe92, line 84: let's use the client's context here instead of `context.Background`

frontend/src/Pages/LibraryViewer.tsx
```diff
@@ -148,6 +289,35 @@ export default function LibraryViewer() {
...
+                            />
+                            <div className="flex justify-end space-x-4">
+                                <SubmitButton />
+                            </div>
+                            <CloseX
+                                close={() => favoriteModalRef.current?.close()}
+                            />
+                        </form>
+                    }
+                />
```
calisio, line 320: now that the modal refactor is merged, this will need to be updated using the new form modal

frontend/src/Pages/LibraryViewer.tsx
```diff
@@ -111,11 +133,130 @@ export default function LibraryViewer() {
...
+            } else {
+                toaster(response.message, ToastState.error);
+            }
+        } catch (error) {
+            toaster('Error updating favorite status', ToastState.error);
+            console.error(error);
+        } finally {
+            favoriteModalRef.current?.close();
+            reset();
+        }
```
calisio, line 237: i would also change this format. implement using `useCheckResponse`

carddev81 (changes requested):
Noticed that the favorites in the User's Favorites accordion are not displaying the custom name that the user enters (see image).
image

Favoriting and Featuring functionality is not working correctly, star and flag icons do not toggle on/off:
[FavoritingAndFeaturing.webm](https://github.com/user-attachments/assets/0c7e3842-3f34-4542-a4a3-5e51b47e0136)


Hm....that's very interesting

CK-7vn:
> Noticed that the favorites in the User's Favorites accordion are not displaying the custom name that the user enters (see image). image
> 
> Favoriting and Featuring functionality is not working correctly, star and flag icons do not toggle on/off: [FavoritingAndFeaturing.webm](https://github.com/user-attachments/assets/0c7e3842-3f34-4542-a4a3-5e51b47e0136)
> 
> Hm....that's very interesting

The favoriting/featuring issue was because the OpenContentUrlId in the OpenContentFavorite was a uint, not a uint pointer, so it was defaulting to its 0 value, I changed it to a pointer and that issue is fixed.

frontend/src/Pages/LibraryViewer.tsx
CK-7vn: Unsure of how to do this as we have no mutate throughout the component, not quite versed well enough to figure this one out.

calisio: Ah, that makes sense. The web socket is pulling the bookmark and then updating it to true/false, right? In that case, there isnt a mutate and you are probably fine with this.

carddev81 (approved):
Looks good! Tested Good.

PThorpe92 (approved):

PR #773, opened 2025-02-21
fix: remove admin ability to alter data in student view

### LINKED ISSUE #771: Student Preview Mode Shouldn't Modify Data
## Description

Currently if an Admin enters _Preview Mode_ and **favorites a library**, the library becomes **featured.** This could be confusing because the UI is as a Resident sees it but it functionally behaves as if an Admin is featuring it.

A simpler approach to handling this would prohibit Admins from modifying or changing any data when they are in _Preview Mode._

CK-7vn:
## Description of the change
This PR removes the ability of admins to favorite while in preview mode on Libraries, Helpful Links, and Videos. 

- **Related issues**: Closes #771 

## Screenshot(s)
https://www.loom.com/share/1b6986f253a04f1fbebea7477f5f103e?sid=fde981af-4d93-4ee2-aaec-866eaa29bb6f

---
  - https://app.asana.com/0/0/1209457884707537

### DIFF
--- a/frontend/src/Components/FavoriteCard.tsx
+++ b/frontend/src/Components/FavoriteCard.tsx
@@ -69,6 +69,10 @@ const FavoriteCard: React.FC<FavoriteCardProps> = ({
         }
         const response = await API.put(endpoint, payload);
         if (adminWithStudentView()) {
+            toaster(
+                "You're in preview mode. Changes cannot be made.",
+                ToastState.null
+            );
             return;
         }
         if (response.success) {
--- a/frontend/src/Components/FavoriteCard.tsx
+++ b/frontend/src/Components/FavoriteCard.tsx
@@ -1,9 +1,11 @@
-import { useNavigate } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 import { StarIcon as SolidStar } from '@heroicons/react/24/solid';
 import { KeyedMutator } from 'swr';
 import { ToastState, ServerResponseMany, OpenContentItem } from '@/common';
 import API from '@/api/api';
 import { useToast } from '@/Context/ToastCtx';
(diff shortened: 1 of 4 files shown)


PThorpe92 (approved):

calisio (changes requested):
Hm, I would probably tell the user that they are not able to favorite the library in student view with a tooltip. If not, this can be confusing as to why the favorite button is not working in that view. Probably should disable the button, and also just modify the tooltip to say "Unable to favorite while in student view" or something of the sort. Either that, or you could also return a toast that says the same information onclick for the admin. Up to you, but something should show the user _why_ this feature is not available

jtucholski:
@CK-7vn 

I haven't reviewed this PR but just seeing @calisio comment there's a couple of ideas that jump out:
- Leave the icons in there let them behave as normal but don't persist the state changes. Maybe include a message or alert somewhere that tells them "Changes made while in student mode won't be saved."
  - Pros: Gives Admin the chance to see how a user interacts with it.
  - Cons: Could be confusing to implement because we need to turn off anything that changes state in this mode. 
- Remove the icons so that the Admin isn't tempted to interact with them.
  - Pros: Removes the chance that they could accidentally favorite/feature the library.
  - Cons: Requires special messaging why the UI is slightly different in this student view.
- Add a tooltip and disable the icon.
  - Pros: Easy fix. 
  - Cons: Tooltips are suddenly conditional ("Favorite Library" vs. "Disabled in student view") and we're likely to miss them.

CK-7vn:
Have we come to any conclusions on what we're thinking here?

jtucholski:
I added a toast message that displays if they're in Admin mode trying to take an action

image

jtucholski (approved):

PR #778, opened 2025-02-21
Search fix

PThorpe92:
search through all open content, returning videos/title results first, then text results from libraries.

(image is backwards, videos would be first)
image

---
  - https://app.asana.com/0/0/1209519798867981


frontend/src/Pages/VideoManagement.tsx
```diff
@@ -30,23 +30,58 @@ export default function VideoManagement() {
...
     const { toaster } = useToast();
     const { data, mutate, error, isLoading } = useSWR<
         ServerResponseMany<Video>,
         AxiosError
     >(
-        `/api/videos?search=${searchQuery[0]}&page=${pageQuery}&per_page=${perPage}&order_by=${sortQuery}`
+        `/api/videos?page=${pageQuery}&per_page=${perPage}&order_by=${sortQuery}`
     );
 
+    const navToViewer = (
```
CK-7vn, line 69: Is this switch statement for navigation something that could be extracted out to the search modal, or is it more effective/easier to keep it here and the library layout?

CK-7vn (approved):
💯 Everything looks solid, just had that one question about the navToViewer function, but definitely not an issue. 🥇

frontend/src/Pages/VideoManagement.tsx
PThorpe92: hmm unsure. The component takes "On item click" function. I'm not sure if we would ever need any different behavior to be injected there? @calisio

backend/src/models/library.go
```diff
@@ -74,41 +74,56 @@ type Book struct {
...
-				DatabaseFields: DatabaseFields{
-					ID: library.ID,
-				},
+		thumbnail := ""
+		if library == nil {
+			continue
+		} else if library.ThumbnailUrl != nil {
+			thumbnail = *library.ThumbnailUrl
+		}
+		kiwixItem := SearchResultItem{
```
jtucholski, line 119: thoughts about just renaming this to something else so its not attached to Kiwix as well?

backend/src/models/library.go
```diff
@@ -74,41 +74,56 @@ type Book struct {
...
+			OpenContentItem: OpenContentItem{
+				ContentId:    library.ID,
 				Url:          fmt.Sprintf("/api/proxy/libraries/%d%s", library.ID, item.Link),
```
jtucholski, line 122: This here overwrites the URl that's generated when doing the title search in `database/libraries.go` (see other comment about writing the query for front-end routes)

frontend/src/Pages/VideoManagement.tsx
```diff
@@ -27,29 +25,66 @@ import {
...
-    const searchQuery = useDebounceValue(searchTerm, 300);
     const videoErrorModal = useRef<HTMLDialogElement>(null);
     const [polling, setPolling] = useState<boolean>(false);
     const [perPage, setPerPage] = useState(12);
     const [pageQuery, setPageQuery] = useState(1);
     const [sortQuery, setSortQuery] = useState<string>(
         FilterLibrariesVidsandHelpfulLinksAdmin['Title (A to Z)']
     );
+    const modalRef = useRef<HTMLDialogElement>(null);
+    const [searchModalOpen, setSearchModalOpen] = useState<number | null>(null);
```
jtucholski, line 44: Is there a reason to using a number for this or is it preference?

frontend/src/Pages/VideoManagement.tsx
```diff
@@ -27,29 +25,66 @@ import {
     TextModalType,
     TextOnlyModal
 } from '@/Components/modals';
+import { LibrarySearchBar } from '@/Components/inputs';
+import LibrarySearchResultsModal from '@/Components/LibrarySearchResultsModal';
```
jtucholski, line 29: @PThorpe92 or @calisio Should we be using FormModal when working with the LibrarySearchResults modal?

frontend/src/Pages/VideoManagement.tsx
```diff
@@ -109,10 +139,17 @@ export default function VideoManagement() {
...
+                        <div onClick={() => setSearchModalOpen(1)}>
+                            <LibrarySearchBar
+                                onSearchClick={openSearchModal}
+                                searchPlaceholder="Search..."
+                                searchTerm={searchTerm}
+                                changeCallback={setSearchTerm}
+                                isSearchValid={searchTerm.trim() !== ''}
+                            />
+                        </div>
+                    )}
```
jtucholski, line 152: Should the OrderBy control be included in the conditional check above?

frontend/src/Pages/LibraryViewer.tsx
```diff
@@ -39,21 +39,17 @@ export default function LibraryViewer() {
         }
     };
 
-    const handleSearchResultClick = (
+    const navToLibraryViewer = (
+        _kind: string,
```
jtucholski, line 43: This stuck out to me. Is there a naming collision with `kind`?

jtucholski (commented):
I noticed that the Resident view uses `VideoContent` not `VideoManagement` so the search modal is never triggered when the user clicks on it.

jtucholski (changes requested):
I left a few comments throughout the PR but forgot to click Request Changes.

frontend/src/Pages/VideoManagement.tsx
PThorpe92: was copied from LibraryManagement i believe, i'll change them both

frontend/src/Pages/LibraryViewer.tsx
PThorpe92: previously it wasn't going to use that variable because it was only going to search the current library

backend/src/models/library.go
PThorpe92: I left it because in this case, this function is operating on all kiwix items returned from FTS. I'm not against changing it though

backend/src/models/library.go
PThorpe92: ok nevermind i see what you are saying

frontend/src/Pages/VideoManagement.tsx
calisio: I thought about it, but I think that because that search modal is very drastically different than the form modal, I think I am okay with having it as a completely separate component. Does that answer your question?

frontend/src/Pages/VideoManagement.tsx
PThorpe92: in the `{videoData && videoData.length > 0 && (`  conditional ?

jtucholski (approved):

PR #779, opened 2025-02-25
feat: modify library visibility to be per facility according to ticket #718

### LINKED ISSUE #718: Allow Facility Admins to Manage Library Visibility in Knowledge Center
## **User Story**  
As a **Facility Admin**, I want to **control the visibility of Libraries at my facility**, so that I can manage content access for Residents at my specific location.  

## **Description**  
Currently, Libraries can only be set as **Visible or Hidden** currently). This issue enables **Facility Admins to manage libraries ensuring they can:  
- Toggle Libraries **Visible/Hidden** for their facility only.  
- **View which Libraries** are currently visible at their facility.  
- Ensure Residents **only see Libraries** that has been set to **Visible** by either a Facility Admin or a Department Admin.  

### **Backend Updates**  
- [ ] Modify the database schema to support **Facility Admin-controlled visibility settings** 
- [x] Update API endpoints to:  
  - Allow **Facility Admins to toggle visibility** only for their assigned facility.  
  - Prevent Facility Admins from modifying visibility settings for other facilities.  

### **Facility Admin UI Updates**  
- [x] Ensure the existing **Visible/Hidden toggle** on Library cards **works at the facility level**.  
- [x] Update **Facility Admin Knowledge Center view** to only show resources **available at their facility**.  

### **Resident View Updates**  
- [x] Modify Resident queries to ensure **only content marked "Visible" for their facility is displayed**.  
- [x] Ensure “Hidden” content is **invisible** to Residents, regardless of facility.  

### **For Facility Admins**  
1. Logs into UnlockEd → Navigates to the **Knowledge Center Admin Panel**.  
2. Views the list of Libraries **available at their facility**.  
3. Clicks the **Visible/Hidden toggle** for a Library.  
4. If toggling **Hidden**, the Library disappears for residents at their facility (but remains available at other facilities).  
5. If toggling **Visible**, the Library becomes accessible to residents at their facility.  
6. Saves changes → The updated visibility settings are reflected immediately.  

## **Acceptance Criteria**  
- Facility Admins can set visibility **only for their facility**.  
- The **Visible/Hidden toggle functions** at the facility level without affecting other locations.  
- Residents **only see content marked "Visible" for their facility**.  
- API correctly **stores facility-specific visibility settings**.  
- UI dynamically updates based on facility selection.

carddev81:
## Description of the change

Modified the library visibility functionality to be toggled on a per facility basis rather than globally. The changes include:

- Created a new table (facility_visibility_statuses) to manage library visibility per facility.  
- Removed the visibility_status column from the libraries table.  
- Added a new struct with gorm mapping to the new table.  
- Updated SQL queries to reflect the new visibility logic  

- **Related issues**:  Closes #718

---
  - https://app.asana.com/0/0/1209485070836358

### DIFF
--- a/backend/src/database/libraries.go
+++ b/backend/src/database/libraries.go
@@ -31,12 +31,15 @@ func (db *DB) GetAllLibraries(args *models.QueryContext, visibility string) ([]L
 	libraries := make([]LibraryResponse, 0, args.PerPage)
 	//added the below to display featuring flags for all admins per facility
 	selectIsFavoriteOrIsFeatured := `
-        libraries.*,
-        EXISTS (
-            SELECT 1
-            FROM open_content_favorites f
-            WHERE f.content_id = libraries.id
-              AND f.open_content_provider_id = libraries.open_content_provider_id
+		libraries.*,
+		CASE WHEN fvs.visibility_status IS NULL THEN false
+			ELSE fvs.visibility_status
+		END AS visibility_status,
+		EXISTS (
+			SELECT 1
+			FROM open_content_favorites f
+			WHERE f.content_id = libraries.id
+				AND f.open_content_provider_id = libraries.open_content_provider_id
             AND f.open_content_url_id IS NULL
 			  AND %s) AS is_favorited`
 
@@ -46,6 +49,9 @@ func (db *DB) GetAllLibraries(args *models.QueryContext, visibility string) ([]L
 		criteria, id = "f.user_id = ?", args.UserID
 	}
 	tx := db.Model(&models.Library{}).Preload("OpenContentProvider").Select(fmt.Sprintf(selectIsFavoriteOrIsFeatured, criteria), id)
+	tx = tx.Joins(`left outer join facility_visibility_statuses fvs on fvs.open_content_provider_id = libraries.open_content_provider_id
+		and fvs.content_id = libraries.id
+		and fvs.facility_id = ?`, args.FacilityID)
 	visibility = strings.ToLower(visibility)
 
 	isFeatured := false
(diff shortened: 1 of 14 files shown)


backend/src/database/libraries.go
```diff
@@ -79,21 +85,24 @@ func (db *DB) GetAllLibraries(args *models.QueryContext, visibility string, cate
...
+					AND f.user_id = ?
+			) AS is_favorited`, args.UserID)
 		if !isFeatured {
-			tx = tx.Joins(`LEFT JOIN open_content_favorites f 
+			tx = tx.Joins(`JOIN open_content_favorites f 
 				ON f.content_id = libraries.id 
 				AND f.open_content_provider_id = libraries.open_content_provider_id`)
 		}
-		tx = tx.Group("libraries.id").Order("favorite_count DESC")
+		tx = tx.Group("libraries.id, 12").Order("favorite_count DESC")
```
calisio, line 105: what does the ", 12" do? I haven't seen this in a Group clause with this included, and couldn't find anything about it in the docs.

CK-7vn (approved):
Looks good.

backend/src/database/libraries.go
carddev81: actually the 12 represents the column number of visibility_status, I used the 12 because I originally thought that I wasn't going to be able to be to use the named column 'visibility_status' because it is part of a case statement, but I checked and I am able to, I'll make sure to change this to use the name since it can be used. It's easier to read with the name.

carddev81:
@jtucholski The search issue stems from the newest iteration of Search where the library id is no longer required to be passed into the search request. The scenario where no libraries are visible and a user search is executed, a request is still sent to kiwix regardless. The logic previously always passed in a library id or ids to search through.  My plan of attack is to merge the latest code into my forked branch, make the necessary changes, and commit/push.

carddev81:
@jtucholsk I made all changes to fix search issue we spoke of. I have one question:  should the videos and library title search only return results where the content is visible or should it always display result regardless? Currently it is displaying regardless whether or not is visible.

jtucholski:
Rich and I spoke and we're going to make sure that the PR includes the change to hide results from the user if they don't have access to those libraries.

jtucholski (approved):

PR #782, opened 2025-02-25
feat: category tag dropdown to programs on admin page

### LINKED ISSUE #765: Programs Overview Page Updates (for Admins)
## Description

The following fixes and updates need to be made for the Programs page (image using seed data):

- [ ] Every card should display the following:
    - Program name & Description (using clamped text)
    - Enrollment: ***** (placeholder text)
    - Facility Availability (number of facilities its available in)
    - Number of sections (nume
- [ ] Remove the tags from the cards and list view
- [ ] Remove the Sort By and use the Tags as filter values

Image

## Acceptance criteria

- Department Admin users see a list of **every program** available in the system.
- Department Admin sees the following data when view is **All Faciltliies**
  - Program name & description
  - Enrollment number across all facilities
  - Number of facilities running the program
- Facility Admin sees the following data for their facilitiy (or Department Admin sees this data when viewing a facility)
  - Program name & description
  - Number of active sections running in the facility
  - Enrollment number for this facility
- The data on the page is sorted by program name (ascending) and can be filtered by program type.
- The user can use the filter (search?) textbox to search (case-insensitive) for specific programs with a title or description matching their text

calisio:
## Description of the change

Frontend:
- modifies programs page on admin portal to display correct information about programs, including the number of facilities it is available at and the num students enrolled (a placeholder number at the moment).
- modifies the dropdown to filter by categories (tags in the backend)

Backend:
- adds a migration to modify the open content categories table because these are the same categories used in programs.
- updates program tags table to reflect similar pattern to open content tags join table
- randomly assigns different tags to different programs in the seeder (can be updated in the future to realistically reflect what those programs are)

- **Related issues**: Closes #765 

## Screenshot(s)

Screenshot 2025-02-26 at 5 01 02 PM

---
  - https://app.asana.com/0/0/1209485070839478


backend/src/models/open_content.go
```diff
@@ -94,11 +94,9 @@ func (cp *OpenContentProvider) BeforeCreate(tx *gorm.DB) error {
...
-type OpenContentCategory struct {
+type Tag struct {
 	ID   uint   `gorm:"primaryKey" json:"key"`
 	Name string `gorm:"size:255;not null" json:"value"`
 }
 
-func (OpenContentCategory) TableName() string {
-	return "open_content_categories"
-}
+func (Tag) TableName() string { return "tags" }
```
jtucholski, line 102: Should we move this into its own model file?

frontend/src/Components/CategoryDropdownFilter.tsx
jtucholski, line 1: Do we need to be more consistent? In the backend we call them tags but in the front-end they are categories. I don't know the answer because they are serving a dual purpose.

frontend/src/Components/ProgramCard.tsx
```diff
@@ -80,13 +71,18 @@ export default function ProgramCard({
...
+                        {program.name}
+                    </ClampedText>
+                    <ClampedText as={'p'} lines={4} className="body-small h-16">
                         {program.description}
+                    </ClampedText>
+                    <p className="body-small">
+                        {enrollments} students enrolled
+                    </p>
+                    <p className="body-small">
+                        Available in {program.facilities.length} facilities
```
jtucholski, line 84: Can you update this so that it says "facility" if the value is 1?

frontend/src/Pages/ProgramManagement.tsx
```diff
@@ -2,33 +2,34 @@ import { isAdministrator, useAuth } from '@/useAuth';
...
-        return null;
-    }
-
     const [activeView, setActiveView] = useState<ViewType>(ViewType.Grid);
     const [searchTerm, setSearchTerm] = useState('');
-    const [order, setOrder] = useState('asc');
+    const [categoryQueryString, setCategoryQueryString] = useState<string>('');
     const { data, error, mutate } = useSWR<ServerResponse<Program>, AxiosError>(
-        `/api/programs?search=${searchTerm}&order=${order}`
+        `/api/programs?search=${searchTerm}&${categoryQueryString}`
```
jtucholski, line 24: Can we sort programs by name by default?

jtucholski (changes requested):
I ended up pushing a syntax change so make sure you get updates.

I want to confirm my understanding that we're using a single `tags` table going forward (instead of each entity type having its own) and creating association tables to connect the entity type to tags. Right? I think that's the right call but we should include that in the PR description.

backend/src/models/open_content.go
calisio: Yes, i was actually going to leave a comment when i made the PR but completely forgot. is that something i should do? I wasnt sure where to put it since its a more general thing

calisio:
Yes; sorry, should have been more explicit in the description. We will have a general tags table and then a join table for whatever item will use those tags, since they will probably be general throughout the application. Editing comment now.

backend/migrations/00033_alter_categories.sql
```diff
@@ -0,0 +1,29 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE public.open_content_categories RENAME TO tags;
+ALTER TABLE public.open_content_types RENAME TO open_content_tags;
+ALTER TABLE open_content_tags RENAME COLUMN category_id TO tag_id
+DROP TABLE IF EXISTS public.program_tags;
+CREATE TABLE public.program_tags (
+    tag_id integer NOT NULL,
+    program_id integer NOT NULL,
+    facility_id integer NOT NULL,
```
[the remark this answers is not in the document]
calisio: I don't think they need to have foreign keys. We didn't include foreign keys when we did the open content tags, we just had each col be a primary key.

calisio:
had to do a few force pushes because of conflicts, let me know if that messed anything up. @jtucholski all comments should be addressed

backend/migrations/00033_alter_categories.sql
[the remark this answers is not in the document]
jtucholski: Hmm. That might have been a miss from the original implementation. The FKs help ensure the consistency/integrity. @PThorpe92 Have we left FKs out in other places?

jtucholski (approved):

carddev81 (changes requested):

backend/src/database/programs.go
```diff
@@ -15,44 +15,24 @@ func (db *DB) GetProgramByID(id int) (*models.Program, error) {
...
-		}
-		if err := tx.Count(&args.Total).Error; err != nil {
-			return nil, newGetRecordsDBError(err, "programs")
-		}
-		if err := tx.Limit(args.PerPage).Offset(args.CalcOffset()).Find(&content).Error; err != nil {
-			return nil, newGetRecordsDBError(err, "programs")
-		}
+		tx = tx.Joins("JOIN program_tags t ON t.program_id = programs.id").Where("t.tag_id IN (?) AND t.facility_id = ?", args.Tags, args.FacilityID)
+	}
+	if args.OrderBy != "" && args.Order != "" {
```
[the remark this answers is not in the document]
calisio: great! Thank you- will remove

backend/src/handlers/tags.go
```diff
@@ -0,0 +1,21 @@
+package handlers
+
+import (
+	"UnlockEdv2/src/models"
+	"net/http"
+)
+
+func (srv *Server) registerTagRoutes() []routeDef {
+	axx := models.Feature(models.OpenContentAccess)
```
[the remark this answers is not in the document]
calisio: ugh that was total oversight on my part and a GREAT catch- i grabbed the axx from the open content page and totally did not realize that i had only that feature turned on

calisio:
@carddev81 should be updated now !

carddev81 (approved):
Looks great and tested good!

PR #783, opened 2025-02-27
fix: add gh action + pre commit hook to detect duplicate migration files

PThorpe92:


.github/workflows/go_test.yml
```diff
@@ -19,3 +19,14 @@ jobs:
         run: |
           cd backend
           go test -v ./...
+
+      - name: Check duplicate migration files
+        run: |
+          DUPLICATE_MIGRATIONS=$(ls -1 migrations | grep '.*\.sql' | cut -c1-5 | sort | uniq -d)
```
jtucholski, line 25: I'm having a difficult time triggering this. I know for sure that it detects it in bash, but I can't get it to abort the commit. When I run it in fish it doesn't catch the duplicate files since it uses eza. Is this a concern or is it something we should coach the team not to do?

docker-compose.yml
```diff
@@ -173,24 +156,6 @@ services:
       - config:/config
       - nats_data:/data
 
-  keto:
```
carddev81, line 176: Just a question here:  Curious as to what this was used for and why we are getting rid of it? Also should the environment variable KETO_ADMIN_URL be removed from the /config/helmchart/templates/server-deployment.yaml file?

carddev81 (approved):
Looks good, just had a question and a possible change that you may have to make.

docker-compose.yml
PThorpe92: we don't use the service yet at this point, so its just running a container for no reason. yeah i'll remove that other env var too, good catch

.github/workflows/go_test.yml
PThorpe92: bash is really the desired shell here, because that is what will run in CI... Yeah I forgot that on push, all we can do is have it fail the test, we cannot have it abort the commit. but what i can do is have it run b4 we build + deploy containers

PR #787, opened 2025-03-04
feat: add proper tab separation change trending content to home

CK-7vn:
## Description of the change
This pull request adds more tab separation in our "TabView" so that on the parts of the application that uses tabs there is more intuitive tab navigation. It also changes StudentLayer0 (which is essentially if the user has 0 features enabled) to "/temp-home" and added a "Contact your administrators" dialogue to help assist them in why they're at that page. Also, StudentLayer1 (trending content) was changed to "ResidentHome", and also assigned the "/home" route. 

- **Related issues**: Closes 32 and 57 in Asana

## Screenshot(s)
image
image
image
image
**This is now the StudentLayer0 when no features are available**
image

### DIFF
--- a/frontend/src/Components/TabView.tsx
+++ b/frontend/src/Components/TabView.tsx
@@ -22,7 +22,7 @@ export default function TabView({
                         className={`focus:outline-none py-3 px-8 ${
                             isActive
                                 ? 'font-bold text-teal-4 border-b-2 border-teal-4'
-                                : 'hover:scale-105 hover:text-teal-3 hover:drop-shadow hover:border-b-2 hover:border-teal-3'
+                                : 'hover:text-teal-3 hover:drop-shadow hover:border-b-2 hover:border-teal-3'
                         }`}
                     >
                         {tabName}
--- a/frontend/src/Components/TabView.tsx
+++ b/frontend/src/Components/TabView.tsx
@@ -10,19 +10,20 @@ export default function TabView({
     setActiveTab: (tab: Tab) => void;
 }) {
     return (
-        <div className="flex flex-row gap-16 w-100 border-b-2 border-grey-2 py-3">
-            {tabs.map((tab: Tab, index: number) => {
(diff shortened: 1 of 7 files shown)


frontend/src/Components/TabView.tsx
```diff
@@ -10,19 +10,20 @@ export default function TabView({
     setActiveTab: (tab: Tab) => void;
 }) {
     return (
-        <div className="flex flex-row gap-16 w-100 border-b-2 border-grey-2 py-3">
-            {tabs.map((tab: Tab, index: number) => {
+        <div className="flex flex-row gap-16 w-full border-b-2 border-gray-300 py-3">
```
calisio, line 13: revert these changes back

calisio: okay, i was playing around with these UI changes to make it more consistent with what other applications have implemented, while still maintaining a pretty modern ui feel. I came up with this
```
        <div className="flex flex-row w-100 border-b-2 border-grey-2">
            {tabs.map((tab, index) => {
                const tabName =
                    tab.name.charAt(0).toUpperCase() + tab.name.slice(1);
                const isActive = activeTab.value === tab.value;
                return (
                    <button
                        key={index}
                        onClick={() => setActiveTab(tab)}
                        className={`focus:outline-none py-3 px-8 ${
                            isActive
                                ? 'font-bold text-teal-4 border-b-2 border-teal-4'
                                : 'hover:scale-105 hover:text-teal-3 hover:drop-shadow hover:border-b-2 hover:border-teal-3'
                        }`}
                    >
                        {tabName}
                    </button>
                );
            })}
        </div>
```
The image of what this looks like is below:
    
image

and on hover of a tab (like helpful links), we have it below:

image

This also moves the padding onto the button itself so it increases the click range to trigger the action. What do you think about this Keighan?

calisio (approved):

calisio (approved):

jtucholski (approved):

PR #792, opened 2025-03-09
feat: modify helpful link management so helpful links are managed and defined at facility level

carddev81:
## Description of the change

Implemented logic to modify the Helpful link management so that helpful links are no longer system-wide but rather defined at the level of a facility. 

Modified SQL queries to account for helpful links being managed at the facility level.

- **Related issues**: Link to Asana ticket: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209485070836361?focus=true

### DIFF
--- a/backend/src/database/helpful_links.go
+++ b/backend/src/database/helpful_links.go
@@ -11,42 +11,44 @@ type HelpfulLinkResp struct {
 	IsFavorited bool `json:"is_favorited"`
 }
 
-func (db *DB) GetHelpfulLinks(page, perPage int, search, orderBy string, onlyVisible bool, userID uint) (int64, []HelpfulLinkResp, error) {
-	links := make([]HelpfulLinkResp, 0, perPage)
+func (db *DB) GetHelpfulLinks(args *models.QueryContext, onlyVisible bool) ([]HelpfulLinkResp, error) {
+	links := make([]HelpfulLinkResp, 0, args.PerPage)
 
 	subQuery := db.Table("open_content_favorites f").
 		Select("1").
 		Where(`f.content_id = helpful_links.id AND f.user_id = ?
-			AND f.open_content_provider_id = helpful_links.open_content_provider_id`, userID)
+			AND f.open_content_provider_id = helpful_links.open_content_provider_id`, args.UserID)
 	tx := db.Model(&models.HelpfulLink{}).Select("helpful_links.*, EXISTS(?) as is_favorited", subQuery)
-	var total int64
 
 	if onlyVisible {
 		tx = tx.Where("visibility_status = ?", true)
 	}
 
-	if search != "" {
-		search = "%" + strings.ToLower(search) + "%"
-		tx = tx.Where("LOWER(title) LIKE ?", search)
+	tx = tx.Where("helpful_links.facility_id = ?", args.FacilityID)
+
+	if args.Search != "" {
+		args.Search = "%" + strings.ToLower(args.Search) + "%"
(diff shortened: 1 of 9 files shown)


backend/src/database/helpful_links.go
```diff
@@ -11,42 +11,43 @@ type HelpfulLinkResp struct {
...
+	links := make([]HelpfulLinkResp, 0, args.PerPage)
 
 	subQuery := db.Table("open_content_favorites f").
 		Select("1").
 		Where(`f.content_id = helpful_links.id AND f.user_id = ?
-			AND f.open_content_provider_id = helpful_links.open_content_provider_id`, userID)
-	tx := db.Model(&models.HelpfulLink{}).Select("helpful_links.*, EXISTS(?) as is_favorited", subQuery)
-	var total int64
+			AND f.open_content_provider_id = helpful_links.open_content_provider_id`, args.UserID)
+	tx := db.Debug().Model(&models.HelpfulLink{}).Select("helpful_links.*, EXISTS(?) as is_favorited", subQuery)
```
CK-7vn, line 21: probably can remove this debug here?

backend/src/database/helpful_links.go
```diff
@@ -11,42 +11,43 @@ type HelpfulLinkResp struct {
...
+	tx = tx.Where("facility_id = ?", args.FacilityID)
+
+	if args.Search != "" {
+		args.Search = "%" + strings.ToLower(args.Search) + "%"
+		tx = tx.Where("LOWER(title) LIKE ?", args.Search)
 	}
-	switch strings.ToLower(orderBy) {
+	switch strings.ToLower(args.OrderBy) {
 	case "most_popular":
 		tx = tx.Order("COUNT(f.id) DESC")
```
CK-7vn, line 35: i know this isn't/wasn't you, but the tables not even joined here is it? will this cause problems?

CK-7vn (commented):
just the debug and then your opinion on that other most popular join issue.

backend/src/database/helpful_links.go
carddev81: good catch! I use the Debug() function religiously...making sure GORM queries are doing what they say.

backend/src/database/helpful_links.go
carddev81: You are correct this won't work, even though it's never called I went ahead and coded it similar to videos.

CK-7vn (approved):

backend/src/database/helpful_links.go
```diff
@@ -11,42 +11,44 @@ type HelpfulLinkResp struct {
...
 	}
-	switch strings.ToLower(orderBy) {
+	switch strings.ToLower(args.OrderBy) {
 	case "most_popular":
-		tx = tx.Order("COUNT(f.id) DESC")
+		tx = tx.Joins("LEFT JOIN open_content_favorites f ON f.content_id = helpful_links.id AND f.open_content_provider_id = helpful_links.open_content_provider_id").
+			Group("helpful_links.id").Order("COUNT(f.id) DESC")
 	default:
-		tx = tx.Order(orderBy)
+		tx = tx.Order(args.OrderBy + " " + args.Order)
```
PThorpe92, line 38: hmmmm this gets me thinking..

we should add a method to `QueryContext` to get the order clause that does this automatically

carddev81: good call! I'll make this change.

PThorpe92: lmao i added it in my PR too

PThorpe92 (approved):

PR #793, opened 2025-03-10
Add Audit trail

corypride:
## Description of the change

Created an audit trail that records all critical administrative actions across the system for facilities, videos, libraries, programs, and users.

- **Related issues**: 
Implement Admin Auditing to Track Admin Actions

### DIFF
--- a/backend/src/handlers/auth.go
+++ b/backend/src/handlers/auth.go
@@ -18,7 +18,6 @@ const (
 	ClaimsKey         contextKey = "claims"
 	KratosBrowserFlow string     = "/self-service/login/browser"
 	AuthCallbackRoute string     = "/authcallback"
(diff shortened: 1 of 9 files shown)


backend/src/handlers/auth.go
```diff
@@ -132,7 +133,9 @@ func (srv *Server) adminMiddleware(next http.Handler) http.Handler {
 			srv.errorResponse(w, http.StatusUnauthorized, "Unauthorized - not admin")
 			return
 		}
-		next.ServeHTTP(w, r)
+
+		ctx := context.WithValue(r.Context(), AuditKey, struct{}{})
```
PThorpe92, line 137: this will include it for all admin requests. we need to filter out GET requests

PThorpe92 (approved):

PR #795, opened 2025-03-11
feat: add welcome tour

calisio:
## Description of the change

Adds welcome tour to platform. Right now, the temporary button is on the top left corner on the Navbar, where it says "Start Tour". Make sure you are on the home page to start it (working on the fix for this now). Leaving this as a draft for now so I can receive feedback as I make small changes to ui/some phrasing/fix some small bugs here and there. There shouldn't be any major changes.

The start button will be moved to the "get help" section which rich is working on, which is why it has a temporary spot at the moment. I am also still implementing it starting on first log in.

PS- I know that on the second to last tooltip it focuses weird, that is one of the bugs I am trying to fix :)

- **Related issues**: Closes asana task https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209478696323307 

## Screenshot(s)

https://github.com/user-attachments/assets/3da766d2-d20b-42a6-8431-e22dd2123929

## Additional Context

**NOTE: TO BE ABLE TO RUN THIS, YOU MUST UPDATE YOUR DEPENDENCIES!** Running `yarn install` once you are in the frontend folder should do it.

### DIFF
--- a/frontend/src/Components/UnlockEdTour.tsx
+++ b/frontend/src/Components/UnlockEdTour.tsx
@@ -162,7 +162,6 @@ export default function UnlockEdTour() {
 
     const handleCallback = (data: CallBackProps) => {
         const { action, index, type } = data;
-        console.log(data);
         if (
             action === 'close' ||
             action === 'skip' ||
--- a/frontend/src/Components/UnlockEdTour.tsx
+++ b/frontend/src/Components/UnlockEdTour.tsx
@@ -1,0 +1,199 @@
+import { ThemeContext } from '@/Context/ThemeContext';
+import { useTourContext } from '@/Context/TourContext';
+import { useContext, useEffect } from 'react';
+import Joyride, { CallBackProps, EVENTS } from 'react-joyride';
+import { useNavigate } from 'react-router-dom';
+
+export default function UnlockEdTour() {
+    const {
+        tourState: { run, stepIndex, steps },
+        setTourState
+    } = useTourContext();
+    const { theme } = useContext(ThemeContext);
+    const navigate = useNavigate();
+
+    const style =
+        theme === 'dark'
+            ? {
+                  options: {
+                      arrowColor: '#1A3F3B',
+                      backgroundColor: '#1A3F3B',
+                      primaryColor: '#14958A',
+                      textColor: '#EEEEEE'
+                  }
+              }
+            : {
+                  options: {
+                      arrowColor: '#FFFFFF',
+                      backgroundColor: '#FFFFFF',
+                      primaryColor: '#18ABA0',
+                      textColor: '#222222'
+                  }
+              };
+
+    useEffect(() => {
+        setTourState({
+            steps: [
+                {
+                    target: '#resident-home',
+                    content:
+                        "Welcome to UnlockEd, a place for you to access and explore educational content. Let's take a quick tour to show you how things work.",
+                    disableBeacon: true,
+                    showSkipButton: true,
+                    placement: 'center'
+                },
+                {
+                    target: '#visit-knowledge-center',
+                    content:
+                        'Click on Knowledge Center to access a variety of resources. Try it now!',
+                    disableOverlayClose: true,
+                    disableBeacon: true,
+                    spotlightClicks: true,
+                    hideCloseButton: true,
+                    hideFooter: true
+                },
+                {
+                    target: '#knowledge-center-landing',
+                    content:
+                        'Welcome to the Knowledge Center. Here, you can find a range of resources in both written and video format, as well as helpful links.',
+                    disableBeacon: true,
+                    disableOverlayClose: true,
+                    placement: 'center'
+                },
+                {
+                    target: '#knowledge-center-tabs',
+                    content:
+                        'Use these tabs to switch between different types of resources.',
+                    disableBeacon: true,
+                    disableOverlayClose: true
+                },
+                {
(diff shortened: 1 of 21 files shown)


frontend/src/Components/modals/FormModal.tsx
calisio, line 1: ignore- will be removed

calisio:
Issue (& semi a note to self): do we want to prioritize favoriting inside of the library or outside of the library? Right now, if we favorite outside of the library, when we click into it then we already have that favorited. I would do 2 different libraries, but it isn't guaranteed that there will be more than 1 library turned on at a facility.

frontend/src/Components/UnlockEdTour.tsx
```diff
@@ -0,0 +1,199 @@
...
+                    content:
+                        'The star allows you to favorite the library for easy access later. Try favoriting a library to save it for later!',
+                    disableBeacon: true,
+                    spotlightClicks: true,
+                    disableOverlayClose: true
+                },
+                {
+                    target: '#knowledge-center-enter-library',
+                    content:
+                        "Click on a library to open it and see what's inside. Try it now!",
```
carddev81, line 104: noticed some content is wrapped in ' and others in " ... just an inconsistency i noticed

frontend/yarn.lock
```diff
@@ -399,6 +399,16 @@
   dependencies:
     levn "^0.4.1"
 
+"@gilbarbara/deep-equal@^0.1.1":
```
carddev81, line 402: Just curious what the yarn.lock file is used for?  Noticed all these changes

carddev81 (approved):
Awesome tour!

carddev81:
Should this be a part of the tour?

image

frontend/src/Components/UnlockEdTour.tsx
calisio: Yes, its because the ones that are wrapped in "" have an apostrophe on them (so that escapes it). When I try to wrap with "" for all though, because of our prettier configuration, it automatically changes them back to ''.

calisio:
> Should this be a part of the tour? It is the favorites on the Resident Home page.
> 
> image

We decided to leave the favorites sidebar out of the tour because we thought it was an easy step to cut out (we don't want the user to get bored and not finish the tour because of too many steps). Because favorites is already mentioned (where it says using the star will "favorite" the item, we will assume the user can make the connection and know that the "Favorites" sidebar will contain the items they have favorited.

frontend/yarn.lock
PThorpe92: its to keep dependencies locked at specific versions. its normal to change when adding deps :+1:

calisio: I believe it is for the `react-joyride` library!

calisio:
Go tests are ~ probably ~ failing because of the fail in main currently (I rebased with main recently), so will be fixed once main gets fixed.

frontend/src/Components/forms/LoginForm.tsx
```diff
@@ -35,8 +43,10 @@ export default function LoginForm() {
             data
         )) as ServerResponseOne<AuthResponse>;
         if (resp.success) {
-            window.location.href =
-                resp.data.redirect_to ?? resp.data.redirect_browser_to;
+            if (resp.data.first_login) {
+                setTourState({ tourActive: true });
+            }
+            navigate(resp.data.redirect_to ?? resp.data.redirect_browser_to);
```
PThorpe92, line 49: So we unfortunately cannot just swap out in favor of `navigate()` all together, because if it's coming from kratos/hydra in the event that it's an Oauth2 redirect, that will be a full `https://example.com/redirect`, which will break when used with `navigate`.

So we essentially need to find out if it's a relative URI, and if it is, then we can `navigate`, otherwise we need to `window.location.href =`

frontend/src/Components/forms/ChangePasswordForm.tsx
```diff
@@ -73,7 +74,7 @@ export default function ChangePasswordForm() {
             data
         )) as ServerResponseOne<AuthResponse>;
         if (response.success) {
-            window.location.href = response.data.redirect_to;
+            navigate(response.data.redirect_to);
```
PThorpe92, line 77: I think this is safe however (see other comment), as the response will always be
```go
resp["redirect_to"] = AuthCallbackRoute
```

PThorpe92 (changes requested):

frontend/src/Components/forms/LoginForm.tsx
calisio: Ahh got it, okay so if it is `resp.data.redirect_to`, then it should always be the relative URI, right? If not, I can just do a check if it's `/reset-password` to make sure its relative

carddev81 (approved):
Looks great! Tested good!

frontend/src/Components/forms/LoginForm.tsx
calisio: Okay updated 👍

frontend/src/Components/forms/ChangePasswordForm.tsx
calisio: not checking here since you said it should be okay, but if you want me to just in case let me know

jtucholski:
Functionally looks & works great! Like the cross-over with the Help Center!

PThorpe92 (approved):

PR #796, opened 2025-03-11
fix: display proper form on initial password reset, add logs

PThorpe92:
cleans up a couple things from adding the audit logs, and re-enables the new facility name + timezone form when the SuperAdmin makes their password for the first time. 
also follows new Go convention/warnings and replaces interface with any type in a few areas.

### DIFF
--- a/frontend/src/Components/forms/ChangePasswordForm.tsx
+++ b/frontend/src/Components/forms/ChangePasswordForm.tsx
@@ -4,16 +4,9 @@ import InputError from '../../Components/inputs/InputError';
 import { CheckIcon, XMarkIcon } from '@heroicons/react/24/solid';
 import { DropdownInput, TextInput } from '../inputs';
 import API from '@/api/api';
-import {
-    AuthResponse,
-    Facility,
-    ServerResponseOne,
-    UserRole,
-    Timezones,
-    User
-} from '@/common';
-import { useLoaderData } from 'react-router-dom';
-import { fetchUser } from '@/useAuth';
+import { AuthResponse, ServerResponseOne, UserRole, Timezones } from '@/common';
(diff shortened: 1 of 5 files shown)


frontend/src/Components/forms/ChangePasswordForm.tsx
```diff
@@ -60,7 +57,8 @@ export default function ChangePasswordForm() {
     const validFacility =
         facility && facility.length > 2 && facility.trim().length > 2;
     const isFirstAdminLogin =
-        user?.role === UserRole.SystemAdmin && loaderData?.name === 'Default';
+        user?.role === UserRole.SystemAdmin &&
+        user?.facility_name === 'Default';
```
jtucholski, line 62: @PThorpe92 I'm getting undefined for user at this point and not seeing the timezone drop-down. Assuming this is because it's running before fetchUser returns a value?

CK-7vn (approved):

frontend/src/Components/forms/ChangePasswordForm.tsx
carddev81, line 1: The undefined issue is due to the `void` keyword on the `fetchUser` call. This basically throws away fetchUser's return value.  Since fetchUser returns a promise you should be able to do something like:

`fetchUser().then(user => setUser(user)).catch(err => console.log("Error occurred fetching user. Error is: ", err));`

carddev81 (changes requested):

frontend/src/Components/forms/ChangePasswordForm.tsx
PThorpe92: yup i was speeding, it needs to be a state variable and set in the useEffect

PR #798, opened 2025-03-12
fix: add css for fixing clipped text on the trending content when screen size is small

carddev81:
## Description of the change

Modified CSS on the LibraryCard.tsx to handle fixing the clipped text on the trending content when the screen size is small.

- **Related issues**: Link to Asana ticket: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209485070839643?focus=true

## Screenshot(s)
[demo.webm](https://github.com/user-attachments/assets/c25c2367-ee4d-4b13-af2e-e5f0905c377e)

### DIFF
--- a/frontend/src/Components/LibraryCard.tsx
+++ b/frontend/src/Components/LibraryCard.tsx
@@ -92,7 +92,10 @@ export default function LibraryCard({
                             alt={`${library.title} thumbnail`}
                         />
                     </figure>
-                    <ClampedText as="h3" className="w-3/4 body my-auto mr-7">
+                    <ClampedText
+                        as="h3"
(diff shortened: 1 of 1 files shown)


calisio:
Im feeling like for this maybe we should use [`word-break`](https://tailwindcss.com/docs/word-break) or something of the sort from tailwind? This will allow the word to be broken up so that it doesn't get chopped off halfway through. I think thats what the ticket is looking for?

jtucholski:
@calisio I removed the absolute positioning on this because we were getting some really rough looking overlap of icons and text. I think we should use this and address some more responsive approaches in a future change.

jtucholski (approved):

PThorpe92 (approved):

PR #802, opened 2025-03-18
feat: add sql migrations for program features development

carddev81:
## Description of the change

Added SQL migration for program features development.

### DIFF
--- a/backend/migrations/00040_alter_programs.sql
+++ b/backend/migrations/00040_alter_programs.sql
@@ -1,36 +1,73 @@
 -- +goose Up
 -- +goose StatementBegin
-ALTER TABLE public.programs ADD COLUMN program_owner VARCHAR(255);
-ALTER TABLE public.programs ADD COLUMN funding_type VARCHAR(50);
-ALTER TABLE public.facilities_programs ADD COLUMN archived_at timestamp with time zone;
+CREATE TYPE public.program_type AS ENUM (
+    'Educational',
+    'Mental_Health_Behavioral',
+    'Religious_Faith-Based',
+    'Re-Entry',
+    'Life_Skills'
+);
+CREATE TYPE public.funding_type AS ENUM (
+    'Federal_Grants',
+    'State_Grants',
+    'Nonprofit_Organizations',
+    'Educational_Grants',
+    'Inmate_Welfare_Funds',
+    'Other'
+);
+ALTER TABLE public.programs DROP COLUMN credit_type;
+ALTER TABLE public.programs DROP COLUMN program_type;
+ALTER TABLE public.programs ADD COLUMN program_type program_type;
+ALTER TABLE public.programs ADD COLUMN funding_type funding_type;
+ALTER TABLE public.program_tags RENAME TO program_types;
+ALTER TABLE public.program_types DROP COLUMN facility_id;
+ALTER TABLE public.facilities_programs ADD COLUMN program_owner VARCHAR(255);
 ALTER TABLE public.program_sections ADD COLUMN capacity INT;
 ALTER TABLE public.program_sections ADD COLUMN name VARCHAR(255);
 ALTER TABLE public.program_sections ADD COLUMN instructor_name VARCHAR(255);
 ALTER TABLE public.program_sections ADD COLUMN description TEXT;
 ALTER TABLE public.program_sections ADD COLUMN archived_at timestamp with time zone;
 ALTER TABLE public.program_sections ADD COLUMN start_dt date;
-ALTER TABLE public.program_sections ADD COLUMN duration VARCHAR(32);
-ALTER TABLE public.program_sections ADD COLUMN total_hours INT;
+ALTER TABLE public.program_sections ADD COLUMN duration VARCHAR(32) NOT NULL DEFAULT '';
+ALTER TABLE public.program_sections ADD COLUMN status VARCHAR(50);
+ALTER TABLE public.program_sections ADD COLUMN credit_hours INT;
 ALTER TABLE public.program_sections ADD COLUMN is_active boolean;
 ALTER TABLE public.program_section_enrollments ADD COLUMN enrollment_status VARCHAR(255);
 ALTER TABLE public.program_section_events RENAME COLUMN location TO room;
 
+CREATE TYPE public.credit_type AS ENUM (
+    'Completion',
+    'Participation',
+    'Earned-time',
+    'Education'
+);
+
+CREATE TABLE public.program_credit_types (
+    program_id integer NOT NULL,
+    credit_type credit_type NOT NULL,
+
+    FOREIGN KEY (program_id) REFERENCES public.programs(id) ON UPDATE CASCADE ON DELETE CASCADE,
+
+    PRIMARY KEY (program_id, credit_type) 
+);
+
 CREATE TABLE public.program_completions (
-        id SERIAL NOT NULL PRIMARY KEY,
-        program_section_id integer NOT NULL,
-        program_id integer NOT NULL,
-        facility_name CHARACTER VARYING(255) NOT NULL,
-        credit_type CHARACTER VARYING(255) NOT NULL, 
-        program_owner CHARACTER VARYING(255) NOT NULL,
-        program_name CHARACTER VARYING(255), 
-        program_section_name CHARACTER VARYING(255),
-        program_section_start_dt CHARACTER VARYING(255),
-        created_at timestamp with time zone,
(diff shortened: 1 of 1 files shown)


backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE public.programs ADD COLUMN program_owner VARCHAR(255);
+ALTER TABLE public.programs ADD COLUMN funding_type VARCHAR(50);
```
jtucholski, line 4: It sounds like Program Type, Funding Type, and Credit Type are going to be multi-select. 
I think Program Type can mean one of many categories (tags?).
How do we want to handle multiple values for credit type and funding type? Does it need to be with many-to-many relation? Or should we have a Programs_CreditTypes (program_id, credit_type) and a Programs_FundingTypes (program_id, funding_type) table and use enums for the values?

backend/migrations/00040_alter_programs.sql
jtucholski, line 1: Looking good. We didn't get to this when we spoke, but we need to consider a few more changes for programs I think:
- Program Type is multi-select (Educational, Vocational, etc.). I think this is similar to the `tags` table that was created for this?
- Credit Type is multi-select and listed here as a varchar. Do we need a many-to-many table or a programs_credit_types table where we keep the program id with the credit type value?
- Funding Type (same as above)
- User-defined tags for Program Type
  - The Administrator might add their own tags when creating the program. 
  - The `tags` table might work here but right now it has system-defined tags. Should we update it to include a flag for user-defined tags and system-defined tags?

backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
...
+ALTER TABLE public.facilities_programs ADD COLUMN archived_at timestamp with time zone;
```
jtucholski, line 5: Did we mean to omit program_owner for the facilities_programs relationship? We need some way to indicate who is accountable at a facility for a given program. Maybe we can add this and make it optional?

If the archived_at value has a timestamp does that mean the program is no longer offered at the facility? And if we start it back up do we NULL out the archived_at column then?

backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
...
+ALTER TABLE public.program_sections ADD COLUMN capacity INT;
+ALTER TABLE public.program_sections ADD COLUMN name VARCHAR(255);
+ALTER TABLE public.program_sections ADD COLUMN instructor_name VARCHAR(255);
+ALTER TABLE public.program_sections ADD COLUMN description TEXT;
+ALTER TABLE public.program_sections ADD COLUMN archived_at timestamp with time zone;
+ALTER TABLE public.program_sections ADD COLUMN start_dt date;
+ALTER TABLE public.program_sections ADD COLUMN duration VARCHAR(32);
+ALTER TABLE public.program_sections ADD COLUMN total_hours INT;
+ALTER TABLE public.program_sections ADD COLUMN is_active boolean;
```
jtucholski, line 14: How do this design allow for program sections to be 
- Scheduled
- Active
- Completed
- Cancelled
- Archived
- (eventually Paused?)

I think I can infer that:
- SCHEDULED (start_dt is in the future)
- ACTIVE (start_dt is in the past?)
- Not sure how to infer if something is completed or cancelled or paused.

backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
...
+ALTER TABLE public.programs ADD COLUMN program_owner VARCHAR(255);
+ALTER TABLE public.programs ADD COLUMN funding_type VARCHAR(50);
+ALTER TABLE public.facilities_programs ADD COLUMN archived_at timestamp with time zone;
+ALTER TABLE public.program_sections ADD COLUMN capacity INT;
+ALTER TABLE public.program_sections ADD COLUMN name VARCHAR(255);
+ALTER TABLE public.program_sections ADD COLUMN instructor_name VARCHAR(255);
+ALTER TABLE public.program_sections ADD COLUMN description TEXT;
+ALTER TABLE public.program_sections ADD COLUMN archived_at timestamp with time zone;
+ALTER TABLE public.program_sections ADD COLUMN start_dt date;
+ALTER TABLE public.program_sections ADD COLUMN duration VARCHAR(32);
```
jtucholski, line 12: Does NULL here indicate the program has no end date?

backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
...
+ALTER TABLE public.program_sections ADD COLUMN is_active boolean;
+ALTER TABLE public.program_section_enrollments ADD COLUMN enrollment_status VARCHAR(255);
+ALTER TABLE public.program_section_events RENAME COLUMN location TO room;
+
+CREATE TABLE public.program_completions (
+        id SERIAL NOT NULL PRIMARY KEY,
+        program_section_id integer NOT NULL,
+        program_id integer NOT NULL,
+        facility_name CHARACTER VARYING(255) NOT NULL,
+        credit_type CHARACTER VARYING(255) NOT NULL,
```
jtucholski, line 23: If we support many credit types does this need to? Can it just be a comma-separated list since it's the final output? Would that be a bad idea though?

backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
...
+        deleted_at timestamp with time zone,
+        
+        FOREIGN KEY (program_section_id) REFERENCES public.program_sections(id) ON UPDATE CASCADE ON DELETE CASCADE,
+        FOREIGN KEY (program_id) REFERENCES public.programs(id) ON UPDATE CASCADE ON DELETE CASCADE
+);
+
+CREATE TABLE public.programs_sections_history (
+    id SERIAL NOT NULL PRIMARY KEY,
+    parent_ref_d INT,
+    table_name VARCHAR(255),
```
jtucholski, line 39: What is table_name for?

backend/migrations/00040_alter_programs.sql
PThorpe92: thats tough.. do we allow for arbitrary input for the owner or enforce a FK and assume that the user has an account.

yes archived_at would mean that the user removed it from the facility, we would NULL the field if un-removed (in theory it would still be available to show all the previous programs offered at each facility

carddev81: Good catch, yes I wasn't thinking about this correctly here, I added the column 'program_owner' to the programs table when it should have went into the facilities_programs table.

The second question, yes archived_at means that the program is no longer offered at the facility. If we start it back up I think a new record should be created in facilities_programs since that previous program version has been archived?

backend/migrations/00040_alter_programs.sql
```diff
@@ -0,0 +1,90 @@
...
+ALTER TABLE public.programs ADD COLUMN funding_type VARCHAR(50);
+ALTER TABLE public.facilities_programs ADD COLUMN archived_at timestamp with time zone;
+ALTER TABLE public.program_sections ADD COLUMN capacity INT;
+ALTER TABLE public.program_sections ADD COLUMN name VARCHAR(255);
+ALTER TABLE public.program_sections ADD COLUMN instructor_name VARCHAR(255);
+ALTER TABLE public.program_sections ADD COLUMN description TEXT;
+ALTER TABLE public.program_sections ADD COLUMN archived_at timestamp with time zone;
+ALTER TABLE public.program_sections ADD COLUMN start_dt date;
+ALTER TABLE public.program_sections ADD COLUMN duration VARCHAR(32);
+ALTER TABLE public.program_sections ADD COLUMN total_hours INT;
```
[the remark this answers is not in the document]
carddev81: So, total hours would be for the facility admin to fill out on how many hours the class consists of and it would be optional to fill out otherwise the duration would be calculated based off of the program_section_events associated to the program_sections?

backend/migrations/00040_alter_programs.sql
carddev81: I think comma separated is fine, @PThorpe92 what are your thoughts since there can be many credit types to one program?

backend/migrations/00040_alter_programs.sql
carddev81: Yes, both tables will be logging its history into this table.

jtucholski (approved):

PR #803, opened 2025-03-18
fix: modify structs and seeder to conform to program database structure

carddev81:
## Description of the change

Made changes to fix seeder and go tests due to migration changes.


CK-7vn (approved):

jtucholski:
This does not include changes that resolve the issues we were experiencing with migrate down, right?

carddev81:
Right, no this does not include the fix for migrate down.

jtucholski (approved):

PR #807, opened 2025-03-21
feat: add functionality to create and schedule new section for facility admin per ticket id 47

carddev81:
## Description of the change

NOTE: Please before testing this change to run `make reset`  and then run the seeder.

Added functionality for allowing administrators to create program sections for a selected program. This change required SQL to be created, backend api changes, and several frontend changes such as creating date, number, duration, and reset button form components. 

- **Related issues**: Link to Asana task: https://app.asana.com/0/1209460078641109/1209485911445028/f.

## Screenshot(s)
AddClass
NavigationAfterSaving

### DIFF
--- a/backend/src/handlers/programs_handler.go
+++ b/backend/src/handlers/programs_handler.go
@@ -16,6 +16,7 @@ func (srv *Server) registerProgramsRoutes() []routeDef {
 		{"DELETE /api/programs/{id}", srv.handleDeleteProgram, true, axx},
 		{"PATCH /api/programs/{id}", srv.handleUpdateProgram, true, axx},
 		{"PUT /api/programs/{id}/save", srv.handleFavoriteProgram, false, axx},
+		{"GET /api/programs/{id}/overview", srv.handleShowProgamOverview, true, axx},
 	}
 }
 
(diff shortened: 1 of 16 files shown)


backend/src/handlers/programs_handler.go
```diff
@@ -50,6 +51,43 @@ func (srv *Server) handleShowProgram(w http.ResponseWriter, r *http.Request, log
 	return writeJsonResponse(w, http.StatusOK, program)
 }
 
+type ProgramOverviewResponse struct {
```
PThorpe92, line 54: not using the `PaginatedResponse` struct means that we can't use the `API` class on the frontend. we have made a couple small exceptions for cases where it really made sense to have a custom one but if possible I think we should stick to it.

backend/src/handlers/sections_handler.go
```diff
@@ -56,12 +56,20 @@ func (srv *Server) handleIndexSectionsForFacility(w http.ResponseWriter, r *http
 }
 
 func (srv *Server) handleCreateSection(w http.ResponseWriter, r *http.Request, log sLog) error {
+	id, err := strconv.Atoi(r.PathValue("id"))
```
PThorpe92, line 59: we need to make sure to `log.add` a bunch here since it's an action that will emit `Audit` logs

backend/src/handlers/programs_handler.go
carddev81: okay, i'll make the change to use the paginatedreponse. I knew there was going to be another list developed in the future, to be added to this struct for program outcomes and figured this would be the way to go.

backend/src/handlers/sections_handler.go
carddev81: will do.

backend/src/models/program_sections.go
```diff
@@ -53,3 +58,41 @@ type ProgramSectionEnrollment struct {
...
+	FacilityName   string    `json:"facility_name"`
+	InstructorName string    `json:"instructor_name"`
+	StartDt        time.Time `json:"start_dt"`
+	Duration       string    `json:"duration"`
+	Capacity       string    `json:"capacity"`
+	Enrolled       int       `json:"enrolled"`
+	EndDt          time.Time `json:"end_dt"`
+}
+
+func (psd *ProgramSectionDetail) CalcuateEndDt() error {
```
jtucholski, line 73: This is a derived value, right? Never persisted? 

I don't have good requirements for this situation yet, but if the class is paused for some reason and then later resumed, does the duration or start date need to be updated? 

Do we have to account for classes with no end date? What value would this use for that?

carddev81: Yes, this is a derived value. Based on the calculation of start_dt and duration.  Good point on pausing the class. If that was the case then duration would either have to be changed to calculate the correct end date?  I believe having the user enter an actual end date would be better than adding duration here? I think the user would get confused about what duration really means at that point? End date seems clear.  Didn't account for classes with no end date ->7799-12-31?

frontend/src/Pages/SectionManagementForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+            start_dt: new Date(data.start_dt),
+            capacity: Number(data.capacity),
+            credit_hours: Number(data.credit_hours)
+        };
+        const response = await API.post(
+            `programs/${id}/sections`,
+            formattedJson
+        );
+        if (!response.success) {
+            //should we put a toast message here??
```
jtucholski, line 54: Yes, we have a Toast Message getting displayed after creating a new program and navigating the user to the program overview page.

backend/src/models/program_sections.go
jtucholski: I can't remember why, but I wondered if we should derive this value on the front end then. It felt almost out of place on the back end.

carddev81: i made the move, the calculation is done within the frontend now, took all logic out of the backend.

carddev81:
@jtucholski pushed latest changes to include removing duration column and adding end_dt column to program_sections table. Also made requested layout changes per Lisa.

jtucholski (approved):

PThorpe92 (approved):

PR #808, opened 2025-03-21
feat: add search suggestions

PThorpe92:
This PR introduces search query suggestions to help mitigate user spelling errors and reduce the amount of improper FTS requests sent to kiwix

image


frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -235,6 +301,41 @@ const LibrarySearchResultsModal = forwardRef<
...
+                                        >
+                                            {sug}
+                                        </button>
+                                    </li>
+                                ))}
+                            </ul>
+                            {suggestionError && (
+                                <p className="text-error text-sm mt-1">
+                                    {suggestionError}
+                                </p>
```
jtucholski, line 335: I'm on the fence as to whether we display a suggestion error not. It seems like it's a service we're using to enhance the user's experience but do we want to draw attention to it if it doesn't work for some reason?

CK-7vn (approved):

PR #810, opened 2025-03-25
feat: add section status modification and archiving

calisio:
## Description of the change

Admins are now able to manually update the status of the sections, which are displayed inside the program from the program management page. Some notes to consider before merging this PR are below.

- **Related issues**: [Change section status](https://app.asana.com/0/1209460078641109/1209693986941448/f)

## Screenshot(s)
Brief video of the functionality:

https://github.com/user-attachments/assets/c8b21519-fc66-40cc-bbae-1bce70d5dc0c

Furthermore, if the section is not allowed to be archived due to having students and either being in the ACTIVE or SCHEDULED states, the following will appear when they attempt to archive the section. This will NOT allow the invalid section to be archived, but will archive the valid section.
Screenshot 2025-03-24 at 7 51 54 PM

### Migration updates I'd like to make 
There are a few changes that might require a migration (if we feel significant enough), including the following:
- Archive column on create should be NULL rather than the zero timestamp it is now ('0001-01-01T00:00:00Z'). I have it handling for the zero timestamp case, but I don't think its preferred
- we currently do not have an end_dt on in the table. Should this be added? Was this planning on being calculated? Should I remove it from the UI for now?
- in the enum for section_status, we have it as "Cancelled" rather than "Canceled" (two L's vs one). American english commonly uses one L, but since the migration had 2 L's I updated the frontend to handle for that. Should we update to one L or keep it as 2?
- I am currently waiting on enrollment for section. This is why all sections show 0 enrollment, and why all sections at this moment can be archived. This ticket does handle the case where enrollment > 0 and status is active or scheduled (and does not allow the user to archive the section), but the variables might need to be updated on this page when that happens.

### To test
You can run the following query to add some sections into the DB:
```
INSERT INTO program_sections (id, program_id, facility_id, instructor_name, start_dt, status, capacity, archived_at) VALUES
(21, 2, 1, 'Linus Torvalds', '2021-07-01', 'Cancelled', 15, NULL),
(22, 2, 1, 'Alex Trebek', '2021-08-01', 'Active', 14, NULL),
(23, 2, 1, 'Mark Smith', '2021-09-01', 'Paused', 16, NULL),
(24, 2, 2, 'Ada Lovelace', '2021-10-01', 'Scheduled', 20, NULL),
(25, 2, 3, 'Grace Hopper', '2021-11-01', 'Completed', 25, '2021-12-01')
```
You then have to check your DB to see what program has the ID of 2. In my experience it has been AA/NA, but it might vary. Just check the ids on the `program_sections` table.

### Misc
I would like to remove the `delete` api for sections as we have established sections will not be deleted, and instead are archived, however, this is difficult when running the sections handler test, since it creates programs, updates them, and then cleans it up by deleting them. Would we like to keep that endpoint for solely this reason, or is there an alternative solution?

### DIFF
--- a/backend/src/handlers/sections_handler.go
+++ b/backend/src/handlers/sections_handler.go
@@ -24,7 +24,7 @@ func (srv *Server) handleGetSectionsForProgram(w http.ResponseWriter, r *http.Re
 		return newInvalidIdServiceError(err, "program ID")
 	}
 	args := srv.getQueryContext(r)
-	sections, err := srv.Db.GetSectionsForProgram(id, &args)
+	sections, err := srv.Db.GetProgramSectionDetailsByID(id, &args)
 	if err != nil {
 		log.add("program_id", id)
 		return newDatabaseServiceError(err)
--- a/backend/src/handlers/sections_handler.go
+++ b/backend/src/handlers/sections_handler.go
@@ -14,8 +14,7 @@ func (srv *Server) registerSectionsRoutes() []routeDef {
 		{"GET /api/program-sections/{section_id}", srv.handleGetSection, false, axx},
 		{"GET /api/program-sections", srv.handleIndexSectionsForFacility, false, axx},
 		{"POST /api/programs/{id}/sections", srv.handleCreateSection, true, axx},
-		{"PATCH /api/program-sections/{id}", srv.handleUpdateSection, true, axx},
-		{"DELETE /api/program-sections/{section_id}", srv.handleDeleteSection, true, axx},
+		{"PATCH /api/program-sections", srv.handleUpdateSection, true, axx},
 	}
 }
 
@@ -79,30 +78,20 @@ func (srv *Server) handleCreateSection(w http.ResponseWriter, r *http.Request, l
 	return writeJsonResponse(w, http.StatusCreated, newSection)
 }
 
-func (srv *Server) handleDeleteSection(w http.ResponseWriter, r *http.Request, log sLog) error {
-	id, err := strconv.Atoi(r.PathValue("section_id"))
-	if err != nil {
-		return newInvalidIdServiceError(err, "section ID")
-	}
-	err = srv.Db.DeleteProgramSection(id)
-	if err != nil {
-		log.add("section_id", id)
-		return newDatabaseServiceError(err)
-	}
-	return writeJsonResponse(w, http.StatusNoContent, "Section deleted successfully")
-}
-
 func (srv *Server) handleUpdateSection(w http.ResponseWriter, r *http.Request, log sLog) error {
-	id, err := strconv.Atoi(r.PathValue("id"))
-	if err != nil {
-		return newInvalidIdServiceError(err, "section ID")
+	ids := r.URL.Query()["id"]
+	sectionIDs := make([]int, 0, len(ids))
+	for _, id := range ids {
+		if sectionID, err := strconv.Atoi(id); err == nil {
+			sectionIDs = append(sectionIDs, sectionID)
+		}
 	}
 	defer r.Body.Close()
 	section := models.ProgramSection{}
 	if err := json.NewDecoder(r.Body).Decode(&section); err != nil {
 		return newJSONReqBodyServiceError(err)
 	}
-	updated, err := srv.Db.UpdateProgramSection(&section, id)
+	updated, err := srv.Db.UpdateProgramSection(&section, sectionIDs)
 	if err != nil {
 		return newDatabaseServiceError(err)
 	}
--- a/frontend/src/Components/inputs/CancelSubmitRow.tsx
+++ b/frontend/src/Components/inputs/CancelSubmitRow.tsx
@@ -6,11 +6,13 @@ import { DeleteButton } from './DeleteButton';
 export function CancelSubmitRow({
     type,
     onCancel,
-    onSubmit
+    onSubmit,
+    action
 }: {
     type: TextModalType;
     onCancel: () => void;
     onSubmit: () => void;
+    action?: string;
 }) {
     return (
         <form method="dialog" className="flex flex-row justify-between">
@@ -18,7 +20,10 @@ export function CancelSubmitRow({
             {type === TextModalType.Delete ? (
                 <DeleteButton onClick={onSubmit} />
             ) : type === TextModalType.Confirm ? (
-                <ConfirmButton onClick={onSubmit} />
+                <ConfirmButton
+                    onClick={onSubmit}
+                    action={action ?? 'Confirm'}
+                />
             ) : null}
         </form>
     );
--- a/frontend/src/Pages/ProgramOverviewDashboard.tsx
+++ b/frontend/src/Pages/ProgramOverviewDashboard.tsx
@@ -1,16 +1,29 @@
 import { useParams } from 'react-router-dom';
-import { useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import StatsCard from '@/Components/StatsCard';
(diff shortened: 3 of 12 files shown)


calisio:
After talking to Preston, the following changes will be made:
- archive_at is just changed to a pointer so that way it will be null on create (no migration update needed)
- end_dt is being removed from the UI 
- cancelled is spelled with two Ls everywhere else, so we will leave it as is for now
- will remove all traces of delete api including those in the handler

CK-7vn (commented):
I noticed that, if i flip a section to active, it still allows me to archive it. If i then opt to display archived sections, it reappears, and although it's grayed out, still shows "active" maybe there would be some confusion on whether or not something is archived or active in that state? I don't know if making sure to switch the status icon if it is archived, and explicitly asking the user if they're sure they want to archive an active section would or should be an option. 
image

calisio:
@CK-7vn Yes, so actually you are allowed to archive a section if there are no students enrolled. Because I am waiting on Cory's PR on enrollment, enrollment right now is set to zero. Therefore, the section CAN be archived, because although it is active, it has zero enrollment, so the action is allowed. Does this answer your question/solve the bug? I might be missing something, so if I am, let me know!

CK-7vn:
> @CK-7vn Yes, so actually you are allowed to archive a section if there are no students enrolled. Because I am waiting on Cory's PR on enrollment, enrollment right now is set to zero. Therefore, the section CAN be archived, because although it is active, it has zero enrollment, so the action is allowed. Does this answer your question/solve the bug? I might be missing something, so if I am, let me know!

Ahh, alright, alright, yes that does help, for sure. I would assume that once archived there's no going back in the sense of unarchiving? cool, that's good news then! Thank you for the clarification!

backend/src/handlers/sections_handler.go
```diff
@@ -83,16 +83,19 @@ func (srv *Server) handleDeleteSection(w http.ResponseWriter, r *http.Request, l
...
 	defer r.Body.Close()
 	section := models.ProgramSection{}
 	if err := json.NewDecoder(r.Body).Decode(&section); err != nil {
 		return newJSONReqBodyServiceError(err)
 	}
-	updated, err := srv.Db.UpdateProgramSection(&section, id)
+	updated, err := srv.Db.UpdateProgramSection(&section, sectionIDs)
 	if err != nil {
 		return newDatabaseServiceError(err)
 	}
```
jtucholski, line 101: Confirming that this is intended to be a batch endpoint, right? Like we supply a single section object in the body with a list of IDs and it's intended to update that specific field for all records with that ID?

frontend/src/Components/inputs/CancelSubmitRow.tsx
```diff
@@ -6,19 +6,25 @@ import { DeleteButton } from './DeleteButton';
...
 }: {
     type: TextModalType;
     onCancel: () => void;
     onSubmit: () => void;
+    action?: string;
 }) {
+    if (type === TextModalType.Confirm && action === undefined) {
+        console.error('must include action type for modal');
+        return;
+    }
```
jtucholski, line 20: Did we come across needing this? Is there a different way that we should enforce this rather than at runtime?

jtucholski: Should we default the value to Confirm if it isn't provided?

backend/src/handlers/sections_handler.go
calisio: The sections that are being updated are passed in through the query with the ids. What is being passed in through the body is the new value of the column we would like to update. So, for example, for archiving, we can pass in multiple id's, and the value in the body is "archived_at" with the current time, which is what will be updated in the table, and it will update all the rows with those ids with that value of archived_at. We could also just pass in one id, and several values in the body that we would want to update (ie pass in ?id=1 and in the body instructor="Josh" and status="Active"). So, I think the answer to your question is yes.

frontend/src/Components/inputs/CancelSubmitRow.tsx
calisio: Oh, yes that would probably be smart! I can default it to 'Confirm'

backend/src/handlers/sections_handler.go
jtucholski: Awesome!

frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -63,33 +31,82 @@ export default function ProgramOverview() {
...
+    });
+
+    useEffect(() => {
+        if (sections === undefined) return;
+        const unableToArchive = sections.filter(
+            (section) =>
+                selectedSections.includes(section.id) &&
+                (section.status === SelectedSectionStatus.Active ||
+                    section.status === SelectedSectionStatus.Scheduled) &&
+                section.enrolled_residents > 1
```
jtucholski, line 71: Should this be >= 1? You're looking to make sure there is at least one resident enrolled, right?

jtucholski (commented):
Only have the one comment I included about the conditional check. Looks great!

jtucholski (approved):

frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -63,33 +31,82 @@ export default function ProgramOverview() {
...
+        last_page: Math.ceil(sections.length / perPage)
     };
 
-    if (programError) {
-        <Error />;
-    }
+    const nonArchivedSections = sections.filter(
+        (section) =>
+            section.archived_at === null ||
+            section.archived_at === '0001-01-01T00:00:00Z'
```
PThorpe92, line 98: did we end up changing the type to `*time.Time` so we don't have to deal with these? I'm not saying this should be taken out either way. 

I think we should add a helper somewhere in the frontend that takes a time string and returns either the formatted Time, or null for the case of ` "" | '0001-01-01T00:00:00Z`

calisio: I am changing it today, was just working on other bugs for Welcome Tour first (different PR)

calisio: > After talking to Preston, the following changes will be made:
> 
> * archive_at is just changed to a pointer so that way it will be null on create (no migration update needed)
> * end_dt is being removed from the UI
> * cancelled is spelled with two Ls everywhere else, so we will leave it as is for now
> * will remove all traces of delete api including those in the handler


Implementing what I commented above today

calisio:
@PThorpe92 should have the updated APIs now & other changes fixed

calisio:
Before this gets merged, @jtucholski I had a question. It seems like we are only showing the sections for the selected facility (ie we are only showing the sections for Potosi Correctional, and then if we switch to MVCF, it shows only the sections for MVCF). That being said, do we think it is worth it to display the facility name in EACH row? It seems a bit redundant to me.

jtucholski:
IMO, I don't think it's worth it, but in this case we designed it similar to the Figma. I think we can move forward with it for now. Lisa is considering some options as to what we display on the Program Overview page when viewed as a Dept Admin and a Facility Admin.

calisio:
Sounds good, will leave it as is for now

jtucholski:
@PThorpe92 Can you give a second review of the changes?

frontend/src/Pages/ProgramOverviewDashboard.tsx
PThorpe92: :man_facepalming: my fault i missed that

PThorpe92 (approved):

PR #811, opened 2025-03-25
feat: add doc number to profile overview and search

CK-7vn:
## Description of the change

This pull request: 

- 	Adds the ability for an admin to add the users DOC number during creation
- 	Adds the ability for an admin to search for the user with their DOC number
- 	Adds the residents DOC number to the profile overview page

- **Related issues**: Closes id 38 

## Screenshot(s)

image
image
image

## Asana 
https://app.asana.com/0/1209460078641109/1209485070839620/f

### DIFF
--- a/backend/src/models/users.go
+++ b/backend/src/models/users.go
@@ -39,6 +39,7 @@ type User struct {
 	Role       UserRole `gorm:"size:64;default:student" json:"role" validate:"oneof=student system_admin facility_admin department_admin"`
 	KratosID   string   `gorm:"size:255" json:"kratos_id"`
 	FacilityID uint     `json:"facility_id"`
+	DOC_ID     string   `json:"doc_id" gorm:"column:doc_id;size:25"`
 
 	/* foreign keys */
 	Mappings             []ProviderUserMapping `json:"mappings,omitempty" gorm:"foreignKey:UserID;constraint:OnDelete CASCADE"`
--- a/backend/src/models/users.go
+++ b/backend/src/models/users.go
@@ -39,7 +39,7 @@ type User struct {
 	Role       UserRole `gorm:"size:64;default:student" json:"role" validate:"oneof=student system_admin facility_admin department_admin"`
 	KratosID   string   `gorm:"size:255" json:"kratos_id"`
 	FacilityID uint     `json:"facility_id"`
-	DOC_ID     string   `json:"doc_id" gorm:"column:doc_id;size:25"`
+	DocID      string   `json:"doc_id" gorm:"column:doc_id;size:25"`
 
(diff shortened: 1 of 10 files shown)


PThorpe92 (changes requested):
Can we also add the doc id to the Claims field? prob eventually will want it in the Kratos id schema too but for now the claims and the user struct will do fine. We just want to make sure it's returned by the `api/auth` endpoint

calisio:
Is DOC number optional? If so, I think maybe we should include that in the label, so  it would be "Department of Corrections ID (optional)". Similar to how we have email on admin form
Screenshot 2025-03-28 at 3 17 16 PM

CK-7vn:
I mean, I would assume it's optional, but I would also argue going forward DOC # is definitely something admin is going to want. 110%, but that's not my call.

calisio:
Right now its optional in the frontend form (you put required:false), so I would probably write "(optional)" next to it just so they know they don't NEED to fill it out on the form to be able to create the resident.

jtucholski:
If we start making it required we would need to have a migration process for existing users who have it in the beta.

calisio:
So I guess what I am trying to say is that if it is not optional, then we should put required:true and implement that in the backend. If it is optional, then lets specify that in the frontend

jtucholski:
Let's make it required front end and back end.

backend/src/models/users.go
```diff
@@ -39,6 +39,7 @@ type User struct {
 	Role       UserRole `gorm:"size:64;default:student" json:"role" validate:"oneof=student system_admin facility_admin department_admin"`
 	KratosID   string   `gorm:"size:255" json:"kratos_id"`
 	FacilityID uint     `json:"facility_id"`
+	DOC_ID     string   `json:"doc_id" gorm:"column:doc_id;size:25"`
```
jtucholski, line 42: This is really nitpicky. The underscore convention stuck out to me because I haven't seen it used in our code base much yet. I looked for other examples for where we have consecutive acronyms and we have FAQID and OidcID. 

I think the Go convention would be DocID, right?

jtucholski (changes requested):
I removed the migration file because somehow that column was added to a prior pull request that had already been merged. 

Left one more comment in there about the usage of DOCID in Go and feeling like it was inconsistent with a naming convention. But we are also inconsistent with this type of naming elsewhere in the codebase.

backend/src/models/users.go
CK-7vn: No, that's totally right...no idea why I did that. Must've been stuck on Rust. My bad.

PThorpe92 (approved):

PR #813, opened 2025-03-27
feat: add functionality so facility admin can add events to the class calendar per ticket 93

carddev81:
## Description of the change

Added a Scheduling section to the Create Class form for allowing users to schedule events for a class when they create it. Made most of my changes to the UI/frontend (javascript, HTML, CSS). Also added a new react component 'react-big-calendar'. 

Reviewer Note: Due to react libraries I suggest executing a `make reset`

- **Related issues**: Link to Asana task: https://app.asana.com/0/1209460078641109/1209755095767545/f

## Screenshot(s)
ScheduleForm
image


frontend/src/Components/inputs/RRuleControl.tsx
```diff
@@ -0,0 +1,264 @@
...
+                const endTotalMin = timeToMinutes(endTime);
+                if (endTotalMin <= startTotalMin) {
+                    setTimeErrors((prev) => ({
+                        ...prev,
+                        endTime: 'End time must be after start time'
+                    }));
+                    isValid = false;
+                }
+            }
+            return isValid;
```
jtucholski, line 129: One idea would be if you're going to make multiple calls to setTimeErrors you could do this at the end after you've determined the errors. Create an errors object at the start and then set each of the values throughout the function before calling setTimeErrors at the end.

jtucholski:
Looking good! While running through the UI I noticed a few things:
1. **Back** as the button text in the Calendar view might make the user think they're going back to the previous screen. Consider using either Previous or using `<` and `>` as the button text with tooltips of **Previous Day**, **Previous Week**, or **Previous Month**, etc.. 
2. I found it odd that the calendar defaulted to Monday even though I didn't have a day set for my recurrence rule. Should we include that in the error validation check like start time and end time is?
3. 5:00 PM to 6:00 PM on the rule shows up in the calendar as 5:00 PM to 6:01 PM. Any idea what might be causing that?
4. 5:00 PM to 5:59 PM on the rule shows up in the calendar as 5:00 PM to 5:00 PM. 
5. If a start date isn't selected or the class details aren't filled in the View Calendar button won't open up, but the user didn't see any error message indicating that's what's happening. 
6. Should we make View Calendar disabled until the user enters the required fields for it to work?

jtucholski:
Pasted image 20250328090023
Pasted image 20250328090010

jtucholski:
Also noticed a few of the styles for the dark mode of it felt off. Maybe we use the teal color for the off range values if that's possible? Not a deal breaker and we can make it a separate enhancement if more research is necessary. 

image

I think the reason the Next button is still highlighted is because I just clicked it and it's in that state until I give focus to another control.

frontend/src/Components/inputs/RRuleControl.tsx
carddev81: Made change.

carddev81:
@jtucholski 'Maybe we use the teal color for the off range values' - are you talking about the greyed out days?

Also, yes, I tested the Next button color and it was that color because it had focus last.

carddev81:
@jtucholski fixed and pushed these changes for fixing the following.
5:00 PM to 6:00 PM on the rule shows up in the calendar as 5:00 PM to 6:01 PM. Any idea what might be causing that?
5:00 PM to 5:59 PM on the rule shows up in the calendar as 5:00 PM to 5:00 PM.

carddev81:
@jtucholski fixed and pushed changes for fixing the following:

1. I found it odd that the calendar defaulted to Monday even though I didn't have a day set for my recurrence rule. Should we include that in the error validation check like start time and end time is? (FIXED)--added error message
2. If a start date isn't selected or the class details aren't filled in the View Calendar button won't open up, but the user didn't see any error message indicating that's what's happening. (FIXED)--made sure there are error messages for missing fields
3. Should we make View Calendar disabled until the user enters the required fields for it to work? (FIXED)--disabled button

carddev81:
@jtucholski fixed and pushed changes for the following

1. Also noticed a few of the styles for the dark mode of it felt off. Maybe we use the teal color for the off range values if that's possible?---made the off days teal.

frontend/src/Components/inputs/RRuleControl.tsx
jtucholski: Are you able to get the errors to trigger anymore? I can't get them to show up.

jtucholski:
@carddev81 I noticed that the validation for start/end time and weekdays wasn't triggering anymore. Commented on the original thread but that might not be visible.

I also pushed a dark mode stylesheet change for support for the current day on rbc calendar. 

@calisio Can you review `app.css` and make sure we are applying the styles correctly as well as override any decisions we made if its inconsistent?

frontend/src/Components/inputs/RRuleControl.tsx
carddev81: They do show up on my screen when pressing the submit button? That's weird they don't show up for you? Maybe we can do a huddle?

frontend/src/Components/EventCalendar.tsx
```diff
@@ -0,0 +1,69 @@
...
+        const generated = occurrences.map((occurrence) => ({
+            title,
+            start: occurrence,
+            end: new Date(occurrence.getTime() + durationMs)
+        }));
+        setEvents(generated);
+    }, [recurrenceRule, durationStr, title]);
+
+    return (
+        <div className="p-4 bg-base-100 rounded-lg shadow">
```
calisio, line 59: remove `bg-base-100 rounded-lg shadow" (it is not in our tailwind colors, and i dont think you really need any of the other classes)

frontend/src/css/app.css
```diff
@@ -166,3 +166,41 @@
 .remove-input-txt-border input[type='text']:focus {
     box-shadow: none;
 }
+
+.rbc-calendar {
+    @apply bg-base-100 text-body-text font-lexend;
```
calisio, line 171: i would personally remove `bg-base-100`

frontend/src/Components/EventCalendar.tsx
carddev81: removed it.

frontend/src/css/app.css
carddev81: removed it

carddev81:
@jtucholski made all changes. Didn't realize you made a change, was unable to push my changes at first.

jtucholski (approved):

frontend/src/Components/inputs/RRuleControl.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+        return (
+            <div className="space-y-6">
+                <DateInput
+                    label="Start Date"
+                    register={register}
+                    interfaceRef={startDateRef}
+                    required
+                    errors={errors}
+                />
```
calisio, line 178: I noticed that the date and time inputs on dark mode have black icons. is there a way to update this?
Screenshot 2025-03-31 at 1 27 42 PM
If its too complicated I can also look into it later. Don't want that to hold this PR up

calisio:
Screenshot 2025-03-31 at 1 47 11 PM

What do you think of this? I feel like it makes a bit more sense to have the current days of the month lighter and the off days darker. Here is the corresponding code:

```
html[data-theme='dark'] .rbc-day-bg {
    background-color: var(--teal-1);
}
html[data-theme='dark'] .rbc-off-range-bg {
    background-color: var(--inner-background);
}
html[data-theme='dark'] .rbc-today {
    background-color: var(--teal-3);
    
}
```

calisio (approved):
looks good from frontend

frontend/src/Components/inputs/RRuleControl.tsx
carddev81: i'll check into it.

carddev81:
@jtucholski and @calisio made the requested styling changes.

PR #814, opened 2025-03-27
fix: update bug fixes on tour

calisio:
## Description of the change
Fixes a few bugs related to welcome tour, including:
- users click on "knowledge center" and then click back when prompted with "Welcome to the Knowledge Center. Here, you can find a range of resources in both written and video format, as well as helpful links.". It now allows you to click on knowledge center again.
- Click on the specific library and see the pop-up: "Here, you can see everything inside the library." Click the Back button on this pop-up. should take you back to be able to click into the library.
- clicking back from the "top content" pop up takes you back to library viewer sub page favorite pop up.
- Adds steps to the tour so the user can tell how far along they are
- changes the code to refer to the step by its target id, rather than by the step index for easy updating in the future if we wanted to add steps.
- **Related issues**: closes asana task [BUG: Guided Tour Navigation Issues in UnlockEd](https://app.asana.com/0/1209460078641109/1209714374559486/f)
## Screenshot(s)
Added steps:
Screenshot 2025-03-27 at 12 31 53 PM

### DIFF
--- a/frontend/src/Components/UnlockEdTour.tsx
+++ b/frontend/src/Components/UnlockEdTour.tsx
@@ -1,6 +1,6 @@
 import { ThemeContext } from '@/Context/ThemeContext';
-import { useTourContext } from '@/Context/TourContext';
-import { useContext, useEffect } from 'react';
+import { initialTourState, useTourContext } from '@/Context/TourContext';
+import { useContext } from 'react';
 import Joyride, { CallBackProps, EVENTS } from 'react-joyride';
 import { useNavigate } from 'react-router-dom';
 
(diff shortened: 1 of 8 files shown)


frontend/src/Components/UnlockEdTour.tsx
```diff
@@ -161,23 +190,66 @@ export default function UnlockEdTour() {
...
+        const { action, index, type, step } = data;
+        const currentTarget = step.target;
         if (
             action === 'close' ||
             action === 'skip' ||
             type === EVENTS.TOUR_END
         ) {
-            setTourState({ tourActive: false, run: false, stepIndex: 0 });
+            setTourState({
+                tourActive: false,
```
PThorpe92, line 201: this is repeated quite a few times. can we save a `defaultTourState` const somewhere and use that?

calisio: @PThorpe92 I only found here and below, is there anywhere else I should change? I already had an `initialTourState`.. not sure why I didn't use it 🤦‍♀️ I updated it in those two places but let me know if there is anywhere else I missed

calisio: Fixing something quickly that changing that broke

carddev81 (approved):
I've approved this but am not 100% sure that the functionality is correct here: when clicking 'Back' on step 13 of 15 shouldn't it take you back to select Home? It takes you back to Step 11 'Favoriting' a single page?

calisio:
Ya, I had it skip back to favoriting a single page because if they wanted to keep going backwards they wouldn't be able to if I put them on select home. Not sure what other people think it should skip to @jtucholski ? I guess its a bit inconsistent, since in other places I have it go to the one that has no other option but to move forward. We should probably choose one or the other

jtucholski:
I'll take a look.

jtucholski:
Why don't the click actions like Step 2 and 12 have a back button?

calisio:
Lisa and I decided to not include those so that it forces the user to interact with the product by doing real actions (rather than just press the next button), but I can add them in if that is what we prefer.

jtucholski:
I think that we should send them back to Step 12 because that's the prior step. 

If possible, we should show the back button when they're going through it in reverse, but I expect that's probably not worth the effort.

calisio:
I don't know if I can handle for the specific case where they are moving through it backwards, since we render all the steps on initialization of the tour (so having it render a new check when they are going backward could be possible but not without some work arounds). What I could do (I think) is ONLY include the back button and NOT include the "next" button on those steps so that way they have to interact to move forward, but have the option to move backwards.

jtucholski:
That recommendation would be better in my opinion.

calisio:
Hmmm, it actually doesn't seem like we can only hide the next button without also hiding the back button. It would require me building out a custom component to use for all the steps with the ability to hide the next button. It seems like you can only hide either 
1. the entire row (as I am doing now)
2. hide ALL of the next buttons (for all of the steps)
Neither of which are what we are looking for.[ Reference here ](https://github.com/gilbarbara/react-joyride/issues/499)

That being said, how do we want to move forward?

jtucholski:
Let's have it go back 2 steps to Step 11 then.

calisio:
@jtucholski Done!

PR #815, opened 2025-03-27
fix: fix history navigation

CK-7vn:
## Description of the change
PR Does: 

- Fixes the tab navigation/highlighting on the knowledge center, so that if you hit back and your last page was in the knowledge center, it navigates accordingly
- Updated pagination so that it is pushed onto the browsers history stack (through the url) so that if you navigate through pages, you are able to go back. 

- **Related issues**: closes ID # 83 in Asana 

## Screenshot(s)
https://github.com/user-attachments/assets/30545fc1-8223-4fa9-9751-042b84086ba5

## Additional context
This is a slightly different paradigm for our pagination, previously we were only adding the pagination information in the url of the api, this pr does add it to the actual visible url, we could have possibly used react routers navigate state, but this is ephemeral and would not persist across full page refreshes so, i thought that although this makes the url a little uglier, it does actually push that url onto our history stack and give us the ability to go back and forth regardless of refreshes. 

## Asana 
https://app.asana.com/0/1209460078641109/1209699141721785/f

### DIFF
--- a/frontend/src/Hooks/paginationUrlSync.tsx
+++ b/frontend/src/Hooks/paginationUrlSync.tsx
@@ -1,0 +1,29 @@
+import { useSearchParams } from 'react-router-dom';
+
+export function useUrlPagination(defaultPage = 1, defaultPerPage = 20) {
+    const [searchParams, setSearchParams] = useSearchParams();
+
+    const page = parseInt(
+        searchParams.get('page') ?? defaultPage.toString(),
+        10
+    );
+    const perPage = parseInt(
+        searchParams.get('per_page') ?? defaultPerPage.toString(),
+        10
+    );
+
+    const setPage = (newPage: number, options?: { replace?: boolean }) => {
+        const params = new URLSearchParams(searchParams);
+        params.set('page', newPage.toString());
+        setSearchParams(params, { replace: options?.replace ?? false });
+    };
+
+    const setPerPage = (newPerPage: number) => {
+        const params = new URLSearchParams(searchParams);
+        params.set('per_page', newPerPage.toString());
+        params.set('page', '1');
+        setSearchParams(params, { replace: false });
+    };
+
+    return { page, perPage, setPage, setPerPage };
+}
--- a/frontend/src/Components/LibraryLayout.tsx
+++ b/frontend/src/Components/LibraryLayout.tsx
@@ -21,6 +21,7 @@ import { useTourContext } from '@/Context/TourContext';
 import { targetToStepIndexMap } from './UnlockEdTour';
 import ToggleView from '@/Components/ToggleView';
 import { useSessionViewType } from '@/Hooks/sessionView';
+import { useUrlPagination } from '@/Hooks/paginationUrlSync';
 
 export default function LibaryLayout({
     studentView
@@ -66,12 +67,18 @@ export default function LibaryLayout({
     if (studentView) {
         role = UserRole.Student;
     }
-    const [perPage, setPerPage] = useState(20);
-    const [pageQuery, setPageQuery] = useState<number>(1);
+
     const route = useLocation();
     const adminWithStudentView = (): boolean => {
         return !route.pathname.includes('management') && isAdministrator(user);
     };
+    const {
+        page: pageQuery,
+        perPage,
+        setPage: setPageQuery,
+        setPerPage
+    } = useUrlPagination(1, 20);
+
     const {
         data: libraries,
         mutate: mutateLibraries,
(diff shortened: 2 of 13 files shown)


CK-7vn:
I need to update the Favorites pagination, forgot that one -_- i switched tickets but will jump back if this gets looked at before its updated just know it'll be the same idea as the other changes.

frontend/src/Hooks/paginationUrlSync.tsx
```diff
@@ -0,0 +1,29 @@
+import { useSearchParams } from 'react-router-dom';
+
```
PThorpe92, line 2: I like this :+1: 

How are we getting around auto `mutate` tho, am i missing something? Most of the places these are used have mutate hooks that fetch `SWR` when changed, or there is a useEffect that calls mutate with these in the dependency array.

CK-7vn: Dang dude, i commented back to this last night, and then like 4 comments appeared so i deleted the unnecessary ones, and it deleted all of them..what the heck. Anyways, so, i thought, or my thinking, was that whenever the key/url changes, it will call the mutate function so because we have the dynamic params inside of the url, when they get updated it essentially calls mutate. At least that's what was going on in my head...definitely could be wrong, but it was working really well, so i at least assumed i was on the right track.

carddev81:
@CK-7vn I'm currently reviewing the code, but wanted to get this to you right away so that you check into it. I noticed that when Navigating from 'Home' to the knowledge center and then hitting the Back button it doesn't take you home.  See Video below.

[BackButton.webm](https://github.com/user-attachments/assets/b4358a80-7c82-4bcb-9ab1-e092a77e6bfd)

CK-7vn:
> @CK-7vn I'm currently reviewing the code, but wanted to get this to you right away so that you check into it. I noticed that when Navigating from 'Home' to the knowledge center and then hitting the Back button it doesn't take you home. See Video below.
> 
>  BackButton.webm

Ya I see it, there's a race condition in there with the url changing, so it's appending libraries onto the /libraries, and then its appending the pagination parameters onto the URl, I'll figure it out, good catch.

carddev81:
@CK-7vn Another issue found during testing, is pagination number is not changing when selecting pagination?  I checked Helpfulinks pagination to compare and when changing it there, the number changes. See video.

[PaginationNumber.webm](https://github.com/user-attachments/assets/f8c8b93c-f303-43a6-a869-cbc0c7315fe1)

carddev81 (changes requested):
Finished review of code, just make the changes to add missing code to the LibraryLayout and for fixing Home back button

frontend/src/Hooks/paginationUrlSync.tsx
PThorpe92: oh yeah duh if they are templated into the string it should automatically refetch :man_facepalming: 

it was just confusing because those params were taken out of useEffects + wrapper functions that manually called `mutate`.

carry on. :smile:

CK-7vn:
@carddev81 after a really long hair pulling day.....i do believe that the issue with the /libraries is fixed, it's at least working for me, also it's not appending to the url until the pagination is changed, idk if thats what you were talking about in your comment above.

carddev81 (changes requested):
@CK-7vn wasn't able to reproduce the issue from before, but now the 'tour' blinking logic i being fired off sometimes when hitting back button:

[BlinkingActive.webm](https://github.com/user-attachments/assets/b93d82ad-787f-458b-a9eb-155c37da97a3)

CK-7vn:
> @CK-7vn wasn't able to reproduce the issue from before, but now the 'tour' blinking logic i being fired off sometimes when hitting back button:
> 
>  BlinkingActive.webm

Ya...idk about this one man, I couldn't duplicate it, and i've made 3 different new accounts, and done everything I can...I'm not sure here.

carddev81 (approved):
Looks good. Tested Good.

jtucholski:
I apologize for being so late to this. While the issue I described is working there are a few other test cases that I wonder if you can look at. 

1. Navigating forward and backward in the Knowledge Center using the browser buttons does not change the selected tab.




2. This one might be a stretch...while going backward from the Resident Profile to the Residents page the sort resets to "Name A-Z". While it remembers what page I'm on it doesn't retain the sort. 

3. Are you noticing any odd behavior with the searchQuery? It feels like when I use it the paging doesn't work properly. 

https://github.com/user-attachments/assets/c4c37960-ea92-4fcd-b4ec-cd7a45df32a4

jtucholski (changes requested):
sorry..i posted a few comments on the main thread that have to do with unexpected behaviors that I'm getting on the resident page.

jtucholski (approved):

PR #816, opened 2025-03-27
feat: add list view to knowledge center components

CK-7vn:
-  ## Description of the change
This PR: 

- Adds a list view to all knowledge center views
- Uses session storage to maintain preferred view type 

- **Related issues**: Closes ID 16 in Asana 

## Screenshot(s)
image

image

image
image
image
image

## Additional context
If someone recommends we change the list view across favorites, or videos, or anything i won't say i necessarily disagree, as I said, this is a list view based off of the same format as our Program list view.  

## Asana 
https://app.asana.com/0/1209460078641109/1209478696323291/f

### DIFF
--- a/frontend/src/Components/FavoriteCard.tsx
+++ b/frontend/src/Components/FavoriteCard.tsx
@@ -1,7 +1,12 @@
 import { useLocation, useNavigate } from 'react-router-dom';
 import { StarIcon as SolidStar } from '@heroicons/react/24/solid';
 import { KeyedMutator } from 'swr';
-import { ToastState, ServerResponseMany, OpenContentItem } from '@/common';
+import {
+    ToastState,
+    ServerResponseMany,
+    OpenContentItem,
+    ViewType
+} from '@/common';
 import API from '@/api/api';
 import { useToast } from '@/Context/ToastCtx';
 
@@ -13,12 +18,14 @@ interface FavoriteCardProps {
     perPage: number;
     mutate: KeyedMutator<ServerResponseMany<OpenContentItem>>;
     isAdminInStudentView: boolean;
+    view: ViewType;
 }
 
 const FavoriteCard: React.FC<FavoriteCardProps> = ({
     favorite,
     mutate,
-    isAdminInStudentView
+    isAdminInStudentView,
+    view
 }) => {
     const navigate = useNavigate();
     const { toaster } = useToast();
@@ -27,7 +34,9 @@ const FavoriteCard: React.FC<FavoriteCardProps> = ({
     const adminWithStudentView = (): boolean => {
         return !route.pathname.includes('management') && isAdministrator(user);
     };
-
+    if (!route.pathname.includes('knowledge-center')) {
+        view = ViewType.Grid;
+    }
     const handleCardClick = () => {
         if (!favorite.visibility_status) {
             return;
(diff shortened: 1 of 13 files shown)


calisio:
Hm, I feel like this is the format of cards just taking up the width of the screen. I think we should reconfigure the card entirely and make a new component for it that has the columns in a row format, similar to `EnrolledCourseCard.tsx`
Screenshot 2025-03-28 at 11 58 32 AM
You might need to prioritize some information over others. For example, if description is too long, it might be hard to include in the row

CK-7vn:
> Hm, I feel like this is the format of cards just taking up the width of the screen. I think we should reconfigure the card entirely and make a new component for it that has the columns in a row format, similar to `EnrolledCourseCard.tsx` Screenshot 2025-03-28 at 11 58 32 AM You might need to prioritize some information over others. For example, if description is too long, it might be hard to include in the row

So are you saying create a new component for all 3 cards?

frontend/src/Components/cards/HelpfulLinkCard.tsx
calisio, line 1: I think something in here switched up the look of helpful link card in the dashboard 
Screenshot 2025-03-31 at 2 02 24 PM

frontend/src/Components/VisibleHiddenToggle.tsx
```diff
@@ -1,18 +1,28 @@
...
     visible: boolean;
     changeVisibility: (visibilityStatus: boolean) => void;
+    view?: ViewType;
 }) {
     const toggleClass =
-        'py-1 rounded-lg inline-flex items-center justify-center gap-2';
+        view === ViewType.List
+            ? 'flex-1 px-4 py-2 rounded-lg flex items-center justify-center gap-2 text-base'
+            : 'flex-1 px-2 py-1 rounded-lg flex items-center justify-center gap-1 text-xs';
+
```
calisio, line 18: move `flex flex-1 rounded-lg items-center justify-center` to before where `toggleClass` is used below (ie 
```
className={`flex flex-1 rounded-lg items-center justify-center ${toggleClass} ...
```
so that this can just be 
```   
const toggleClass =
        view === ViewType.List
            ? 'px-4 py-2 gap-2 text-base'
            : 'px-2 py-1 gap-1 text-xs';
```

calisio: just condenses a bit and less repetition

frontend/src/Components/LibraryCard.tsx
calisio, line 1: I think we need to check if the user is looking at it from the dashboard, and if they are make the viewtype grid rather than list. I think my dashboard is adopting the list from the view within management?
Screenshot 2025-03-31 at 3 33 24 PM

calisio (changes requested):
I think there are a few updates for this PR to go through, specifically those that change the card view on the dashboard!

frontend/src/Components/FavoriteCard.tsx
```diff
@@ -91,42 +98,97 @@ const FavoriteCard: React.FC<FavoriteCardProps> = ({
...
+                </h3>
+                <p className="body-small text-center">
+                    {favorite.content_type === 'video'
+                        ? favorite.channel_title
+                        : favorite.provider_name}
+                </p>
+            </div>
+        );
+    } else {
+        if (view === ViewType.List) {
```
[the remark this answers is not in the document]
CK-7vn: I'll check it out after i'm done renaming this stuff.

calisio (approved):

carddev81 (approved):
Looks good. Tested good.

PR #820, opened 2025-04-01
feat: add functionality so facility admin can edit an existing section per ticket 48

carddev81:
## Description of the change

Added the functionality to modify class section details. This change included some html, javascript, and backend golang coding modifications and additions. 

NOTE: Scheduling section is currently disabled for edit.

- **Related issues**: Link to Asana task: https://app.asana.com/0/1209460078641109/1209485911445029/f

## Screenshot(s)
[example.webm](https://github.com/user-attachments/assets/0da2b80e-9400-48db-82d2-b45a8d751393)

### DIFF
--- a/backend/src/handlers/classes_handler.go
+++ b/backend/src/handlers/classes_handler.go
@@ -14,7 +14,7 @@ func (srv *Server) registerClassesRoutes() []routeDef {
 		{"GET /api/program-classes/{class_id}", srv.handleGetClass, false, axx},
 		{"GET /api/program-classes", srv.handleIndexClassesForFacility, false, axx},
 		{"POST /api/programs/{id}/classes", srv.handleCreateClass, true, axx},
-		{"PATCH /api/program-classes", srv.handleUpdateClass, true, axx},
+		{"PATCH /api/program-classes/{id}", srv.handleUpdateClass, true, axx},
 	}
 }
 
@@ -79,19 +79,41 @@ func (srv *Server) handleCreateClass(w http.ResponseWriter, r *http.Request, log
 }
 
 func (srv *Server) handleUpdateClass(w http.ResponseWriter, r *http.Request, log sLog) error {
-	ids := r.URL.Query()["id"]
-	classIDs := make([]int, 0, len(ids))
-	for _, id := range ids {
-		if classID, err := strconv.Atoi(id); err == nil {
-			classIDs = append(classIDs, classID)
-		}
+	id, err := strconv.Atoi(r.PathValue("id"))
+	if err != nil {
+		return newInvalidIdServiceError(err, "class ID")
 	}
-	defer r.Body.Close()
+	if id == 0 {
+		ids := r.URL.Query()["id"]
+		classIDs := make([]int, 0, len(ids))
+		for _, id := range ids {
+			if classID, err := strconv.Atoi(id); err == nil {
+				classIDs = append(classIDs, classID)
+			}
+		}
+		defer r.Body.Close()
+		classMap := make(map[string]interface{})
+		if err := json.NewDecoder(r.Body).Decode(&classMap); err != nil {
+			return newJSONReqBodyServiceError(err)
+		}
+		err := srv.Db.UpdateProgramClasses(classMap, classIDs)
+		if err != nil {
+			return newDatabaseServiceError(err)
+		}
+		return writeJsonResponse(w, http.StatusOK, "Successfully updated program class")
+	}
 	class := models.ProgramClass{}
 	if err := json.NewDecoder(r.Body).Decode(&class); err != nil {
 		return newJSONReqBodyServiceError(err)
 	}
-	updated, err := srv.Db.UpdateProgramClass(&class, classIDs)
+	enrolled, err := srv.Db.GetTotalEnrollmentsByClassID(id)
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	if enrolled > class.Capacity {
+		return writeJsonResponse(w, http.StatusBadRequest, "Cannot update class until unenrolling residents")
+	}
+	updated, err := srv.Db.UpdateProgramClass(&class, id)
 	if err != nil {
 		return newDatabaseServiceError(err)
 	}
--- a/backend/src/handlers/classes_handler.go
+++ b/backend/src/handlers/classes_handler.go
@@ -14,7 +14,8 @@ func (srv *Server) registerClassesRoutes() []routeDef {
 		{"GET /api/program-classes/{class_id}", srv.handleGetClass, false, axx},
 		{"GET /api/program-classes", srv.handleIndexClassesForFacility, false, axx},
 		{"POST /api/programs/{id}/classes", srv.handleCreateClass, true, axx},
-		{"PATCH /api/program-classes/{id}", srv.handleUpdateClass, true, axx},
+		{"PATCH /api/program-classes", srv.handleUpdateClasses, true, axx},
+		{"PATCH /api/program-class/{id}", srv.handleUpdateClass, true, axx},
 	}
 }
 
@@ -83,25 +84,6 @@ func (srv *Server) handleUpdateClass(w http.ResponseWriter, r *http.Request, log
 	if err != nil {
 		return newInvalidIdServiceError(err, "class ID")
 	}
-	if id == 0 {
-		ids := r.URL.Query()["id"]
-		classIDs := make([]int, 0, len(ids))
-		for _, id := range ids {
-			if classID, err := strconv.Atoi(id); err == nil {
-				classIDs = append(classIDs, classID)
-			}
-		}
-		defer r.Body.Close()
-		classMap := make(map[string]interface{})
-		if err := json.NewDecoder(r.Body).Decode(&classMap); err != nil {
-			return newJSONReqBodyServiceError(err)
-		}
-		err := srv.Db.UpdateProgramClasses(classMap, classIDs)
-		if err != nil {
-			return newDatabaseServiceError(err)
-		}
-		return writeJsonResponse(w, http.StatusOK, "Successfully updated program class")
-	}
 	class := models.ProgramClass{}
 	if err := json.NewDecoder(r.Body).Decode(&class); err != nil {
 		return newJSONReqBodyServiceError(err)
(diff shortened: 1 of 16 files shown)


frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -149,7 +153,7 @@ export default function ProgramOverview() {
             ableToArchiveSections.some((section) => section.id === id)
         );
         const idString = '?id=' + selectedAbleToArchiveSections.join('&id=');
-        const resp = await API.patch(`program-sections${idString}`, {
+        const resp = await API.patch(`program-sections/0${idString}`, {
```
jtucholski, line 156: I can see this `0` easily getting removed from the code on accident.

frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -301,18 +305,38 @@ export default function ProgramOverview() {
...
+                                                iconClassName={
+                                                    '!w-5 !h-5 cursor-pointer'
                                                 }
+                                                onClick={() => {
+                                                    navigate(
+                                                        `/programs/${id}/class/${section.id}`,
+                                                        {
+                                                            state: {
+                                                                title: `Program: ${program?.name}`
+                                                            }
```
jtucholski, line 334: When I first saw this I thought it was a good find but it introduces the expectation that the user follows the expected flow. If the user navigates enter the URL directly in the address bar to the URL in their browser, the Program name isn't included and instead they see "Class Details." Is that expected?

jtucholski (commented):
I left a few comments on the files here in the PR. Can you respond to those? I don't know if changes are needed.

However...we do have PR #821 out there to contend with which renames sections to classes.

backend/src/database/program_sections.go
```diff
@@ -29,17 +31,37 @@ func (db *DB) CreateProgramSection(content *models.ProgramSection) (*models.Prog
...
+	if err := db.Session(&gorm.Session{FullSaveAssociations: true}).Updates(&existing).Error; err != nil {
 		return nil, newUpdateDBError(err, "program sections")
 	}
-	return content, nil
+	return existing, nil
+}
+
+func (db *DB) GetTotalEnrollmentsBySectionID(id int) (int64, error) {
+	var count int64
+	if err := db.Model(&models.ProgramSectionEnrollment{}).Where("section_id = ?", id).Count(&count).Error; err != nil {
```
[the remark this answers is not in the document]
carddev81: yes, this will need to filter by enrollment status.  I will make the change.

backend/src/handlers/sections_handler.go
```diff
@@ -79,19 +79,41 @@ func (srv *Server) handleCreateSection(w http.ResponseWriter, r *http.Request, l
...
+			if sectionID, err := strconv.Atoi(id); err == nil {
+				sectionIDs = append(sectionIDs, sectionID)
+			}
+		}
+		defer r.Body.Close()
+		sectionMap := make(map[string]interface{})
+		if err := json.NewDecoder(r.Body).Decode(&sectionMap); err != nil {
+			return newJSONReqBodyServiceError(err)
+		}
+		err := srv.Db.UpdateProgramSections(sectionMap, sectionIDs)
```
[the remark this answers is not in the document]
carddev81: This is the logic that runs only for archiving sections and will not execute for updating a section values.

frontend/src/Pages/ProgramOverviewDashboard.tsx
carddev81: ha, good find. I wasn't expecting the user to do that.  I can fix this but i'll have to think on it a bit. First thought---brief logic flow here: 

if 'Class Details'
    then change to Program: <fetch_name_from_db_call>

frontend/src/Pages/ProgramOverviewDashboard.tsx
carddev81: Can we talk about this one?

backend/src/database/program_sections.go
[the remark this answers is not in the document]
carddev81: @jtucholski pushed changes for this.

frontend/src/Pages/ProgramOverviewDashboard.tsx
jtucholski: There's nothing wrong with it saying Class Details other than I think Lisa asked for the requirement.  

But it raises the question if we should be relying on navigation state to affect what the user sees on the next page. I think we should try and make each page self-sufficient and not have it rely on the page the user came from.

carddev81:
@jtucholski merge conflicts resolved + changes we talked about have been committed/pushed.

frontend/src/Pages/ProgramOverviewDashboard.tsx
carddev81: @jtucholski made the modifications for dynamically loading the title on page render versus relying on navigate

backend/src/handlers/classes_handler.go
```diff
@@ -14,7 +14,8 @@ func (srv *Server) registerClassesRoutes() []routeDef {
 		{"GET /api/program-classes/{class_id}", srv.handleGetClass, false, axx},
 		{"GET /api/program-classes", srv.handleIndexClassesForFacility, false, axx},
 		{"POST /api/programs/{id}/classes", srv.handleCreateClass, true, axx},
-		{"PATCH /api/program-classes", srv.handleUpdateClass, true, axx},
+		{"PATCH /api/program-classes", srv.handleUpdateClasses, true, axx},
+		{"PATCH /api/program-class/{id}", srv.handleUpdateClass, true, axx},
```
PThorpe92, line 18: i know is going to change anyway soon but i cant help myself :)
`/api/program-classes/{id}`

jtucholski: We'll have to consolidate to a single handler function then because we can't have both `/api/program-classes` and `api/program-classes/{id}`. The reason two paths exist was because one was used to perform a batch update (querystring the id values and switch status to archived) while the other performed a single more comprehensive update.

frontend/src/Layouts/AuthenticatedLayout.tsx
```diff
@@ -20,9 +26,14 @@ export default function AuthenticatedLayout() {
         window.websocket = new WebsocketSession(user);
     }
     const matches = useMatches() as CustomRouteMatch[];
-    const currentMatch = matches.find((match) => match?.handle?.title);
+    // const currentMatch = matches.find((match) => match?.handle?.title);
```
jtucholski, line 29: Can we remove this commented out line?

backend/src/handlers/classes_handler.go
PThorpe92: why not? that is exactly what we would want, it follows the paradigm of `/api/resources` for many and `/api/resources/{id}` for a singular

frontend/src/Components/EventCalendar.tsx
```diff
@@ -43,6 +43,7 @@ export default function EventCalendar({
         }
         const durationMs = parseDurationToMs(durationStr);
         const rule = RRule.fromString(recurrenceRule);
+        //rule.options.byweekday = rule.options.byweekday.map((idx) => idx + 1);
```
calisio, line 46: remove ?

backend/src/handlers/classes_handler.go
jtucholski: Sorry my brain broke too and thought it would result in a failure to find a match between `/` and `/{id}`. I don't know why I had that block.

frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -303,20 +301,38 @@ export default function ProgramOverview() {
...
+                                                iconClassName={
+                                                    '!w-5 !h-5 cursor-pointer'
                                                 }
+                                                onClick={() => {
+                                                    navigate(
+                                                        `/programs/${id}/class/${program_class.id}`,
+                                                        {
+                                                            state: {
+                                                                title: `Program: ${program?.name}`
+                                                            }
```
calisio, line 330: Just checking- should you be able to update a program if it is archived? In the case of section status, you are not allowed to update it anymore. I would assume it is the same for edit? I think you should disable this action so they are not allowed to edit once archived, but unsure if that was a requirement already previously discussed.

calisio:
For some reason, when I edit a class with no end date, and then update it, it sets the end date as Jan 1, 1970?

Screenshot 2025-04-07 at 1 19 58 PM

calisio:
I am also now getting this bug- I added a class and cannot update the section status. This happens both when I have not edited the class, and also post-editing of the class. I have attached a screenshot of my terminal log with it
Screenshot 2025-04-07 at 1 22 00 PM

frontend/src/Components/EventCalendar.tsx
carddev81: removed

frontend/src/Layouts/AuthenticatedLayout.tsx
carddev81: removed

carddev81:
oh man, good catch on the status column not updating...must have missed the name change.

PThorpe92 (approved):
This LGTM as soon as the `status` bug is fixed :+1:

frontend/src/Pages/ClassManagementForm.tsx
```diff
@@ -189,6 +232,9 @@ export default function ClassManagementForm() {
                             errors={errors}
                             register={register}
                             onChange={setRruleIsValid}
+                            recurrenceRule={cls?.events?.[0]?.recurrence_rule}
+                            duration={cls?.events?.[0]?.duration}
+                            disabled={!!class_id}
```
calisio, line 237: if you do implement fetching the class through route loader, you would then actually be able to remove these and check for them directly in the RRuleControl. When you fetch and pass data through routeLoader, you are able to fetch it through its children using the `useLoaderData` hook in the same way you would fetch it in this parent component. This would be beneficial so you dont have to keep passing these down

frontend/src/Components/inputs/RRuleControl.tsx
```diff
@@ -158,6 +171,48 @@ export const RRuleControl = forwardRef<RRuleFormHandle, RRuleControlProp>(
             onChange?.(canCreateRule());
         }, [startTime, endTime, interval, frequency, byWeekDays, endOption]);
 
+        useEffect(() => {
+            if (!recurrenceRule || !duration) return;
+            try {
+                const rule = RRule.fromString(recurrenceRule);
```
calisio, line 177: Okay, I'm a little bit lost on the purpose of this function here. Is this to set all of the default values?

carddev81: Yes, this is used to set the values on the recurrence rule control form. The recurrence rule is parsed and values are used to set the form.

backend/src/handlers/classes_handler.go
carddev81: made change

frontend/src/Pages/ClassManagementForm.tsx
```diff
@@ -51,48 +38,104 @@ export default function ClassManagementForm() {
...
+                setClass(cls);
+                setEditFormValues(cls);
+            } catch (error) {
+                toaster('Error loading class details', ToastState.error);
+                console.error('Fetch error:', error);
+            }
+        }
+
+        void fetchClassDetails();
+    }, [id, class_id, reset]);
```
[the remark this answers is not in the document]
carddev81: Since there was a switch involved on calling fetchClassDetails I left this here. So when class_id is null or undefined the fetch won't happen.  Should I still make the change to add it to a loader?

jtucholski (approved):

PR #825, opened 2025-04-04
feat: update resident profile and add account history

calisio:
## Description of the change

- Updates a few UI things on resident profile page (not specific to this ticket) but just to keep the UI consistent with other pages.
- changes the top libraries/recently viewed videos to one card with a toggle
- adds in account history card with pagination (UI)
- adds the backend for adding account history with the following actions
    - account creation
    - resetting a password by an admin
    - setting a password by the user
    - added to a facility (only on account creation)
- **Related issues**: Closes [Capture & Display User Activity ](https://app.asana.com/0/1209460078641109/1209825280254697/f)

## Screenshot(s)

https://github.com/user-attachments/assets/d880757a-8480-4223-b1d4-afcda8b3af76

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -530,7 +530,7 @@ func (db *DB) GetUserAccountHistory(args *models.QueryContext, userID uint) ([]m
 	if err := tx.Count(&args.Total).Error; err != nil {
 		return nil, newGetRecordsDBError(err, "user_account_history")
 	}
-	if err := tx.Order("uah." + args.OrderClause()).
+	if err := tx.Order("uah.created_at desc").
 		Offset(args.CalcOffset()).
 		Limit(args.PerPage).
 		Scan(&history).Error; err != nil {
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -98,20 +98,18 @@ func (db *DB) TransferResident(ctx *models.QueryContext, userID int, currFacilit
 		trans.Rollback()
 		return newUpdateDBError(err, "users")
 	}
-	/* HISTORY ACTION START */
-	// uintFacID := uint(transFacilityID)
-	// history := models.UserAccountHistory{
-	// 	UserID:     uint(userID),
-	// 	AdminID:    &ctx.UserID,
-	// 	Action:     models.FacilityTransfer,
-	// 	FacilityID: &uintFacID,
-	// 	CreatedAt:  time.Now(),
-	// }
-	// if err := trans.Create(&history).Error; err != nil {
-	// 	trans.Rollback()
-	// 	return newCreateDBError(err, "user_account_history")
-	// }
-	/* HISTORY ACTION END */
+	uintFacID := uint(transFacilityID)
+	history := models.UserAccountHistory{
+		UserID:     uint(userID),
+		AdminID:    &ctx.UserID,
+		Action:     models.FacilityTransfer,
+		FacilityID: &uintFacID,
+		CreatedAt:  time.Now(),
+	}
+	if err := trans.Create(&history).Error; err != nil {
+		trans.Rollback()
+		return newCreateDBError(err, "user_account_history")
+	}
 	if err := trans.Commit().Error; err != nil {
 		return NewDBError(err, "unable to commit DB transaction")
 	}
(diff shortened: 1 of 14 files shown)


carddev81 (commented):
Noticed that pagination button floats with content? Should this be fixed to the bottom of card?



image

backend/src/database/users.go
```diff
@@ -435,3 +435,45 @@ func (db *DB) IncrementUserFAQClick(args *models.QueryContext, question string)
 	}
 	return nil
 }
+
+func (db *DB) InsertUserAccountHistoryAction(userID uint, action models.UserAccountHistoryAction, adminID *uint, facilityID *uint, programsHistoryID *uint) error {
+	history := models.UserAccountHistory{
```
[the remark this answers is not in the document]
calisio: Ya I can definitely do that!

calisio:
Everything should be updated now and integrated with Rich's PR 827. @jtucholski I can deal with the UI update to the dropdown to manage a resident. Do you want that in this PR or in a separate one?

backend/src/handlers/user_handler.go
```diff
@@ -23,6 +23,7 @@ func (srv *Server) registerUserRoutes() []routeDef {
 		{"GET /api/users/resident-verify", srv.handleResidentVerification, true, axx},
 		{"PATCH /api/users/resident-transfer", srv.handleResidentTransfer, true, axx},
 		{"POST /api/users/student-password", srv.handleResetStudentPassword, true, axx},
+		{"GET /api/user-account-history/{id}", srv.handleGetUserAccountHistory, true, axx},
```
jtucholski, line 26: I'm not sure if this is in @PThorpe92's PR, but what do you think about making this `/api/users/{id}/account-history?

frontend/src/Pages/ResidentProfile.tsx
```diff
@@ -114,263 +140,242 @@ const ResidentProfile = () => {
...
+                                </button>
+                                {user && canSwitchFacility(user) && (
+                                    <button
+                                        className="button bg-grey-1 text-teal-4"
+                                        onClick={() =>
+                                            showModal(verifyResidentModal)
+                                        }
+                                    >
+                                        Transfer Resident
+                                    </button>
```
jtucholski, line 206: Can you center these buttons? When the name is longer and stretches the width of the container the buttons are shifted more toward the left.

frontend/src/Pages/ResidentProfile.tsx
```diff
@@ -26,43 +27,81 @@ import { VerifyResidentModal } from '@/Components/modals/VerifyResidentModal';
...
+            <p className="body text-right">{column}</p>
+            <div className="flex">
+                <p className="body">:</p>
+                <p className="body pl-3">{value}</p>
+            </div>
+        </div>
+    );
+}
+
+function OpenContentCardToggle({
```
jtucholski, line 51: Do you think that we can make the tab that isn't active less prominent? Looking at the UI I can't say I'm able to tell which one is active/inactive just by color. Maybe a lighter font-weight?

jtucholski (approved):
I left a few comments for you to decide what to do with. Other than that, looks great to me!

frontend/src/Pages/ResidentProfile.tsx
calisio: Oh yes for sure- that would be way better

backend/src/handlers/user_handler.go
calisio: I agree with this

backend/src/database/users.go
```diff
@@ -518,3 +504,37 @@ func (db *DB) IncrementUserFAQClick(args *models.QueryContext, question string)
...
+				psh.*`).
+		Joins("INNER JOIN users ON uah.user_id = users.id").
+		Joins("LEFT JOIN users admins ON uah.admin_id = admins.id").
+		Joins("LEFT JOIN facilities ON uah.facility_id = facilities.id").
+		Joins("LEFT JOIN program_classes_history psh ON uah.program_classes_history_id = psh.id").
+		Where("uah.user_id = ?", userID)
+	if err := tx.Count(&args.Total).Error; err != nil {
+		return nil, newGetRecordsDBError(err, "user_account_history")
+	}
+	if err := tx.Order("uah." + args.OrderClause()).
```
PThorpe92, line 533: do we know that the referenced column will always be from user acct history? I see below that we aren't doing any ordering in the query yet, so better to hard-code `uah.created_at asc` rather than relying on args.OrderClause() which could change dynamically and reference a column from a diff table

PThorpe92 (approved):

PR #831, opened 2025-04-09
feat: add attendance tracking to program class events

CK-7vn:
## Description of the change
This Pull Request Does: 

- Adds a full page Attendance page that displays a table of date keyed events for a class
- Adds the ability to mark attendance for each event
- Adds the ability to edit attendance for each event 

- **Related issues**: Closes Asana ID #id_103

## Screenshot(s)
image
image
image
image
image
image

## Additional context
Because of the way that ProgramClassEvent's and ProgramClassEventAttendance works, we had to parse the rrule, and then create a combined model to return to the frontend. For the Class Events page we needed to first get the ProgramClassEvent, this gives us the Duration, RRule, Attendees,and Overrides, we grab the program class event, and parse the rrule, we need the date to be returned in a 08:30-9:30 AM type format, after that, we need to query and get the ProgramClassEventAttendance records for each occurrence of that event per the dates that we parsed using the RRule. We need the attendance records, but we can't get the attendance records without parsing the rrule, so we do both of those steps and then return a newly created struct of ClassEventInstance, which is simply the formatted class time, event_id, date, and the AttendanceRecords. Because of how we combine these we paginate in a slightly different way than usual. The dynamic occurrence generation kind of through me for a loop. I probably could have done a sql query for event attendance...but honestly..skill issue, sorry guys'

---
  - https://app.asana.com/0/0/1209360181871023


calisio (dismissed):
Making a broader comment with some of the things I find:
1. Saving attendance does not do anything for me at the moment

https://github.com/user-attachments/assets/0c26ff1a-e202-4c64-a073-bb576d98d4d2

2. there is something being printed in the terminal that takes up the entire screen- could we remove this? (see right side of terminal) 
Screenshot 2025-04-10 at 11 42 28 AM

3. Should we be marking any attendance by default? And if we decide to, should it be absent - unexcused? I don't think we should mark any by default, but if we decide to have one as default, I kinda think it should be "present" rather than absent? (Kinda assumes the best in people rather than the worst?) but 100% open to push back on that

4. should mark all present be actually marking EVERYONE present, or should it be that we mark the selected individuals present? I am thinking that if we have no one selected, it marks EVERYONE as present, but if we have some selected, it only marks those rows as present.
https://github.com/user-attachments/assets/833951c7-7143-4ac0-ae5c-961daf987c82

5. Honestly, not 100% sure how to reproduce this bug, but i think if you search and then try to save attendance, it doesnt allow you to save the attendance anymore?

https://github.com/user-attachments/assets/0f545a88-e279-45b3-a898-a024f2d2df36

(commenting this so i dont lose the changes, but might update with more as I explore)

calisio:
https://github.com/user-attachments/assets/ed495ce4-dd42-45b5-a195-8a6760252bc3

This may be an enhancement because I'm not quite sure the best solution to this, but this takes a _really_ long time to load. Even when switching pages. We should probably preload the events using route loader? I know this is kinda an exception because we usually use useSWR when we have something paginated like this, but in this case I feel like it really takes away from user experience to have to wait for it to load for so long.
Again, something that probably won't be fixed with this PR but definitely should be a top priority in the backlog @jtucholski

CK-7vn:
1. You need to click on the user to actually save attendance, that way noone accidentally changes attendance without intending to, that is my assumption as to why it wasn't saving in your first video. 
2. Ya, i must have missed that print statement, that is my bad. 
3. No push back on this, I think it should be nothing over something specific necessarily. 
4. Personally I think that Mark All Present should mark all present, I feel as though if somebody has to manually go in and click each individual user, then its not much more to just click present open to more discussion on it for sure. 
5. I think that's the same issue as the first one, you actually have to manually click the row, or else it won't save. 
@calisio

calisio:
I think #5 is a different bug. When i search, the button just completely stops working. I am unable to save any of the changes. See the second half of the video (the beginning of the video is the same, just to demonstrate that the button works initially, but then stops working).

calisio:
Im not sure I am following the "click on the user to save attendance", and so I would assume the admin wouldn't follow either :) . I think if we are going to implement it that way (where I think you are implying that the user HAS to do an action on that user in order for it to save?), then we need a visual element to show that that user's attendance won't be saved if they didn't alter it. If not, we leave the user confused on why they pressed save and nothing happened

But, I still think there is something buggy with the saving of attendance. I have attached a video below where I do select the attendance, but it still does not update.
https://github.com/user-attachments/assets/a0828e54-eacf-4b57-b6fe-72440b61e78e

CK-7vn:
@calisio Yah the checkbox needs to be checked or else it doesn't send the request

CK-7vn:
@calisio i'm going to switch this over to pure sql and not use a map, see if that speeds it up at all.

CK-7vn:
@calisio After switching that method to SQL, much better performance, switched it over so that the button now gray's out, and things seems to be working better. Also added search to the backend now that we're using SQL. Must be out of my brain fart. DK why i was having such a hard time doing the SQL the last couple days...smh..

image

calisio:
Definitely way quicker! I like that no option is selected on load. It makes it way more intuitive to use. 

There were a couple things I did find while using it again with updates:

1. If I mark some attendance, and then search, it clears all of the attendance I have marked.

https://github.com/user-attachments/assets/002fbb01-5f9e-4bf3-8d5b-16c1782c083c


2. Updating the note is delayed. When I edit a note, and then I go back to edit it again, it does not update correctly. It seems to be delayed a cycle? (sorry for the long video)

https://github.com/user-attachments/assets/cdba0290-920c-47e1-99d5-4e71e7884b78

CK-7vn:
@calisio 
Pushed another update should be good now.

backend/src/database/program_classes.go
```diff
@@ -11,6 +11,12 @@ func (db *DB) GetClassByID(id int) (*models.ProgramClass, error) {
 	if err := db.Preload("Events").Preload("Enrollments").First(content, "id = ?", id).Error; err != nil {
 		return nil, newNotFoundDBError(err, "program classes")
 	}
+	enrollment, capacity, err := db.GetProgramClassEnrollmentInfo(id)
```
carddev81, line 14: I don't think this database call needs to return capacity because it is already being retrieved in the query call on line 11? If you remove this you'll have to remove line 18.

Also, what needs to change is the query that is being executed from GetProgramClassEnrollmentInfo. It is backwards. The query needs to pull from program_classes and do a left outer join on program_class_enrollments, if this isn't fixed then when there are no program_class_enrollments then the query returns nothing for the capacity. Your where clause should be looking at the id column of the program_classes record.

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -0,0 +1,164 @@
...
+            hour12: true
+        } as Intl.DateTimeFormatOptions;
+        return `${startDateTime.toLocaleTimeString([], options)} – ${endDateTime.toLocaleTimeString([], options)}`;
+    }
+
+    function getStatus(event: ClassEventInstance): string {
+        const eventDate = new Date(event.date);
+        eventDate.setHours(0, 0, 0, 0);
+        const today = new Date();
+        today.setHours(0, 0, 0, 0);
```
calisio, line 66: can condense
```
const eventDate = new Date(event.date).setHours(0, 0, 0, 0);
const today = new Date().setHours(0, 0, 0, 0);
```

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+            if (
+                enrolled !== undefined &&
+                event.attendance_records.length === enrolled
+            ) {
+                return 'Marked';
+            } else {
+                return 'Not Marked';
+            }
+        }
```
calisio, line 78:
```
return eventDate > today ? 'Scheduled' : event.attendance_records.length === enrolled ? 'Marked' : 'Not Marked';
```

calisio: https://github.com/user-attachments/assets/1e90fb90-dd37-4948-b494-9fa27ddc83ca

As you can see in the video, on the row it says 0/10 attended the class and shows the attendance as "marked" (which in this case implies that every student has an attendance marked, since the attendance records length is equal to the size of the enrolled for the class). Therefore, when I click inside to view that attendance, I expect that the attendance will show all students with some sort of absent marked (either excused or unexcused). Right now, nothing is marked (even though in our backend we seem to have some data for them?).

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    return (
+        <div className="p-4">
+            {isLoading && <div>Loading...</div>}
+            {error && <div className="error">Error loading data</div>}
```
calisio, line 84: i assume you meant `text-error`? Also, should be a `<p>` tag

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -0,0 +1,164 @@
...
+                                            event.date
+                                        ).toLocaleDateString()}
+                                    </td>
+                                    <td className="px-4">
+                                        {formatClassTime(
+                                            event.date,
+                                            event.class_time
+                                        )}
+                                    </td>
+                                    <td className="px-5">{`${getPresentCount(event.attendance_records)} ${capacity ? `/ ${enrolled}` : ''}`}</td>
```
calisio, line 116: confused why this is checking capacity if it doesn't have to do with capacity? Not sure if I am missing something

backend/src/models/users.go
calisio, line 1: was this needed? Honestly just curious because I wasn't sure how this fits into the PR (or what difference it makes?) I am indifferent on which it has

backend/src/handlers/events_attendance_handler.go
```diff
@@ -20,31 +24,42 @@ func (srv *Server) handleGetAttendeesForClass(w http.ResponseWriter, r *http.Req
...
+		return newJSONReqBodyServiceError(err)
 	}
-	date := r.URL.Query().Get("date")
-	attendance, err := srv.Db.LogUserAttendance(eventID, userID, date)
+	attendance, err := srv.Db.LogUserAttendance(&eventAttendance)
 	if err != nil {
 		return newDatabaseServiceError(err)
 	}
 	return writeJsonResponse(w, http.StatusOK, attendance)
 }
```
calisio, line 47: do we use this anywhere? Is this api and function necessary? I didnt see it being called in the frontend anywhere and it seems like its doing the same as `handleAddAttendanceForEvent`, but correct me if I am wrong?

frontend/src/Pages/ClassEvents.tsx
CK-7vn: Guess i'm not really sure on that one, i'll look into it.

backend/src/models/users.go
CK-7vn: ??

CK-7vn: Ya, we're switching over (and should be) all interface{}'s to any, it's just a new Go convention/standard, so when I notice them in the codebase I've been chanigng them.

calisio: Okay awesome, good to know. I will do the same when I come across

backend/src/database/events_attendance.go
```diff
@@ -54,3 +56,73 @@ func (db *DB) DeleteAttendance(eventId, userId int, date string) error {
...
+		FROM program_class_enrollments AS e
+		JOIN users AS u ON u.id = e.user_id
+		LEFT JOIN program_class_event_attendance AS a
+			ON a.user_id = e.user_id AND a.event_id = ? AND a.date = ?
+		WHERE e.class_id = ?`
+
+	args := []any{eventID, date, classID}
+
+	if qryCtx.Search != "" {
+		searchTerm := "%" + strings.ToLower(qryCtx.Search) + "%"
```
PThorpe92, line 75: `qryCtx.SearchQuery()` method does this for you :)

Below, after that pg extension is added, we can alter this search sql to fuzzy match on the user info

CK-7vn: ahh, ok ok, cool. I like it, and go figure, should've known that ><

frontend/src/common.ts
```diff
@@ -707,6 +707,48 @@ export interface ProgramClassEvent {
...
+    created_at: Date;
+    updated_at: Date;
+    event_id: number;
+    user_id: number;
+    date: string;
+    attendance_status: Attendance;
+    note: string;
+}
+
+export interface ProgramClassEnrollment {
```
carddev81, line 728: I may be missing something here, but I don't see where this enum is being used?

carddev81 (changes requested):
I would go back through all of your handlers that are Creating/Updating/Deleting data and adding critical values for auditing purposes using the log.add() function. @PThorpe92 can you confirm that we should be adding these log.add statements?

frontend/src/common.ts
```diff
@@ -959,3 +1001,18 @@ export type UserAccountHistoryAction =
     | 'facility_transfer'
     | 'set_password'
     | 'reset_password';
+
+export enum Months {
```
carddev81, line 1005: I may be missing something here, but I don't see where this enum is being used?

backend/src/handlers/events_attendance_handler.go
```diff
@@ -20,29 +23,26 @@ func (srv *Server) handleGetAttendeesForClass(w http.ResponseWriter, r *http.Req
...
 	if err != nil {
-		return newBadRequestServiceError(err, "user ID")
+		return newJSONReqBodyServiceError(err)
 	}
-	date := r.URL.Query().Get("date")
-	attendance, err := srv.Db.LogUserAttendance(eventID, userID, date)
-	if err != nil {
-		return newDatabaseServiceError(err)
+	for _, att := range attendances {
+		if _, err := srv.Db.LogUserAttendance(&att); err != nil {
```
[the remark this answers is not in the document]
carddev81: yes, but you would control the transaction with Begin() [transaction start] | loop through records executing the insert/update query (this will build a stack of statements within the transaction) | on error you would call Rollback() | if everything is successful you would call Commit() (stack of statements are executed once commit() is fired off.

The database should be able to handle a large batch of statements so you should need to worry about committing until you are completely finished with the loop. I was used to doing 1000 statements per batch.

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -0,0 +1,183 @@
...
+            ? 'Scheduled'
+            : event.attendance_records.length === enrolled
+              ? 'Marked'
+              : 'Not Marked';
+    }
+
+    return (
+        <div className="p-4">
+            <div className="flex mb-4 justify-start">
+                <DateInput
```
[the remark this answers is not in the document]
carddev81: I also get a 502 response in the browser console

backend/src/database/class_events.go
```diff
@@ -442,3 +446,129 @@ func generateEventInstances(event models.ProgramClassEvent, startDate, endDate t
...
+			Date:              occDateStr,
+			AttendanceRecords: relevantAttendance,
+		}
+		instances = append(instances, instance)
+	}
+	slices.Reverse(instances)
+
+	qryCtx.Total = int64(len(instances))
+
+	if !qryCtx.All {
```
carddev81, line 561: Just curious why this value is checked against? Didn't see the `all` query parameter on the request endpoint:

`/api/program-classes/class/${class_id}/events?month=${month}&year=${year}&page=${pageQuery}&per_page=${perPage}`

carddev81 (changes requested):

carddev81 (approved):
Tested good!

backend/src/handlers/events_attendance_handler.go
```diff
@@ -11,35 +11,29 @@ import (
 func (srv *Server) registerAttendanceRoutes() []routeDef {
 	axx := models.Feature(models.ProviderAccess)
 	return []routeDef{
-		{"GET /api/program-classes/{id}/attendees", srv.handleGetAttendeesForClass, true, axx},
-		{"GET /api/program-classes/{class_id}/event-attendance", srv.handleGetEventAttendance, true, axx},
-		{"POST /api/events/{id}/attendances", srv.handleAddAttendanceForEvent, true, axx},
-		{"DELETE /api/events/{id}/attendees/{user_id}", srv.handleDeleteAttendee, true, axx},
+		{"GET /api/program-classes/{class_id}/events/{event_id}/attendance", srv.handleGetEventAttendance, true, axx},
```
PThorpe92, line 14: we need to keep the routes consistent. this will only become more and more important

backend/src/database/events_attendance.go
```diff
@@ -101,17 +103,8 @@ func (db *DB) GetEnrollmentsWithAttendanceForEvent(qryCtx *models.QueryContext,
 			a.note AS note`
 	finalQuery := selectClause + baseQuery
 
-	allowedOrderColumns := map[string]string{
```
PThorpe92: I used to do this a lot but its a lot of overhead compared to the iteration and comparison so we should avoid it

PThorpe92 (approved):

PR #834, opened 2025-04-15
Added sort to Search Modal dropdown

corypride:
## Description of the change

Added sort to array.

## Screenshot(s)

image

---
  - https://app.asana.com/0/0/1209924956055675

### DIFF
--- a/frontend/src/Components/LibrarySearchResultsModal.tsx
+++ b/frontend/src/Components/LibrarySearchResultsModal.tsx
@@ -34,6 +34,7 @@ const LibrarySearchResultsModal = forwardRef<
     const { libraryOptions } = (useLoaderData() as {
         libraryOptions: Option[];
     }) || { libraryOptions: [] };
+
     const navigate = useNavigate();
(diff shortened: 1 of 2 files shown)


jtucholski (approved):

CK-7vn (approved):

frontend/src/Components/LibrarySearchResultsModal.tsx
```diff
@@ -34,6 +34,7 @@ const LibrarySearchResultsModal = forwardRef<
     const { libraryOptions } = (useLoaderData() as {
         libraryOptions: Option[];
     }) || { libraryOptions: [] };
+
```
CK-7vn, line 37: If it's not too much of a hassle (super nitpicky) can we just remove this extra empty line that was added? It just adds the file to the git history, which is unnecessary for a newline, unless you had a reason for adding it there, then totally keep it, but if it was just an accidental than we should definitely remove it so it's not added to the history.

PR #835, opened 2025-04-16
invalid character weird UX behavior - Fix

corypride:
## Description of the change
added error logic to submit handler

## Screenshot(s)

https://github.com/user-attachments/assets/e90771a2-a9ec-4b55-a883-62e2fa698cd5

---
  - https://app.asana.com/0/0/1209977328642898

### DIFF
--- a/frontend/src/Components/modals/AddUserModal.tsx
+++ b/frontend/src/Components/modals/AddUserModal.tsx
@@ -63,7 +63,6 @@ export const AddUserModal = forwardRef(function (
                         }
                     });
                     throw new Error(error_msgs.userexists);
-                    break;
                 }
                 case 'alphanum': {
                     setFormError({
@@ -74,7 +73,6 @@ export const AddUserModal = forwardRef(function (
                         }
                     });
                     throw new Error(error_msgs.alphanum);
-                    break;
                 }
             }
         } else {
--- a/frontend/src/Components/modals/AddUserModal.tsx
+++ b/frontend/src/Components/modals/AddUserModal.tsx
@@ -49,15 +49,20 @@ export const AddUserModal = forwardRef(function (
         );
         if (!response.success) {
             const msg = response.message.trim();
+            const error_msgs = {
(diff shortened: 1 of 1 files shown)


frontend/src/Components/modals/AddUserModal.tsx
```diff
@@ -49,31 +49,37 @@ export const AddUserModal = forwardRef(function (
...
                 case 'userexists': {
                     setFormError({
                         name: 'username',
                         error: {
                             type: 'custom',
-                            message: 'Username already exists'
+                            message: error_msgs.userexists
                         }
                     });
+                    throw new Error(error_msgs.userexists);
```
jtucholski, line 65: @corypride It looks like the fix was adding the `else` block down below. You don't need to throw these errors here. Doing so makes the `break` statements unreachable.

frontend/src/Components/modals/AddUserModal.tsx
```diff
@@ -49,31 +49,37 @@ export const AddUserModal = forwardRef(function (
...
                     setFormError({
                         name: 'username',
                         error: {
                             type: 'custom',
-                            message:
-                                'Username must contain only letters and numbers'
+                            message: error_msgs.alphanum
                         }
                     });
+                    throw new Error(error_msgs.alphanum);
```
jtucholski, line 76: Same here as above.

jtucholski (changes requested):
Looks good. Needs one small change and then it is good to go.

frontend/src/Components/modals/AddUserModal.tsx
corypride: Ok then I need another slight fix for the part of the ticket where you want the invalid characters to remain in the form fields. Without throwing the error, the form errors will not be an instance of Errors and the form values will be reset each time. Note: onSubmitHandler on the FormModal.tsx

jtucholski (approved):

PR #846, opened 2025-04-23
Disable add class button

corypride:
## Description of the change
Disable the Add Class button for Facilities where a Program isn't Offered

## Screenshot(s)

https://github.com/user-attachments/assets/8c2b6a1d-b63e-41e4-a91a-b6f92cb4b9c3

---
  - https://app.asana.com/0/0/1209962528167593

### DIFF
--- a/frontend/src/Pages/ProgramOverviewDashboard.tsx
+++ b/frontend/src/Pages/ProgramOverviewDashboard.tsx
@@ -26,9 +26,12 @@ import { XMarkIcon } from '@heroicons/react/24/solid';
 import API from '@/api/api';
 import { useCheckResponse } from '@/Hooks/useCheckResponse';
 import { useNavigate } from 'react-router-dom';
+import { useAuth } from '@/useAuth';
 
 export default function ProgramOverview() {
     const { id } = useParams<{ id: string }>();
+    const user = useAuth();
+    const userFacilityId = user.user?.facility_id;
     const program = useLoaderData() as Program;
     const [page, setPage] = useState(1);
     const [perPage, setPerPage] = useState(20);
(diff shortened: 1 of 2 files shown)


frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -62,7 +65,7 @@ export default function ProgramOverview() {
     } else if (classesError) {
         navigate('/error');
     }
-
+    // console.log(program.facilities.includes())
```
CK-7vn, line 68: Probably should make sure to remove this comment/console log.

frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -247,6 +250,11 @@ export default function ProgramOverview() {
                             onClick={() =>
                                 navigate(`/programs/${id}/classes/new`)
                             }
+                            disabled={
```
CK-7vn, line 253: I think we should probably gray out the button and maybe add a tooltip in here for when it's disabled, that way the user's not going "Ok...why isn't this app working? Whats going on here?" @jtucholski What do you think?

jtucholski (changes requested):

frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -243,10 +248,15 @@ export default function ProgramOverview() {
...
                         <button
-                            className="button flex items-center"
+                            className={`button flex items-center ${isAddClassDisabled ? 'opacity-50 cursor-not-allowed' : ''}`}
                             onClick={() =>
                                 navigate(`/programs/${id}/classes/new`)
                             }
+                            disabled={
+                                !program.facilities.some(
+                                    (facility) => facility.id === userFacilityId
+                                )
```
jtucholski, line 258: @corypride Rather than defining it again, you can reuse the same expression value that you added to handle conditional styling.

jtucholski (changes requested):
One small change and then it's good to go.

jtucholski (approved):

CK-7vn (approved):

PR #848, opened 2025-04-25
Prepare middleware for offline deployments

PThorpe92:
This PR changes the middleware to prefer writing to disk instead of assuming we have s3 access, which is preferred because aws's `mounts3` is great and each pod will have a PVC for a subdirectory of the main bucket.

We can still use presigned URL's in the mean-time for serving videos, but eventually will probably have the same s3 directory mounted to nginx to serve them and then we can setup a local cache for content at the deployments.

### DIFF
--- a/provider-middleware/video_dl.go
+++ b/provider-middleware/video_dl.go
@@ -9,15 +9,11 @@ import (
 	"io"
 	"mime/multipart"
 	"net/http"
-	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/config"
-	"github.com/aws/aws-sdk-go-v2/service/s3"
-	"github.com/sirupsen/logrus"
 	"github.com/wader/goutubedl"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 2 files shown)


provider-middleware/video_dl.go
```diff
@@ -28,196 +24,133 @@ type Thumbnail struct {
...
 	}
-	logger().Infof("Successfully uploaded file to %s/%s with ETAG: %s", yt.bucketName, video.GetS3KeyJson(), *jsonOutput.ETag)
-	return nil
-}
-
-func (yt *VideoService) putAllCurrentVideoMetadata(ctx context.Context) error {
-	if yt.s3Svc == nil {
-		logger().Info("no s3 client found, skipping sync")
+	// early return if we are already sync'd (common path)
+	if int(count) == len(jsonIDs) {
```
carddev81, line 82: Is it possible that the counts match but ids are out sync? just me thinking

provider-middleware/video_dl.go
```diff
@@ -28,196 +24,133 @@ type Thumbnail struct {
...
-	page, err := yt.s3Svc.ListObjectsV2(ctx, input)
-	if err != nil {
-		return fmt.Errorf("failed to list objects in bucket %s: %w", yt.bucketName, err)
-	}
-	for _, item := range page.Contents {
-		if strings.HasSuffix(*item.Key, ".json") {
-			jsonFiles = append(jsonFiles, *item.Key)
+	// load any new videos from JSON that are not in DB
+	for _, id := range jsonIDs {
+		if _, exists := existingMap[id]; exists {
```
carddev81, line 103: Could you do this instead? If so you can remove a few declaration lines above this.

```
		if slices.Contains(existingVideos, id){
			continue
		}
```

provider-middleware/video_dl.go
```diff
@@ -562,10 +438,12 @@ func (yt *VideoService) downloadVideo(ctx context.Context, vidInfo *goutubedl.Re
...
 		vidInfo = &result
 	}
-	if yt.videoExistsInS3(ctx, vidInfo.Info.ID) {
+
+	// if we already have the video on disk, skip the download
+	if yt.videoIsPresent(vidInfo.Info.ID) {
 		return nil
 	}
-	downloadResult, err := vidInfo.Download(ctx, "best")
+	downloadResult, err := vidInfo.Download(ctx, "b")
```
carddev81, line 446: not sure if you meant to change `best` to `b` here? Will this still grab the best?

carddev81 (approved):
@PThorpe92 Tested good, just left a couple comments and a question.

provider-middleware/video_dl.go
PThorpe92: that would be O(n2), the map is to prevent doing a full iteration of the existing videos for every found id

provider-middleware/video_dl.go
PThorpe92: yeah these are never deleted so that would be very unlikely

PR #849, opened 2025-04-28
fix: add logic check for class status of completed and cancelled per ticket 167

carddev81:
## Description of the change

Added code so that when a class's status is set to 'Completed' or 'Cancelled' it cannot be change-able.
  
- **Related issues**: Link to Asana ticket [Class status should not be change-able after Completed or Canceled](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210072188481728?focus=true)


jtucholski (approved):
I like that you mapped out the empty array for the closed states. Makes it clear the intention.

PThorpe92 (approved):

PR #850, opened 2025-04-29
fix: remove class status Pending per ticket 134

carddev81:
## Description of the change

- Removed the class status of Pending from backend api
- Removed the class status of Pending from frontend code, 
- Removed the class status of Pending from database enum type.  
    - The enum type that was being used was named section_status so I renamed this to be class_status.

- **Related issues**: Link to Asana ticket [Creating a class w/ a "Pending" status means no status field & not-editable](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209932428651178?focus=true)


backend/seeder/main.go
```diff
@@ -232,7 +232,6 @@ func seedTestData(db *gorm.DB) {
 		statuses := [8]models.ProgramEnrollmentStatus{
```
PThorpe92, line 232: the array length needs to be updated. looks like it is [6] now somehow

PR #851, opened 2025-04-29
fix: cancel enrollments upon cancelled class

CK-7vn:
## Description of the change
This pull request makes sure that when an admin cancels a class all Enrolled students enrollment status is marked as Cancelled

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210072150791608?focus=true

### DIFF
--- a/backend/src/database/class_enrollments.go
+++ b/backend/src/database/class_enrollments.go
@@ -42,15 +42,14 @@ func (db *DB) GetEnrollmentsForClass(page, perPage, classId int) (int64, []model
 	return total, content, nil
 }
 
-func (db *DB) GetAllActiveEnrollmentsForClassByClassID(classID int) ([]models.ProgramClassEnrollment, error) {
-	var content []models.ProgramClassEnrollment
-	err := db.
-		Where("class_id = ? AND enrollment_status = ?", classID, models.Enrolled).
-		Find(&content).Error
-	if err != nil {
-		return nil, newNotFoundDBError(err, "class enrollments")
+func (db *DB) GetAllActiveEnrollmentsForClassIDs(classIDs []int) ([]models.ProgramClassEnrollment, error) {
+	var enrollments []models.ProgramClassEnrollment
+	if err := db.
+		Where("class_id IN ? AND enrollment_status = ?", classIDs, models.Enrolled).
+		Find(&enrollments).Error; err != nil {
+		return nil, newNotFoundDBError(err, "active class enrollments")
 	}
-	return content, nil
+	return enrollments, nil
 }
 
 func (db *DB) GetProgramClassEnrollmentsForFacility(page, perPage int, facilityID uint) (int64, []models.ProgramClassEnrollment, error) {
@@ -181,6 +180,17 @@ func (db *DB) UpdateProgramClassEnrollments(classId int, userIds []int, status s
 		Where("class_id = ? AND user_id IN (?)", classId, userIds).
 		Update("enrollment_status", status).Error; err != nil {
 		return newUpdateDBError(err, "class enrollment status")
+	}
+	return nil
+}
+
+func (db *DB) UpdateProgramClassEnrollmentsPerEnrollment(classIDs []int, status string) error {
+	if err := db.
+		Model(&models.ProgramClassEnrollment{}).
+		Where("class_id IN ? AND enrollment_status = ?", classIDs, models.Enrolled).
+		Update("enrollment_status", status).
+		Error; err != nil {
+		return newUpdateDBError(err, "class enrollment statuses")
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


backend/src/handlers/classes_handler.go
```diff
@@ -116,12 +116,28 @@ func (srv *Server) handleUpdateClasses(w http.ResponseWriter, r *http.Request, l
...
+			}
+			var userIDs []int
+			for _, user := range enrollments {
+				userIDs = append(userIDs, int(user.UserID))
+			}
+			err = srv.Db.UpdateProgramClassEnrollments(classID, userIDs, string(models.EnrollmentCancelled))
+			if err != nil {
+				return newDatabaseServiceError(err)
+			}
+		}
```
calisio, line 139: I think generally we want to avoid calling db queries in a loop, and instead would prefer to do a batch query to just have one call to the db (learned this from Rich 🙂). 
Would we be able to fetch all active enrollments for all classIDs in a single query, and then update all the enrollments for all classIDs in a single query? So instead of passing in one classID you could pass in an array of classIDs (`classIDs []int`), and then you could do `Where("class_id IN ? AND enrollment_status = ?", classIDs, models.Enrolled)`, and something similar for `UpdateProgramClassEnrollments`?

CK-7vn: Well, right now you can't even update more than one class at a time, so, yes, i don't disagree, I put it in a loop because of the fact that this handler CAN accept multiple classID's, but, only for archiving. So, I don't know, I mean, ya, i can do it in a batch. No biggie I suppose.

carddev81: Yes, if you can avoid making a database call for every iteration of a loop you should. It is recommended to either put your looping logic within a database transaction or in this case the best option would be if you can modify the update query to be as Carolina suggested `where class_id in (?) and enrollment_status = ?`. One catch-all statement.

backend/src/database/class_enrollments.go
```diff
@@ -42,6 +42,16 @@ func (db *DB) GetEnrollmentsForClass(page, perPage, classId int) (int64, []model
...
+func (db *DB) GetAllActiveEnrollmentsForClassIDs(classIDs []int) ([]models.ProgramClassEnrollment, error) {
+	var enrollments []models.ProgramClassEnrollment
+	if err := db.
+		Where("class_id IN ? AND enrollment_status = ?", classIDs, models.Enrolled).
+		Find(&enrollments).Error; err != nil {
+		return nil, newNotFoundDBError(err, "active class enrollments")
+	}
+	return enrollments, nil
+}
+
```
calisio, line 54: Do we even need this anymore since we don't use it anywhere? Maybe just delete?

backend/src/handlers/classes_handler.go
```diff
@@ -116,12 +116,21 @@ func (srv *Server) handleUpdateClasses(w http.ResponseWriter, r *http.Request, l
...
-	classMap := make(map[string]interface{})
+	classMap := make(map[string]any)
 	if err := json.NewDecoder(r.Body).Decode(&classMap); err != nil {
 		return newJSONReqBodyServiceError(err)
 	}
 	claims := r.Context().Value(ClaimsKey).(*Claims)
 	classMap["update_user_id"] = claims.UserID
+
+	if classMap["status"] == string(models.EnrollmentCancelled) {
+		if err := srv.Db.UpdateProgramClassEnrollmentsPerEnrollment(
```
carddev81, line 127: @CK-7vn This looks good, but I think we should rework this a bit so that there is only one database transaction created and one commit.  So maybe update the `UpdateProgramClasses` database call to handle the entire transaction in case of failure you can rollback? Something like:

```
	tx := db.Begin()
	if tx.Error != nil {
		//handle error
	}

	if status === string(models.Cancelled) {
		if err := tx.Model(&models.ProgramClassEnrollment{}).Where("class_id IN ? AND enrollment_status = ?", classIDs, models.Enrolled).Update("enrollment_status", status).
			Error; err != nil {
			tx.Rollback()
			//handle error
		}
	}
	if err := tx.Model(&models.ProgramClass{}).Where("id IN ?", classIDs).Updates(classMap).Error; err != nil {
		tx.Rollback()
		//handle error
	}

	if err := tx.Commit().Error; err != nil {
		//handle error
	}

```

jtucholski (approved):

backend/src/database/class_enrollments.go
```diff
@@ -174,6 +174,27 @@ func (db *DB) UpdateProgramClassEnrollments(classId int, userIds []int, status s
 	return nil
 }
 
+func (db *DB) UpdateProgramClassEnrollmentsPerEnrollment(classIDs []int, status string) error {
```
carddev81, line 177: just making sure I don't forget this: move logic into UpdateProgramClasses and make into one db transaction and then you should be able to remove this method altogether

carddev81 (changes requested):

carddev81 (approved):
Looks great! Tested good!

PR #853, opened 2025-04-30
Enhance transfer validation logic

corypride:
## Description of the change

 Added functionality where the confirmation button should be disabled by default and only be enabled when the user enters the DOC ID correctly.

## Screenshot(s)
https://github.com/user-attachments/assets/b553af8e-1cf2-4dcf-a9fd-9802a3b942c2

---
  - https://app.asana.com/0/0/1209928767233485

### DIFF
--- a/frontend/src/Components/inputs/TextInput.tsx
+++ b/frontend/src/Components/inputs/TextInput.tsx
@@ -18,7 +18,6 @@ interface TextProps {
     disabled?: boolean;
     defaultValue?: string;
     onChange?: React.ChangeEventHandler<HTMLInputElement>;
-    // isValidDoc?: boolean; //TODO:  remove
 }
 export function TextInput({
     label,
--- a/frontend/src/Components/inputs/TextInput.tsx
+++ b/frontend/src/Components/inputs/TextInput.tsx
@@ -18,7 +18,7 @@ interface TextProps {
     disabled?: boolean;
     defaultValue?: string;
     onChange?: React.ChangeEventHandler<HTMLInputElement>;
-    isValidDoc?: boolean;
+    // isValidDoc?: boolean; //TODO:  remove
 }
 export function TextInput({
     label,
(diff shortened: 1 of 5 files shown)


jtucholski (changes requested):
@corypride

My comments in the PR focus on removing the DOCID specific logic from the SubmitButton, TextInput, FormModal, and `index.ts`. 

Your approach at defining the logic in VerifyResidentModal is correct. 

You can generalize the FormModal and allow for the parent containers to pass in something like `enableSubmit` (default value of true) which would be used to drive the logic for when the SubmitButton is enabled or not.

frontend/src/Components/inputs/TextInput.tsx
```diff
@@ -18,6 +18,7 @@ interface TextProps {
     disabled?: boolean;
     defaultValue?: string;
     onChange?: React.ChangeEventHandler<HTMLInputElement>;
+    // isValidDoc?: boolean; //TODO:  remove
```
carddev81, line 21: should this comment be removed?

frontend/src/Components/modals/VerifyResidentModal.tsx
```diff
@@ -30,13 +30,17 @@ export const VerifyResidentModal = forwardRef(function (
     const [errors, setErrors] = useState('');
     const [facility, setFacility] = useState('');
     const facilities = useLoaderData() as Facility[];
+    const target_user = target?.user.doc_id;
```
carddev81, line 33: i would rename this to include docId? something like targetDocId?

carddev81 (changes requested):

frontend/src/Components/modals/index.ts
```diff
@@ -9,7 +9,7 @@ import {
 } from '@/common';
 import { KeyedMutator } from 'swr';
 import { Validate } from 'react-hook-form';
-import React from 'react';
+import React, { JSX } from 'react';
```
jtucholski, line 12: @corypride Was it necessary to import `JSX`?

jtucholski (commented):
Just the one comment about the import, otherwise I like the solution you took to generalize this approach!

PR #854, opened 2025-04-30
fix: fix dropdown display names for timezones

CK-7vn:
## Description of the change

Fixes how timezones are being displayed, and adds an optional dispalyNames parameter to the DropdownInput for future enum display issues. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210072188481737?focus=true

## Screenshot(s)
image

### DIFF
--- a/frontend/src/Components/modals/index.ts
+++ b/frontend/src/Components/modals/index.ts
@@ -49,6 +49,7 @@ export interface Input {
     interfaceRef: string;
     required: boolean;
     enumType?: Record<string, string>;
+    displayNames?: Record<string, string>;
     length?: number;
     pattern?: Pattern;
     validate?:
--- a/frontend/src/Components/modals/index.ts
+++ b/frontend/src/Components/modals/index.ts
@@ -49,7 +49,6 @@ export interface Input {
     interfaceRef: string;
     required: boolean;
     enumType?: Record<string, string>;
-    displayNames?: Record<string, string>;
     length?: number;
     pattern?: Pattern;
     validate?:
--- a/frontend/src/Components/inputs/DropdownInput.tsx
+++ b/frontend/src/Components/inputs/DropdownInput.tsx
@@ -36,7 +36,7 @@ export function DropdownInput({
             >
                 {Object.entries(enumType).map(([key, value]) => (
                     <option key={key} value={value}>
-                        {value}
+                        {key}
                     </option>
                 ))}
             </select>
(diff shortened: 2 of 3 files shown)


calisio (commented):
Okay, this is definitely not what you are going to want to hear 😂 
@carddev81 and I were actually talking about this when he was creating some dropdowns and we noticed that dropdown doesn't even use the key anywhere in the component, when it probably should? 
I think to fix this issue (and probably some other issues), we should probably refactor dropdown and everywhere where it is used to have the key be the display name, and then have the value be the value submitted to the backend (I think... it might be the opposite but brain is working slow at the moment...). I think this would solve this issue so that way we could just have the key be "Chicago" and then the value be "America/Chicago". Thoughts?

CK-7vn:
> Okay, this is definitely not what you are going to want to hear 😂 @carddev81 and I were actually talking about this when he was creating some dropdowns and we noticed that dropdown doesn't even use the key anywhere in the component, when it probably should? I think to fix this issue (and probably some other issues), we should probably refactor dropdown and everywhere where it is used to have the key be the display name, and then have the value be the value submitted to the backend (I think... it might be the opposite but brain is working slow at the moment...). I think this would solve this issue so that way we could just have the key be "Chicago" and then the value be "America/Chicago". Thoughts?

I mean, we can do that, but whether we use the key or the value, there will still be issues with what's being displayed...unless i'm missing something...like...if we have     'america/los_angeles' = 'America/Los_Angeles', it's still going to display America/Los_angeles, unless i'm totally off base on something? no?

frontend/src/Components/modals/index.ts
```diff
@@ -87,6 +88,15 @@ export interface TargetItem<T> {
 }
 
 export { FormModal } from './FormModal';
+const getTimezoneDisplayName = (tz: string) => {
+    const parts = tz.split('/');
```
PThorpe92, line 92: I think it prob makes more sense to show them regular timezones
```ts
const timezones = {
"EST": "America/New_York",
"CST": "America/Chicago",
"PST": "America/Los_Angeles",
}
```
then display the keys and use the values

CK-7vn:
image

frontend/src/Components/modals/index.ts
```diff
@@ -49,6 +49,7 @@ export interface Input {
     interfaceRef: string;
     required: boolean;
     enumType?: Record<string, string>;
+    displayNames?: Record<string, string>;
```
calisio, line 52: i dont think we need this anymore?

calisio (approved):
just remove the line in index.ts for input, and I just want to double check that you made sure all of the other places `DropdownInput.tsx` was used is not altered by this. Add class looks a little different since they are all uppercase now (but honestly I feel like its fine). Just want to make sure I don't miss anywhere else where this is used- I don't think it is many other places.
Screenshot 2025-05-02 at 11 33 50 AM

PR #855, opened 2025-04-30
fix: fix hyphenated name validation

CK-7vn:
## Description of the change
This allows hyphens in first and last names, and updates validation logic for those two struct fields for our users model. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209977398530311?focus=true

## Screenshot(s)

image

### DIFF
--- a/backend/src/database/DB.go
+++ b/backend/src/database/DB.go
@@ -25,18 +25,9 @@ func NewDB(db *gorm.DB) *DB {
 	return &DB{db}
 }
 
-func ValidateNameCharacters(fl validator.FieldLevel) bool {
-	for _, char := range fl.Field().String() {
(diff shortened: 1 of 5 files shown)


backend/src/database/DB.go
```diff
@@ -25,6 +25,15 @@ func NewDB(db *gorm.DB) *DB {
 	return &DB{db}
 }
 
+func ValidateNameCharacters(fl validator.FieldLevel) bool {
```
PThorpe92, line 28: the only place the `ValidateAlphaNumSpace`/ 'alphanumspace'  was used was the same instances, so prob should change that one instead of make a whole new one here

calisio (approved):

PThorpe92 (approved):

PR #860, opened 2025-04-30
feat: add script to add kiwix libraries to dedicated instance

PThorpe92:
https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210072150791630

### DIFF
--- a/config/kiwix.sh
+++ b/config/kiwix.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+set -euo pipefail
+
+S3_BUCKET="" # needs bucket name
+INSTANCE_ID="" # needs instance ID
+AWS_REGION="us-east-1"
+
+if [ -z "$@" ]; then
+  echo "No arguments provided. Please provide a .zim file path or HTTPS URL."
+  read -p "Enter .zim file path or HTTPS URL (e.g., https://library.kiwix.org/...): " INPUT
+else
+  INPUT="$1"
+fi
+
+COMMANDS_ARRAY=""
+ZIM_FILE_NAME=""
+
+if [[ "$INPUT" =~ ^https?:// ]]; then
+  ZIM_FILE_NAME=$(basename "$INPUT")
+  if [[ ! "$ZIM_FILE_NAME" =~ \.zim$ ]]; then
+    echo "Error: remote file must end in .zim"
+    exit 1
+  fi
+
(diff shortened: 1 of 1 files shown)


config/kiwix.sh
```diff
@@ -0,0 +1,67 @@
...
+
+COMMANDS_ARRAY=""
+ZIM_FILE_NAME=""
+
+if [[ "$INPUT" =~ ^https?:// ]]; then
+  ZIM_FILE_NAME=$(basename "$INPUT")
+  if [[ ! "$ZIM_FILE_NAME" =~ \.zim$ ]]; then
+    echo "Error: remote file must end in .zim"
+    exit 1
+  fi
```
calisio, line 24: forgot to close this if statement out- you are missing a `fi` at the end of this block

config/kiwix.sh
```diff
@@ -0,0 +1,67 @@
...
+
+  echo "Uploading $ZIM_FILE_NAME to s3://$S3_BUCKET/ ..."
+  aws s3 cp "$INPUT" "s3://$S3_BUCKET/$ZIM_FILE_NAME" --region "$AWS_REGION"
+
+  COMMANDS_ARRAY=(
+	"set -e"
+	"aws s3 cp \"s3://$S3_BUCKET/$ZIM_FILE_NAME\" \"/zims/$ZIM_FILE_NAME\" --region \"$AWS_REGION\""
+	"/home/admin/.local/bin/kiwix-manage /zims/library.xml add \"/zims/$ZIM_FILE_NAME"\"
```
calisio, line 52: should this be
`"/home/admin/.local/bin/kiwix-manage /zims/library.xml add \"/zims/$ZIM_FILE_NAME\""` (closing quote should be after the `\`?)

carddev81 (approved):
Looks good, of course I was only able to test parts of it.

config/kiwix.sh
PThorpe92: This block continues, the COMMANDS_ARRAY below is still in the scope, as it has to download the zim file if it begins with `https://` :)

calisio: ohhh totally missed that sorry

config/kiwix.sh
PThorpe92: technically yes I believe you are right, seems odd that it still works this way

PR #861, opened 2025-04-30
fix: keep navbar pinned position across client route changes

PThorpe92:
https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209485070839570

the navbar now correctly reflects the state when a navigation item is clicked.


calisio (approved):

carddev81 (approved):
Weird issue with `Tour` occurred when logging on as a resident for the first time. This only happened when (1) logging out as superadmin (navbar not pinned) then (2) logging in first time as resident.

image

PR #862, opened 2025-05-01
fix: change error text, update error page capability

CK-7vn:
## Description of the change
- updates error page text coloring
- adds option to update error message
- adds option to update button text
- adds option to update navigation behavior of button

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209485070839577?focus=true

## Screenshot(s)
image
image

## Additional context
This started as just updating the text color of our error page, but then i realized we are kind of doing error's and error pages a little odd, or what felt a little odd, so I tried to make a dynamic page that we can render dynamically. I'm open to changes, and on some of the places we have the error component, i wasn't sure necessarily if we needed to pass the error type in or just leave it as the default, open to changes, or updates, just want us to have a future proof easy error page we can use anywheres


jtucholski (commented):
I like the approach that you took here. I think we need though if we finish the implementation.

For exampe, right now someone could assume that because `auth-required` is a valid error status then I should be able to see it. I'm unable to trigger it though. Did I do it wrong or is it not implemented throughout the codebase? 

Same goes for `unauthorized`. I can't get it to trigger. I tried logging in as a resident and going to `/learning-insights` which is what uses AdminLayer2 but all that happens is I get redirected to `/home`. Did I do it wrong?

What do you think? Do we need a separate ticket to implement these new error types elsewhere in the front end or is it something that we do in this PR?

frontend/src/Pages/Error.tsx
```diff
@@ -1,23 +1,71 @@
...
+    navigateTo?: string;
+}
+
+const errorMap: Record<
+    ErrorType,
+    { message: string; buttonText: string; onClick: () => void }
+> = {
+    unauthorized: {
+        message:
+            'You are not authorized to view this page, please contact your administrator if you believe you have reached this page in error.',
```
carddev81, line 18: Not sure about this, but the message doesn't match what the [ticket](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209485070839577?focus=true) calls for?

frontend/src/Pages/Error.tsx
```diff
@@ -1,23 +1,71 @@
...
+        buttonText: 'Home Page',
+        onClick: () => (window.location.href = AUTHCALLBACK)
+    },
+    'not-found': {
+        message: 'The page you requested was not found.',
```
carddev81, line 23: Same here about the message not matching

frontend/src/Pages/Error.tsx
```diff
@@ -1,23 +1,71 @@
...
+        onClick: () => (window.location.href = AUTHCALLBACK)
+    }
+};
+export default function Error({
+    type = 'server-error',
+    back = false,
+    message: customMessage,
+    navigateTo
+}: ErrorPageProps) {
+    const { message: defaultMessage, buttonText, onClick } = errorMap[type];
```
carddev81, line 44: i'm getting a linting error on `type` here? Error is: `Computed name [type] resolves to an `error` typed value.`

CK-7vn:
> I like the approach that you took here. I think we need though if we finish the implementation.
> 
> For exampe, right now someone could assume that because `auth-required` is a valid error status then I should be able to see it. I'm unable to trigger it though. Did I do it wrong or is it not implemented throughout the codebase?
> 
> Same goes for `unauthorized`. I can't get it to trigger. I tried logging in as a resident and going to `/learning-insights` which is what uses AdminLayer2 but all that happens is I get redirected to `/home`. Did I do it wrong?
> 
> What do you think? Do we need a separate ticket to implement these new error types elsewhere in the front end or is it something that we do in this PR?

Pretty sure this is because we're getting redirected from the backend, we can definitely go around and find we're there's potential redirects and either trigger the error page, or just continue with being redirected, i'm not entirely sure what exactly we want to do, but i'm definitely assuming that's why that's happening after looking more into it.

frontend/src/Pages/Error.tsx
```diff
@@ -1,23 +1,75 @@
...
+        message: 'The page you requested does not exist.',
+        buttonText: 'Home Page',
+        onClick: () => (window.location.href = AUTHCALLBACK)
+    },
+    'server-error': {
+        message: 'An unexpected error occurred. Please try again later.',
+        buttonText: 'Home Page',
+        onClick: () => (window.location.href = AUTHCALLBACK)
+    },
+    'auth-required': {
```
corypride, line 32: @CK-7vn what are the steps to reproduce this and unauthorized. I was able to test the others and they looked good?

CK-7vn: We've had a hard time duplicating this error type because we redirect from the backend so essentially, at least as it stands right now it in theory should be impossible, I added it so that we have it in there because there definitely could be a use case, it just depends on what kind of paradigm we want to go with for the error type, redirecting from the backend, or actually making the user aware of it. So, for the time being I think it's essentially just there. @jtucholski correct me if i'm wrong, were you ever able to get it to kick-off?

jtucholski: @CK-7vn  I hadn't been able to look into it further. I'll look at it today.

jtucholski, line 32: @CK-7vn Can we remove the auth-required option? The preferred outcome of an unauthenticated user trying to access a page behind authentication is that they get redirected to the auth page automatically.

jtucholski (changes requested):
Can you update the RouteGuard so that it navigates users who aren't authorized to the Error page unauthorized? 

Otherwise, it looks good and is good to go.

carddev81 (approved):
Good to go!

jtucholski (approved):
Nice, like this addition.

PR #870, opened 2025-05-06
check for existing user when loggging in to prevent ident mismatch

PThorpe92:
Fix login issue described in slack


CK-7vn (approved):
Looks good, seems to be fixed!

jtucholski:
Authentication still works, I'm not sure why this was needed?

jtucholski (approved):

PR #871, opened 2025-05-06
fix: modify dropdowns to be consistent for sorting resident names per ticket 139

carddev81:
## Description of the change

Modified dropdowns to be consistent for sorting resident names per ticket 139.

- **Related issues**: Link to Asana ticket is [Standardize on sorting fields for pages with Resident names](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209940257724121?focus=true)

### DIFF
--- a/backend/src/database/class_enrollments.go
+++ b/backend/src/database/class_enrollments.go
@@ -4,7 +4,6 @@ import (
 	"UnlockEdv2/src/models"
 	"context"
 	"fmt"
-	"strings"
 )
 
 func (db *DB) GetProgramCompletionsForUser(args *models.QueryContext, userId int, classId *int) ([]models.ProgramCompletion, error) {
@@ -202,12 +201,8 @@ func (db *DB) GetProgramClassEnrollmentsForProgram(args *models.QueryContext, cl
 	if err := tx.Count(&args.Total).Error; err != nil {
 		return nil, newNotFoundDBError(err, "program class enrollments")
 	}
-	orderBy := args.OrderClause()
-	if strings.Contains(orderBy, "name_last") {
-		orderBy = fmt.Sprintf("%s, name_first", orderBy)
-	}
 	if err := tx.Limit(args.PerPage).
-		Offset(args.CalcOffset()).Order(orderBy).
+		Offset(args.CalcOffset()).Order(adjustUserOrderBy(args.OrderClause())).
 		Find(&content).Error; err != nil {
 		return nil, newNotFoundDBError(err, "program class enrollments")
 	}
--- a/backend/src/database/class_enrollments.go
+++ b/backend/src/database/class_enrollments.go
@@ -4,6 +4,7 @@ import (
 	"UnlockEdv2/src/models"
 	"context"
 	"fmt"
+	"strings"
 )
 
 func (db *DB) GetProgramCompletionsForUser(args *models.QueryContext, userId int, classId *int) ([]models.ProgramCompletion, error) {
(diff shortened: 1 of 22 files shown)


jtucholski:
After reviewing the PR I realize the ticket was not clear. However the final implementation does achieve the goal of standardizing how we sort residents by name. 

It looks like the Enrollments page isn't properly sorting by last name ascending where are conflicts with the last name. I think it has something to do with the query using `name_full` for the concatenated column name.  Can you fix that? Good to go once that's addressed.
image

jtucholski:
@carddev81 You can keep the existing sort by options in there and this PR only focuses on the way we sort resident names.

jtucholski (approved):

backend/src/models/resource.go
```diff
@@ -98,6 +99,8 @@ func (q QueryContext) OrderClause() string {
 	val := fmt.Sprintf("%s %s", q.OrderBy, q.Order)
 	if val == " " {
 		return "created_at desc"
+	} else if strings.Contains(val, "name_last") {
+		val = fmt.Sprintf("%s, name_first", val)
```
PThorpe92, line 103: This has an issue (one that I at least somewhat tried to begin to address in #873)

What if `val` contains something like `pce.name_last`? or even `u.name_last`?

carddev81: @PThorpe92 I thought this would be okay since `name_first` is only found within the `users` table, so there shouldn't be any ambiguity. So using a table aliases like `u.` or `pce.` wouldn't be necessary in the sql?

PThorpe92: yeah right now that is the case, but this is used in quite a few places. What if we have a CTE that grabs those fields from users, then we select those fields from the CTE aliased to something else?

PThorpe92:
```sql
with (
   select * from users where role = 'student'
) as students
  select e.*, students.* from enrollments e join
```
I don't think we are doing this exact thing at the moment but it just seems feasible enough to not add here where it could be forgotten about. I'm open to pushback tho

carddev81: You made a good point on the CTE, but now I'm more concerned with `created_at desc` since it's within most tables? Seems like this should change too since ambiguity would may be a problem here too?

`pse.created_at` or `u.created_at`

carddev81: @PThorpe92 I feel like this is a better/cleanest option for fixing the possibility of an issue with name_first, let me know your thoughts:

export enum FilterResidentNames {
    'Resident Name (A-Z)' = 'name_last asc, name_first asc',
    'Resident Name (Z-A)' = 'name_last desc, name_first asc'
}

carddev81: @PThorpe92 made the change, I removed the name_first from the OrderClause() method and move it to the specific query calls.

backend/src/database/class_enrollments.go
```diff
@@ -201,8 +202,12 @@ func (db *DB) GetProgramClassEnrollmentsForProgram(args *models.QueryContext, cl
 	if err := tx.Count(&args.Total).Error; err != nil {
 		return nil, newNotFoundDBError(err, "program class enrollments")
 	}
+	orderBy := args.OrderClause()
+	if strings.Contains(orderBy, "name_last") {
```
PThorpe92, line 206: I know it's stupid but can we make a helper function like
```go
func adjustUserOrderBy(arg string) string {
    if strings.Contains(arg, "name_last") {
      return arg + ", name_first"
   }
return arg
}
```
just because we repeat this 3 or 4 times. (btw doing `+` instead of `fmt.Sprintf` is preferred when possible as its much cheaper)

carddev81: @PThorpe92 pushed the changes. No, not stupid, good programming strategy, should have done this in the first place.  I got in the habit of using the Sprintf function from it's consistent use. :(

PThorpe92 (approved):

PR #872, opened 2025-05-06
Create Resident overview of programs

corypride:
## Description of the change

Generate data that Help residents clearly see what they’re currently enrolled in, view upcoming classes, and track participation history.

## Screenshot(s)
image 
image
image
image

---
  - https://app.asana.com/0/0/1210072189020770


backend/src/database/users.go
```diff
@@ -542,23 +543,93 @@ func (db *DB) GetUserProgramInfo(args *models.QueryContext, userId int) ([]model
...
+		Group("pc.program_id, pc.name, pc.start_dt, pc.end_dt, pce.enrollment_status, pce.updated_at").
+		Find(&residentWeeklySchedule).Debug().Error
+
+	if err != nil {
+		return nil, NewDBError(err, "program_class_enrollments")
 	}
 
-	return userEnrollments, nil
+	return residentWeeklySchedule, nil
 }
```
jtucholski, line 635: I think I have two questions about this function. They might contradict each other, but here goes.

1. If I understand this correctly, it's looking for Classes with a START DATE between today and 7 days. The purpose of the weekly schedule is so that a Resident can answer the question, "What classes do I have this week? Or Next week, etc." A class can start and meet daily, weekly, monthy over an extended period of time. 

2. Previous point aside...this function isn't much different than `GetUserProgramClassHistory` and `GetUserProgramInfo`, right...with the addition of the start_dt filter. Even if we keep it, it feels like a lot of extra duplicate code for that one additional filter.

backend/src/handlers/user_handler.go
```diff
@@ -17,14 +18,16 @@ func (srv *Server) registerUserRoutes() []routeDef {
 	return []routeDef{
 		{"GET /api/users", srv.handleIndexUsers, true, axx},
 		{"GET /api/users/{id}", srv.handleShowUser, false, axx},
+		{"GET /api/users/{id}/program-history", srv.handleGetUserProgramHistory, false, axx},
```
jtucholski, line 21: I wonder if `/programs` and `program-history` can do the same thing? Do they need to be separate?

backend/src/models/users.go
```diff
@@ -157,4 +157,21 @@ type ResidentProgramClassInfo struct {
...
+	CreditTypes       string      `json:"credit_types"`
+}
+
+type ResidentProgramClassWeeklySchedule struct {
+	ClassName   string `json:"class_name"`
+	StartDate   string `json:"start_date"`
+	EndDate     string `json:"end_date"`
+	CreditTypes string `json:"credit_types"`
+	Status      string `json:"status"`
 }
```
jtucholski, line 177: A weekly schedule should include
- a start date and end date which represent the window that the schedule is good for (e.g., 5/1/2025 through 5/31/2025 or 5/5/2025 through 5/9/2025).
- an array of events for that schedule which include the class name, the date, start time, and duration (or end time)


I don't believe it needs to include status or credit types.

frontend/src/common.ts
```diff
@@ -743,17 +743,33 @@ export interface Class {
...
+    absent_attendance: number;
+}
+
+export interface ResidentProgramClassHistory extends BaseResidentProgramClass {
+    program_name: string;
+    date_status_changed: string;
+}
+
+export interface ResidentProgramClassWeeklySchedule
+    extends BaseResidentProgramClass {
```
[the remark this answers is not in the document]
jtucholski: See my comment in `users.go` about the new `ResidentProgramClassWeeklySchedule`. I think the structure for this type needs to change to be more applicable to a calendar with a list of events on the calendar.

jtucholski (changes requested):

backend/src/database/users.go
```diff
@@ -542,23 +543,93 @@ func (db *DB) GetUserProgramInfo(args *models.QueryContext, userId int) ([]model
...
+		if err := base.Find(&userEnrollments).Error; err != nil {
+			return nil, NewDBError(err, "program_class_enrollments")
+		}
+
+	}
+	return userEnrollments, nil
+}
 
-	if err := base.Count(&args.Total).Error; err != nil {
+func (db *DB) GetUserProgramClassHistory(args *models.QueryContext, userId int) ([]models.ResidentProgramClassHistory, error) {
```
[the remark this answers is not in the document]
corypride: @jtucholski I tried to use this func to handle both cases, but if we are sure that these are the fields, grouping, and ordering that we want in these two structs  then research shows that keeping them separate may be the better choice because their purpose is slightly different.  

Although both functions query the program_class_enrollments table, and it's technically possible to modify GetUserProgramInfo to accommodate the different query logic, it would likely make the function more complex and harder to maintain.

By having a separate function, the code is more modular, and each function has a single responsibility. This makes it easier to understand, test, and maintain the codebase.

backend/src/handlers/user_handler.go
jtucholski: After looking at it further I still think we can get what we need from `/programs`. Happy to talk through my throught process.

frontend/src/Pages/ResidentOverview.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+export default function ResidentOverview() {
+    const { user } = useAuth();
+    const user_id = user?.id;
+    const {
+        data: programsResp
+        // error: programsError,
+        // isLoading
+    } = useSWR<ServerResponseMany<ResidentProgramClassInfo>, Error>(
+        `/api/users/${user_id}/programs?order=ASC&order_by=program_name`
```
jtucholski, line 19: If we implement it on the backend we can pass `&all=true`

backend/src/handlers/user_handler.go
```diff
@@ -17,14 +17,16 @@ func (srv *Server) registerUserRoutes() []routeDef {
...
+		{"GET /api/users/{id}/weekly-schedule", srv.handleGetUserWeeklySchedule, false, axx},
```
jtucholski, line 21: @corypride Can we look at the handler in `class_events.go`? 

`GET /api/student-calendar", srv.handleGetStudentCalendar` should do what we need. But I think Preston is pushing some changes it needs to support a date range.

backend/src/models/users.go
```diff
@@ -157,4 +157,33 @@ type ResidentProgramClassInfo struct {
...
+type Class struct {
+	Name      string     `json:"class_name"`
+	StartDate time.Time  `json:"start_date"`
+	EndDate   *time.Time `json:"end_date"`
+}
+
+type WeeklySchedule struct {
+	Day      string
+	TimeSlot string
+	Classes  []Class
```
jtucholski, line 188: @corypride We shouldn't need these if we use the existing schedule on class_events.go

backend/src/models/users.go
```diff
@@ -157,4 +157,33 @@ type ResidentProgramClassInfo struct {
...
+	CreditTypes          string                  `json:"credit_types"`
+}
+
+type ResidentProgramClassHistory struct {
+	ProgramName       string      `json:"program_name"`
+	ClassName         string      `json:"class_name"`
+	Status            ClassStatus `json:"status"`
+	DateStatusChanged string      `json:"date_status_changed"`
+	CreditTypes       string      `json:"credit_types"`
+}
```
jtucholski, line 169: @corypride We shouldn't need this if we use the one handler in `user_handler.go` for `/programs`

backend/src/database/users.go
```diff
@@ -537,23 +538,157 @@ func (db *DB) GetUserProgramInfo(args *models.QueryContext, userId int) ([]model
...
             pce.enrollment_status,
             p.name, p.id,
             pc.name, pc.status, pc.start_dt, pc.end_dt, pc.id
-        `).
-		Order("pc.start_dt DESC")
+        `)
+
+	if args.OrderBy != "" {
+		base = base.Order(args.OrderClause())
+	}
```
jtucholski, line 554: @corypride I don't understand what this does.

frontend/src/Pages/ResidentOverview.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const weekly_schedule_metrics = weekly_scheduleResp?.data;
+    console.log('weekly_schedule_metrics:>>   ', weekly_schedule_metrics);
+
+    const {
+        data: historyResp
+        // error: historyError,
+        // isLoading
+    } = useSWR<ServerResponseMany<ResidentProgramClassHistory>, Error>(
+        `/api/users/${user_id}/program-history`
+    );
```
jtucholski, line 39: @corypride We won't need this if we consolidate the call to one handler, though there will be some work on the front-end to split it between active classes and historical data. You could probably put it in a map and then render the two different parts on the page using that data.

frontend/src/Routes/Programs.tsx
```diff
@@ -20,6 +20,22 @@ import ProgramClassManagement from '@/Pages/ProgramClassManagement';
...
 import { Navigate } from 'react-router-dom';
+import ResidentOverview from '@/Pages/ResidentOverview';
+
+export const ProgramRoutes = DeclareAuthenticatedRoutes(
+    [
+        {
+            path: 'programs-residents',
+            element: <ResidentOverview />,
+            handle: {
+                title: 'Current Enrollments',
```
jtucholski, line 31: @corypride This has more than "Current Enrollments". Do you think we should call it "My Programs" or "Programs"?

frontend/src/common.ts
```diff
@@ -743,17 +743,37 @@ export interface Class {
...
     class_id: number;
     program_id: number;
+    present_attendance: number;
+    absent_attendance: number;
+}
+
+export interface ResidentProgramClassHistory extends BaseResidentProgramClass {
+    program_name: string;
+    date_status_changed: string;
+}
```
jtucholski, line 767: @corypride If we decide to consolidate the two handlers and use `/programs` then this or the Base type wouldn't be needed.

frontend/src/common.ts
```diff
@@ -743,17 +743,37 @@ export interface Class {
...
+
+export interface ResidentProgramClassWeeklySchedule {
+    className: string;
+    startDate: string;
+    endDate: string | null;
+    creditTypes: string;
+    status: string;
+    dateStatusChanged: string;
+    dayOfWeek: number; // 0 = Sunday, 6 = Saturday
 }
```
jtucholski, line 777: @corypride I don't think this matches the type definition in the backend. But what we need is probably found in `EventCalendar`, especially if we can use `/api/student-calendar`.

jtucholski (changes requested):

backend/src/handlers/user_handler.go
```diff
@@ -408,6 +411,7 @@ func (srv *Server) handleResidentTransfer(w http.ResponseWriter, r *http.Request
 	return writeJsonResponse(w, http.StatusOK, "successfully transferred resident")
 }
 
+// TODO: The access level of this handler was changed from true to false to allow for user routing. This is a temporary fix for the development of this feature and should be addressed
```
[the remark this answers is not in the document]
jtucholski: @corypride Do we need this comment anymore? Can we use `srv.canViewUserData(r)` to make sure that the requestor has access to the data?

frontend/src/common.ts
```diff
@@ -752,19 +752,55 @@ export interface Class {
...
+    day_index: number;
+    date: string;
+    events: CalendarEvent[];
 }
 
+export interface StudentCalendarResponse {
+    data: {
+        days: StudentCalendar[];
+    };
+}
```
jtucholski, line 792: @corypride I could be wrong but it feels like we don't use this pattern very often and instead we do things on the page we want the data with code like `enrollmentResp?.data`. Any opposition to taking that approach instead of mapping the response at the `data` level here?

frontend/src/common.ts
```diff
@@ -752,19 +752,55 @@ export interface Class {
...
+    date: string;
+    events: CalendarEvent[];
+}
+export function mapStudentCalendarResponseToDayData(
+    resp: StudentCalendarResponse | undefined
+): DayData[] {
+    // if resp is undefined or resp.data.days isn’t there, return empty array
+    const days = resp?.data?.days ?? [];
+    return days.map(({ date, events }) => ({ date, events }));
+}
```
jtucholski, line 803: @corypride Is this something we can move to the page or does it need to exist here?

jtucholski (changes requested):
@corypride I added a few more comments. Did not review the resident overview page or weekly schedule table since we are waiting on UI input from the designer. However I do want to take a look at it to validate the data for my resident is correct and I still have to do that.

frontend/src/Pages/ResidentOverview.tsx
```diff
@@ -0,0 +1,139 @@
...
+        `/api/student-calendar`
+    );
+
+    const weekly_schedule_metrics = scheduleResp?.data;
+    console.log('schedule_metrics:>>   ', weekly_schedule_metrics);
+    const { data: activityResp } = useSWR<
+        ServerResponseMany<ResidentProgramOverview>,
+        Error
+    >(
+        `/api/users/${user_id}/programs?view=activity&order=DESC&order_by=updated_at&all=false`
```
carddev81, line 31: `all` parameter should be set to `true` here? also, i noticed that you make to calls out to the database to the same url here just for ordering the data. I think you should make one call and then sort the data in the frontend here? Also, I would probably separate out the Historic programs/classes from the Current Programs/Classes. You could create a map with key value pairs,  for example the key being 'history' and value being the list of programs.

frontend/src/common.ts
```diff
@@ -752,19 +752,48 @@ export interface Class {
...
+export interface StudentCalendar {
+    day_index: number;
+    date: string;
+    events: CalendarEvent[];
 }
 
+export interface StudentCalendarResponse {
+    data: {
+        days: StudentCalendar[];
```
carddev81, line 790: you shouldn't need the data. Change it to the following, and then you will need to make sure within ResidentOverview to change useSWR to use ServerResponseOne.  
```
export interface StudentCalendarResponse {
    days: StudentCalendar[];
}
```

carddev81 (changes requested):

backend/src/database/class_events.go
```diff
@@ -325,33 +318,22 @@ func (db *DB) getCalendarFromEvents(events []models.ProgramClassEvent, month tim
...
 		return days[i].Date.Before(days[j].Date)
 	})
-	return models.NewCalendar(year, month.String(), days), nil
+	return models.NewCalendar(days), nil
 }
 
-func (db *DB) GetCalendar(month time.Month, year int, facilityId uint, userId *uint) (*models.Calendar, error) {
+func (db *DB) GetCalendar(dtRng *models.DateRange, facilityId uint, userId *uint) (*models.Calendar, error) {
 	content := []models.ProgramClassEvent{}
+	tx := db.Model(&models.ProgramClassEvent{}).
```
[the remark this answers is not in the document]
PThorpe92: pushing to the branch so we can get this through

PThorpe92 (approved):

PR #873, opened 2025-05-09
feat: add table prefix argument to OrderClause method on QueryContext

PThorpe92:
This PR fixes cases where the client does not add `order_by` & `order` to the query string. Previously, we would fall back to `created_at desc`, which is at least on 99% of our tables. However this is problematic for queries with multiple joins, that fallback becomes an ambiguous argument to the query and can fail.

This adds the `fallbackPrefix` argument to `queryContext.OrderClause()`, so the caller can specify _which_ table they would like it to reference.

e.g.
```sql
select users.*, e.* from users join enrollments e on u.id = e.user_id
```
or:
```go
tx := db.Model(&models.User{}).Select("users.*, e.*).Joins("enrollments e on e.user_id = users.id")
```

Now the caller can provide:
```go
tx = tx.Order(args.OrderClause("e")).Find(&result)
```

and it will default to using the enrollments `created_at` field if the query context doesn't contain an OrderBy.

### DIFF
--- a/backend/src/database/events_attendance.go
+++ b/backend/src/database/events_attendance.go
@@ -108,7 +108,7 @@ func (db *DB) GetEnrollmentsWithAttendanceForEvent(qryCtx *models.QueryContext, 
 	finalQuery := selectClause + baseQuery
 
 	if slices.Contains([]string{"name_first", "name_last"}, qryCtx.OrderBy) {
-		finalQuery += " ORDER BY " + qryCtx.OrderClause()
+		finalQuery += " ORDER BY " + qryCtx.OrderClause("e")
 	} else {
 		finalQuery += " ORDER BY e.id ASC"
 	}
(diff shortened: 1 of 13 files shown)


backend/src/database/events_attendance.go
jtucholski, line 1: @PThorpe92 Do we need to add this for `GetAttendees` too? I don't see that function call anywhere else.

jtucholski (approved):

backend/src/database/users.go
```diff
@@ -35,7 +35,7 @@ func (db *DB) GetCurrentUsers(args *models.QueryContext, role string) ([]models.
 		return nil, newGetRecordsDBError(err, "users")
 	}
 	users := make([]models.User, 0, args.PerPage)
-	if err := tx.Order(adjustUserOrderBy(args.OrderClause())).
+	if err := tx.Order(adjustUserOrderBy(args.OrderClause(""))).
```
corypride, line 38: Why not users here?

PR #874, opened 2025-05-13
fix: update url tampering on front and backend

CK-7vn:
## Description of the change
1. Makes sure that you can't update the date in the URL and get away with it. 
2. Make sure the date in the ClassEvents table matches the date in the url when you click on an event
3. Fix the backend GetRRule from truncating the final day. 

- **Related issues**: Asana id 186

## Screenshot(s)
image

## Additional context
The ticket only required that we make sure that users are unable to update attendance for future dates, but I noticed a couple more bugs while working on it. The date's weren't matching up because of a timezone issue with the date that the user clicks on in the attendance table it was a day ahead. Also, the backend was truncating the last day of class because it was stopping at 00:00, so I fixed how the rrule is gotten so that the 24 hours of the final day is included. 

Asana: 
https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210129036130407?focus=true

### DIFF
--- a/backend/src/handlers/class_events.go
+++ b/backend/src/handlers/class_events.go
@@ -94,6 +94,15 @@ func (srv *Server) handleGetProgramClassEvents(w http.ResponseWriter, r *http.Re
 	log.add("class id", classID)
 	month := r.URL.Query().Get("month")
 	year := r.URL.Query().Get("year")
+	justDates := r.URL.Query().Get("dates")
+	if justDates == "true" {
+		timezone := srv.getQueryContext(r).Timezone
+		dates, err := srv.Db.GetClassEventDatesForRecurrence(classID, timezone, month, year)
+		if err != nil {
+			return newDatabaseServiceError(err)
+		}
+		return writeJsonResponse(w, http.StatusOK, dates)
+	}
 
 	qryCtx := srv.getQueryContext(r)
 	instances, err := srv.Db.GetClassEventInstancesWithAttendanceForRecurrence(classID, &qryCtx, month, year)
--- a/frontend/src/Pages/EventAttendance.tsx
+++ b/frontend/src/Pages/EventAttendance.tsx
@@ -16,7 +16,7 @@ import Pagination from '@/Components/Pagination';
 import DropdownControl from '@/Components/inputs/DropdownControl';
 import Error from '@/Pages/Error';
 
-export function isFutureYyyyMmDd(dateStr: string): boolean | string {
+export function isFutureIsoDate(dateStr: string): boolean | string {
     const m = /^(\d{4})-(\d{2})-(\d{2})$/.exec(dateStr);
     if (!m) {
         return 'Date must be in YYYY‑MM‑DD format';
@@ -62,7 +62,7 @@ export default function EventAttendance() {
         class_id: string;
     }>();
     if (typeof date === 'string') {
-        const future = isFutureYyyyMmDd(date);
+        const future = isFutureIsoDate(date);
         if (typeof future === 'string') {
             return <Error back message={future} type="unauthorized" />;
         } else if (future) {
--- a/frontend/src/Pages/EventAttendance.tsx
+++ b/frontend/src/Pages/EventAttendance.tsx
@@ -70,7 +70,7 @@ export default function EventAttendance() {
                 <Error
                     back
                     type="unauthorized"
-                    message={`Attendance is unable to be recorded for ${date}`}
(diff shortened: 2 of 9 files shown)


frontend/src/Pages/EventAttendance.tsx
```diff
@@ -14,6 +14,34 @@ import SearchBar from '@/Components/inputs/SearchBar';
 import API from '@/api/api';
 import Pagination from '@/Components/Pagination';
 import DropdownControl from '@/Components/inputs/DropdownControl';
+import Error from '@/Pages/Error';
+
+export function isFutureYyyyMmDd(dateStr: string): boolean | string {
```
PThorpe92, line 19: I think maybe `isFutureDateString` or `isFutureIsoDate` + doc comment. That spelling is tripping me out :laughing:

CK-7vn: Just sound it out...ymmmyyyday :octocat: 

I got ya.

jtucholski:
@CK-7vn Still looking at the PR but one of the things that I noticed was that I am able to create a class that starts today, 05/14/2025 and meets on Wednesdays and record attendance for yesterday, 05/13/2025 (`http://127.0.0.1/program-classes/..../events/..../attendance/2025-05-13`).

While it technically is in the past, it is not a scheduled meeting date for the class. Can you make sure that: 
- the rules check for dates on after the class begins 
- that the date is one of the valid dates according to the RRule

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -33,6 +61,20 @@ export default function EventAttendance() {
...
+            return <Error back message={future} type="unauthorized" />;
+        } else if (future) {
+            return (
+                <Error
+                    back
+                    type="unauthorized"
+                    message={`Attendance is unable to be recorded for dates in the future`}
+                />
+            );
+        }
```
jtucholski, line 76: @CK-7vn Can we modify this logic so that the following behavior is achieved?

- Invalid date format (e.g., 25-05-14 or May-14-2025)
  - Redirect to /404 and.
- Valid date but in the past and not a scheduled time. 
  - Render the page with empty state and a message, "No class session was scheduled for this date."
- Valid date but in the future and not a scheduled time.
  - Same as above with one word different. "No class session is scheduled for this date."
- Valid future date on the schedule but hasn't happened.
  - Render the page with empty state and a message, "This session is scheduled for a future date. Attendance will be available after it occurs."

I hadn't thought about the Error component but it feels wrong to use for invalid input like not being an iso date. Maybe for the ones where we need to show a message we do something like what appears on lines 260 or 261?

jtucholski (changes requested):
Thanks for taking this first pass. I realize I could have added some more detail on error messages in the ticket. Still, this initial implementation helped me think it through a bit more. 

If you can make the requested changes I'll be good with the fix.

backend/src/handlers/class_events.go
```diff
@@ -94,6 +94,15 @@ func (srv *Server) handleGetProgramClassEvents(w http.ResponseWriter, r *http.Re
...
 	year := r.URL.Query().Get("year")
+	justDates := r.URL.Query().Get("dates")
+	if justDates == "true" {
+		timezone := srv.getQueryContext(r).Timezone
+		dates, err := srv.Db.GetClassEventDatesForRecurrence(classID, timezone, month, year)
+		if err != nil {
+			return newDatabaseServiceError(err)
+		}
+		return writeJsonResponse(w, http.StatusOK, dates)
+	}
```
jtucholski, line 105: @CK-7vn Does this or the calling function `GetClassEventDatesForRecurrance` take into consideration the overrides?

jtucholski (approved):

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -8,12 +8,20 @@ import {
...
 import API from '@/api/api';
 import Pagination from '@/Components/Pagination';
 import DropdownControl from '@/Components/inputs/DropdownControl';
+import Error from '@/Pages/Error';
+
+const isoRE = /^\d{4}-\d{2}-\d{2}$/;
+function parseLocalDay(isoDate: string): Date {
+    const [year, month, day] = isoDate.split('-').map(Number);
+    return new Date(year, month - 1, day);
+}
```
calisio, line 24: is this pretty general? Could it be moved to helper functions under formatting? so it can be reused across components

calisio (approved):
Besides the one small comment about moving the parse date to helper functions, looks good!

PR #876, opened 2025-05-13
Block Modifications on Cancelled or Completed Classes

corypride:
## Description of the change
Once a class’s status is set to Cancelled or Completed, users are prevented from enrolling new residents, recording attendance, or editing any core details.

- **Related issues**:
Assana: ID-200

## Screenshot(s)

---
  - https://app.asana.com/0/0/1210207869971837


backend/src/handlers/server.go
```diff
@@ -608,3 +608,22 @@ func (srv *Server) errorResponse(w http.ResponseWriter, status int, message stri
...
+	classId, err := strconv.Atoi(r.PathValue("class_id"))
+	if err != nil {
+		return newInvalidIdServiceError(err, "class ID")
+	}
+	class, err := srv.Db.GetClassByID(classId)
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	return srv.classStatusCheck(w, class)
+}
```
jtucholski, line 629: @corypride Is there a reason this was written as two separate functions? 

It looks like `validateClassStatus` is the only function calling `classStatusCheck`. Should they be consolidated and renamed to help the calling functions know that this checks to see if the class can be modified?

backend/src/handlers/server.go
```diff
@@ -608,3 +608,22 @@ func (srv *Server) errorResponse(w http.ResponseWriter, status int, message stri
...
+		return writeJsonResponse(w, http.StatusBadRequest, "Cannot perform update action on class that has been completed or cancelled")
+	}
+	return nil
+}
+
+func (srv *Server) validateClassStatus(w http.ResponseWriter, r *http.Request) error {
+	classId, err := strconv.Atoi(r.PathValue("class_id"))
+	if err != nil {
+		return newInvalidIdServiceError(err, "class ID")
+	}
```
jtucholski, line 623: @corypride From the looks of it, it looks like every function that calls this performs this exact same check prior to calling it.

frontend/src/Components/ClassLayout.tsx
```diff
@@ -21,6 +22,7 @@ import StatsCard from './StatsCard';
 
 function ClassInfoCard({ classInfo }: { classInfo?: Class }) {
     const navigate = useNavigate();
+    const thisSelectedClasStatus = classInfo?.status;
```
jtucholski, line 25: @corypride Is this variable necessary or can `classInfo?.status` be used wherever it's accessed?

frontend/src/Pages/ClassEnrollmentDetails.tsx
```diff
@@ -28,6 +29,8 @@ export default function ClassEnrollmentDetails() {
     const { class_id } = useParams<{ class_id: string }>();
     const navigate = useNavigate();
     const { redirect } = useLoaderData() as ClassLoaderData;
+    const { class: clsInfo } = useLoaderData() as ClassLoaderData;
+    const thisSelectedClasStatus = clsInfo?.status;
```
jtucholski, line 33: @corypride Is this necessary? I can see if you were to create this variable it could hold the expression itself (e.g., `const blockEdits = clsInfo?.status === SelectedClassStatus.Cancelled || clsInfo?.status === SelectedStatus.Completed`)

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -44,6 +45,7 @@ export default function ClassEvents() {
 
     const this_program = program_class?.data;
     const enrolled = this_program?.enrolled;
+    const thisSelectedClassStatus = this_program?.status;
```
jtucholski, line 48: Same comment as on other pages.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -58,7 +60,8 @@ export default function EventAttendance() {
     >(
         `/api/program-classes/${class_id}/events/${event_id}/attendance?date=${date}&page=${pageQuery}&per_page=${perPage}&search=${searchTerm}&order_by=${sortQuery}`
     );
-
+    const clsInfo = useLoaderData() as Class;
+    const thisSelectedClassStatus = clsInfo.status;
```
jtucholski, line 64: Same comment as on other pages for this variable.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -185,8 +188,16 @@ export default function EventAttendance() {
...
     }
-
     async function onSubmit() {
+        const canEditAttendance =
+            thisSelectedClassStatus !== SelectedClassStatus.Completed &&
+            thisSelectedClassStatus !== SelectedClassStatus.Cancelled;
+        if (!canEditAttendance) {
+            console.error(
+                'Cannot update attendance for completed or canceled classes'
+            );
```
jtucholski, line 198: @corypride We shouldn't be keeping console in here. This should be an error visible to the user.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -208,7 +219,13 @@ export default function EventAttendance() {
                 </div>
                 <button
                     onClick={() => void handleMarkAllPresent()}
-                    disabled={anyRowSelected}
+                    disabled={
+                        anyRowSelected ||
+                        (thisSelectedClassStatus !==
```
jtucholski, line 224: @corypride Good opportunity to use the `canEditAttendance` variable from above.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -359,6 +376,7 @@ export default function EventAttendance() {
                         <button
                             type="submit"
                             className="button"
+                            // disabled={!anyRowSelected || canEditAttendance}
```
jtucholski, line 379: We should remove this.

jtucholski (changes requested):
@corypride Can you make sure that the tooltips are also provided too?
- I should see the following tooltips: 
  - "This class is {state} and cannot be modified" when Edit Class Details is disabled.
  - "This class is {state}. No new residents can be enrolled" when Add Residents is disabled.
  - ""This class is {state}. No new attendance records can be recorded" when Mark Attendance is disabled.


I should have included this as a detail, but: 
- We only need to block **new** attendance records. We should still allow the user to view (but not edit or add) attendance records.
- If the user navigates to pages and tries to get around the disabled buttons (like  `/programs/{id}/classes/{class_id}`) when a class that is cancelled or completed, then they should see the error message as a Toast error, not a Submit Error (see attached). **I realize this is not how the page was built so if we need to, we can create a separate ticket for this. Just let me know all of the pages where this needs to be addressed so I can write the ticket accordingly. If you think you can make the change without significant effort though, that would be appreciated.**

image

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -208,8 +222,9 @@ export default function EventAttendance() {
                 </div>
                 <button
                     onClick={() => void handleMarkAllPresent()}
-                    disabled={anyRowSelected}
-                    className={`button ${anyRowSelected ? `bg-gray-400 cursor-not-allowed` : ``}`}
+                    disabled={anyRowSelected || !canEditAttendance}
+                    className={`button  tooltip tooltip-left ${anyRowSelected ? `bg-gray-400 cursor-not-allowed` : ``}`}
```
jtucholski, line 226: @corypride We should be using one of the `bg-grey-{1,2,3,4} variants and not `bg-gray-400`. You can find these in `tailwind.config.cjs`.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -355,11 +370,12 @@ export default function EventAttendance() {
...
                     </div>
-                    <div className="flex justify-end pt-4">
+                    <div className="flex justify-end pt-4 ">
                         <button
                             type="submit"
-                            className="button"
-                            disabled={!anyRowSelected}
+                            className="button tooltip tooltip-left"
+                            disabled={!anyRowSelected || !canEditAttendance}
+                            data-tip={`This class is ${clsInfo?.status} and cannot be modified.`}
```
jtucholski, line 378: @corypride Can you toLower this too?

jtucholski (changes requested):
It's functional and most of my recommendations have to do with final updates that affect either the readability of the code or the user experience.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -185,8 +193,14 @@ export default function EventAttendance() {
             return newMods;
         });
     }
-
     async function onSubmit() {
+        if (!canEditAttendance) {
+            toaster(
+                'Cannot update attendance for completed or canceled classes',
```
jtucholski, line 199: @corypride Our version of spelling is "cancelled" (I've been burned by it about 5 times)

frontend/src/Pages/ClassEnrollmentDetails.tsx
```diff
@@ -28,6 +29,10 @@ export default function ClassEnrollmentDetails() {
     const { class_id } = useParams<{ class_id: string }>();
     const navigate = useNavigate();
     const { redirect } = useLoaderData() as ClassLoaderData;
+    const { class: clsInfo } = useLoaderData() as ClassLoaderData;
+    const canAddResident =
+        clsInfo?.status !== SelectedClassStatus.Completed &&
+        clsInfo?.status !== SelectedClassStatus.Cancelled;
```
jtucholski, line 35: @corypride Do you want to use the same expression that you added in the other page to improve readability?

jtucholski (changes requested):
Almost there, there's a few outstanding comments to resolve.

frontend/src/Pages/ClassEnrollmentDetails.tsx
corypride: smh. I swear i changed this...  mustve came bk when handling merge conflicts!

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -58,14 +61,19 @@ export default function EventAttendance() {
     >(
         `/api/program-classes/${class_id}/events/${event_id}/attendance?date=${date}&page=${pageQuery}&per_page=${perPage}&search=${searchTerm}&order_by=${sortQuery}`
     );
+    const rawClsInfo = useLoaderData() as { class?: Class };
+    const clsInfo = rawClsInfo?.class;
```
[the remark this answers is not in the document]
corypride: @jtucholski the extra step was needed because the data returned from the api call included a title property alongside the class object that looked something like this
 `{
  title: 'Test Class Completed',
  class: { ... }  // <--- class data was here
}`

it returned that despite being typed as "Class"  like in this case: 
`const clsInfo = useLoaderData() as Class;`

so the extra step was to extract the class from that

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -44,6 +45,10 @@ export default function ClassEvents() {
 
     const this_program = program_class?.data;
     const enrolled = this_program?.enrolled;
+
+    const blockEdits =
+        this_program?.status === SelectedClassStatus.Completed ||
+        this_program?.status == SelectedClassStatus.Cancelled;
```
calisio, line 51: Would we be able to extract this check into a function that can be reused across pages? It might cut down on repetitiveness throughout the pages, and in case we want to modify this in the future only have to change it in one place rather than all.

jtucholski:
Ready for second review.

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -208,8 +222,9 @@ export default function EventAttendance() {
                 </div>
                 <button
                     onClick={() => void handleMarkAllPresent()}
-                    disabled={anyRowSelected}
-                    className={`button ${anyRowSelected ? `bg-gray-400 cursor-not-allowed` : ``}`}
+                    disabled={anyRowSelected || blockEdits}
+                    className={`button  tooltip tooltip-left ${anyRowSelected ? `bg-grey-4 cursor-not-allowed` : ``}`}
+                    data-tip={`This class is ${clsInfo?.status.toLowerCase()} and cannot be modified.`}
```
calisio, line 227: This is just one example, but it seems like you added the data tip on every button, not only when they are disabled. This is a scheduled section and the tooltips still appear:

Screenshot 2025-05-19 at 12 35 30 PM

This happens for all occasions of the buttons (on the edit and add resident buttons too)

go.work.sum
calisio, line 1: I don't believe there should be any changes to this file?

calisio (changes requested):
I left a few comments on this PR. I also wanted to note that in the ticket it mentions "I want to view but not modify classes that are cancelled, completed, or archived". This PR only deals with cancelled and completed classes. @jtucholski Is there clarification if archived classes should also be included in these blocked actions? Currently they are not.

jtucholski:
@calisio @corypride We should not allow for changes to occur on archived classes.

calisio (changes requested):
Updates you made look good. 
There are a few more things I found on review a second time:
1. when i try to update a class that is scheduled, it shows the error toast for some reason. see video below

https://github.com/user-attachments/assets/b50e7d37-a04d-4838-a407-3c089fc06b6f

2. the tooltip still appears when i try to go edit a class attendance I am allowed to modify
Screenshot 2025-05-19 at 3 34 33 PM

I think those are the only two outstanding things!

calisio (changes requested):
Screenshot 2025-05-22 at 12 23 15 PM
still getting the tooltip - you'll have to check it at the class name also.

PR #877, opened 2025-05-16
fix: eliminate possibility for facility admin not of same facility to…

CK-7vn:
## Description of the change
This pull request provides a solution to the issue of a facility admin being able to update a class for a facility they have no admin privileges in. 

- **Related issues**: Asana ticket 184 
- Asana: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210129030386867?focus=true

## Additional context
This is a smaller solution to the ticket, this is the same essential logic that the other pull request utilizes, but is only implemented here.

### DIFF
--- a/backend/src/database/facilities.go
+++ b/backend/src/database/facilities.go
@@ -56,3 +56,9 @@ func (db *DB) DeleteFacility(id int) error {
 	}
 	return nil
 }
+
+func (db *DB) GetResourceFacilityID(tableName any, resourceID any) (uint, error) {
+	var facilityID uint
+	err := db.Model(&tableName).Select("facility_id").Where("id = ?", resourceID).Scan(&facilityID).Error
+	return facilityID, err
+}
--- a/backend/src/handlers/auth.go
+++ b/backend/src/handlers/auth.go
@@ -158,6 +158,20 @@ func (srv *Server) adminMiddleware(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+func (srv *Server) UserCanEditResource(r *http.Request, resourceType any, resourceID any) (bool, error) {
+	claims := r.Context().Value(ClaimsKey).(*Claims)
+	if claims.canSwitchFacility() {
+		return true, nil
+	}
+	resourceFacilityID, err := srv.Db.GetResourceFacilityID(resourceType, resourceID)
+	if err != nil {
+		return false, err
+	}
+	if resourceFacilityID != claims.FacilityID {
+		return false, nil
+	}
+	return true, nil
+}
 
 // Auth endpoint that is called from the client before each <AuthenticatedLayout /> is rendered
 func (srv *Server) handleCheckAuth(w http.ResponseWriter, r *http.Request, log sLog) error {
--- a/backend/src/handlers/classes_handler.go
+++ b/backend/src/handlers/classes_handler.go
@@ -97,6 +97,13 @@ func (srv *Server) handleUpdateClass(w http.ResponseWriter, r *http.Request, log
 	if err := json.NewDecoder(r.Body).Decode(&class); err != nil {
 		return newJSONReqBodyServiceError(err)
 	}
+	canUpdate, err := srv.UserCanEditResource(r, models.ProgramClass{}, id)
+	if err != nil {
+		return newInternalServerServiceError(err, "facility check to edit resource")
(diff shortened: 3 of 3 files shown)


backend/src/handlers/auth.go
```diff
@@ -158,6 +158,20 @@ func (srv *Server) adminMiddleware(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+func (srv *Server) UserCanEditResource(r *http.Request, resourceType any, resourceID any) (bool, error) {
```
jtucholski, line 161: @CK-7vn -- This is similar to `canViewUserData` with the exception that the other function includes a check to see if the resource has the same ID as the claims. Is it possible that these are redundant?

Another thought. Do you think we should consider an access type, like "view" or "edit"

CK-7vn: I don't think they're redundant, at least personally, I think we need a broader way to do this, in particular when it comes to making sure that the user trying to edit data is an admin but also in the same facility. But, that's just my 2 cents. The canViewUserData doesn't include the facility check either.

jtucholski (approved):

carddev81:
@CK-7vn I'm unable to create a class following these steps below.  I cannot get past step 3 here.  I put a video here too and a screen shot

Steps to Reproduce

1. As a Department Admin: Create two Facilities, A and B
2. As a Department Admin: Create one Program, offered only at Facility A. (/programs/1)
3. As a Department Admin: Create a class for Facility A. (/program-classes/1/dashboard)
4. As Facility Admin (B): Log in and access the route for the the class offered in Facility A (/program-classes/1/dashboard)
     a. Enrolling new residents is blocked.
     b. Editing the class and updating enrollment status is not blocked.

[DeptAdminCantAddClass.webm](https://github.com/user-attachments/assets/2429e6cb-1c74-4347-84bb-ecb62d869276)
image

CK-7vn:
> @CK-7vn I'm unable to create a class following these steps below. I cannot get past step 3 here. I put a video here too and a screen shot
> 
> Steps to Reproduce
> 
> 1. As a Department Admin: Create two Facilities, A and B
> 2. As a Department Admin: Create one Program, offered only at Facility A. (/programs/1)
> 3. As a Department Admin: Create a class for Facility A. (/program-classes/1/dashboard)
> 4. As Facility Admin (B): Log in and access the route for the the class offered in Facility A (/program-classes/1/dashboard)
>    a. Enrolling new residents is blocked.
>    b. Editing the class and updating enrollment status is not blocked.
> 
> [DeptAdminCantAddClass.webm](https://github.com/user-attachments/assets/2429e6cb-1c74-4347-84bb-ecb62d869276) image

I will look into this when I get in in the morning.

backend/src/handlers/classes_handler.go
```diff
@@ -120,6 +127,13 @@ func (srv *Server) handleUpdateClasses(w http.ResponseWriter, r *http.Request, l
 		if classID, err := strconv.Atoi(id); err == nil {
 			classIDs = append(classIDs, classID)
 		}
+		canUpdate, err := srv.UserCanEditResource(r, models.ProgramClass{}, classIDs)
```
carddev81, line 130: You can move this database call outside of the loop? For every iteration it not only makes another call to the database but also makes another call with the same id(s), plus an additional id.

backend/src/database/facilities.go
```diff
@@ -56,3 +56,9 @@ func (db *DB) DeleteFacility(id int) error {
 	}
 	return nil
 }
+
+func (db *DB) GetResourceFacilityID(tableName any, resourceID any) (uint, error) {
+	var facilityID uint
+	err := db.Model(&tableName).Select("facility_id").Where("id = ?", resourceID).Scan(&facilityID).Error
```
carddev81, line 62: Change `= ?` here to IN(?). This is for when you have multiple IDs being passed into the SQL query. If you leave as it, it will fail to execute for multiple IDs.

carddev81 (changes requested):

carddev81 (changes requested):
just a couple of issues found

carddev81:
@CK-7vn Don't worry about fixing the issue where I sent you the steps to reproduce, screenshot, and video, this was occurring on staging too and Preston fixed it in main. So once yours get merged it will work.

PR #882, opened 2025-05-19
feat: add functionality to allow administrator to cancel a scheduled event per ticket 197

carddev81:
## Description of the change

- Added the functionality for allowing an administrator to cancel a scheduled event. 
    - Modified/Added logic within the backend API
    - Modified/Added logic within the frontend API
- Modified history logic to be consistent with how Programs is being created displayed.
    - Modified/Added/Removed logic within the backend API

NOTES
- For development/testing purposes I have "Temporarily" added the Cancel functionality to the View Calendar button on the Class Edit screen (ClassManagementForm.tsx).  See Screen shot below.
- The changes for this PR are located within this commit: [feat: add functionality to allow administrator to cancel a scheduled event](https://github.com/UnlockedLabs/UnlockEdv2/commit/b0ff448939be79b703a2d39a7038fab073ca14ab)

- **Related issues**: Link to Asana ticket [Cancel Event on Facility Schedule](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210197345470684?focus=true).

## Screenshot(s)
image


backend/src/database/DB.go
```diff
@@ -135,6 +137,11 @@ func MigrateTesting(db *gorm.DB) {
 			logrus.Fatal("Failed to migrate table: ", err)
 		}
 	}
+	if !db.Migrator().HasColumn(&models.FacilitiesPrograms{}, "deleted_at") {
+		if err := db.Migrator().AddColumn(&models.FacilitiesPrograms{}, "DeletedAt"); err != nil {
+			logrus.Fatal("Failed to add deleted_at column: ", err)
+		}
+	}
```
jtucholski, line 144: @carddev81 Was there not a deleted_at column already in facilities_programs?

backend/src/models/class_event.go
```diff
@@ -88,15 +88,30 @@ type ProgramClassEventOverride struct {
...
+func (pce *ProgramClassEventOverride) GetFormattedCancelledDate() (*string, error) {
+	if !pce.IsCancelled {
+		return StringPtr(""), nil
+	}
+	rRule, err := rrule.StrToRRule(pce.OverrideRrule)
+	if err != nil {
+		return nil, err
+	}
+	cancelledDate := rRule.All()[0]
+	cancelledDateStr := cancelledDate.Format("1/02/2006")
```
jtucholski, line 111: @carddev81 Something about this sticks out to me as UI formatting, but I can't think of a better solution right now given the challenge of storing everything as a string...that and also `event_cancelled` seems to be a very special type of logged item.

carddev81: I could save the RRule into the new value column and then I could do the formatting in the front end?

frontend/src/Components/EventCalendar.tsx
```diff
@@ -47,27 +67,70 @@ export default function EventCalendar({
...
+                isCancelled: override?.is_cancelled,
+                reason: override?.reason,
+                location: override?.location
+            };
+        });
         setEvents(generated);
-    }, [recurrenceRule, durationStr, title]);
+    }, [recurrenceRule, durationStr, title, classEvent?.overrides]);
+
+    function getEventOverrideForEventDate(
```
jtucholski, line 93: @carddev81 Does this need to handle rescheduled events? I know the ticket was only for cancelled, but wasn't sure how moving a single occurrence would affect this logic. If it's only cancelled, then maybe we call it `findCancellationForEventDate`?

backend/src/models/class_event.go
jtucholski: Hmm..yeah that works because then the front-end can format it into a single event date value.

jtucholski (approved):

jtucholski (commented):
I reviewed the changes associated with the commit for Cancelling Events. Looks good

backend/src/models/class_event.go
carddev81: @jtucholski pushed changes for this and also I pushed changes to account for making sure to delete attendance tied to a cancelled event.

frontend/src/Components/EventCalendar.tsx
carddev81: @jtucholski I remove this function in the reschedule code. I'll make that change here to be consistent with naming.  Change has been pushed.

backend/src/handlers/events_attendance_handler.go
```diff
@@ -77,13 +79,41 @@ func (srv *Server) handleGetEventAttendance(w http.ResponseWriter, r *http.Reque
...
+		}
+		overrideEvent := rRule.All()[0]
+		overrideFmtDate := overrideEvent.Format("2006-01-02")
+		if overrideFmtDate == attDate.Format("2006-01-02") {
+			return true
+		}
+	}
+	return false
+}
+
```
calisio, line 116: if im following, this is the event date, right? I wonder if it would be better to be more general with this param (ie use "eventDate" instead) so we can reuse it in other places? I am just thinking about the PR i just put in, where in the future we will have to check if the event is cancelled, and (I think?) we could use this function in there too. Thoughts? Also let me know if I misinterpreted

frontend/src/Pages/EventAttendance.tsx
```diff
@@ -216,7 +216,14 @@ export default function EventAttendance() {
...
             {isLoading && <div>Loading...</div>}
-            {error && <div className="error">Error loading data</div>}
+
+            {error && error.message === 'Conflict' ? (
+                <div className="text-error">
+                    Unable to mark attendance for a cancelled event
+                </div>
+            ) : (
+                error && <div className="text-error">Error loading data</div>
+            )}
```
calisio, line 226: This implementation is different than what Cory has implemented on his PR currently (it gives a toast error message if you try to submit the attendance for a day that is cancelled). We should probably pick one path or the other? Not sure what @jtucholski input is on this.

calisio (commented):
We'll have a good amount of conflicts with this PR and my PR (because I change the way we pass in things into EventCalendar) so we will have to chat to see what is the best way to resolve those.
A couple follow up questions but looks great and tests good ! I strictly looked at the PR you tagged in the description.

backend/src/handlers/events_attendance_handler.go
carddev81: Yes, you are following this right, this is the event date.  I renamed it to be more general--eventDate. Yes, I'm sure you are right about being dependent on this method in the future.

frontend/src/Pages/EventAttendance.tsx
carddev81: This message was intended to be displayed upon load of the page when a user navigates to it rather than display the Error loading data. It didn't seem consistent to display the toast on the loading of the page?

carddev81:
> We'll have a good amount of conflicts with this PR and my PR (because I change the way we pass in things into EventCalendar) so we will have to chat to see what is the best way to resolve those. A couple follow up questions but looks great and tests good ! I strictly looked at the PR you tagged in the description.

I was hoping you could just overwrite the entire event calendar with your work since my work was only added for testing the functionality of the cancel/reschedule of the event.  I could go back and wire in the components after yours gets merged?

frontend/src/Pages/EventAttendance.tsx
calisio: Right, I think that blocking the action is likely a better path, but the fix that is in place right now allows you to click the "Submit" button, and then tells you it couldnt save due to the class being cancelled. I was just thinking we choose one path or another-- either we implement it similar to how it is done right now, or we fix what is currently in the codebase to redirect to the 404 like how you have it here.
Now that I think about it, this fix might actually do that already? Lets look at it again once you rebase main into here

jtucholski: @calisio Agree. If we know that there is going to be an error, then we block the action with an error/validation message and/or then a modal. 

Using a toast is not a preferred way to block something because the user has no way to confirm they saw the message (and it disappears off-screen on a timeout).

I have a different ticket in the system to deal with this behavior elsewhere (trying to archive a program that has scheduled/active classes in a facility). I'll push forward with that ticket and look for other areas where we do this and create ones to address those too.

carddev81: sounds good, waiting on #881 to be merged to main before rebasing.

backend/src/database/class_events.go
```diff
@@ -519,6 +561,23 @@ func (db *DB) GetClassEventInstancesWithAttendanceForRecurrence(classId int, qry
...
+		for _, override := range event.Overrides { //for now just checking for cancellations
+			if !override.IsCancelled { //don't bother
+				continue
+			}
+			rRule, err := rrule.StrToRRule(override.OverrideRrule)
+			if err != nil {
+				logrus.Warnf("unable to parse cancelled override rule: %s", override.OverrideRrule)
+				continue
+			}
+			overrideInstance := rRule.All()[0]
```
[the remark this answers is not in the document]
carddev81: I'll make the change after my meet with josh, I didn't add that check only because a canceled override is only one event.

backend/src/handlers/events_attendance_handler.go
```diff
@@ -96,13 +98,41 @@ func (srv *Server) handleGetEventAttendance(w http.ResponseWriter, r *http.Reque
...
+	attDate, err := time.Parse("2006-01-02", eventDate) ///making sure the date is good
+	if err != nil {
+		return false
+	}
+	for _, override := range overrides {
+		rRule, err := rrule.StrToRRule(override.OverrideRrule)
+		if err != nil {
+			continue
+		}
+		overrideEvent := rRule.All()[0]
```
PThorpe92, line 127: Again here*

backend/src/database/class_events.go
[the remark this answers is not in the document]
carddev81: @PThorpe92 change/pushed

backend/src/models/class_event.go
```diff
@@ -99,15 +99,31 @@ type ProgramClassEventOverride struct {
...
+// format argument will take a string in the format of "2006-01-02", "1/02/2006", ect
+func (pce *ProgramClassEventOverride) GetFormattedCancelledDate(format string) (*string, error) {
+	if !pce.IsCancelled {
+		return StringPtr(""), nil
+	}
+	rRule, err := rrule.StrToRRule(pce.OverrideRrule)
+	if err != nil {
+		return nil, err
+	}
+	cancelledDate := rRule.All()[0]
```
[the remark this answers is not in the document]
carddev81: @PThorpe92 change/pushed...i'm going to work on fixing conflicts and making updates to account for Carolina's changes to the event calendar.

frontend/src/Pages/Schedule.tsx
```diff
@@ -1,11 +1,25 @@
...
+import { CancelClassEventModal } from '@/Components/modals/CancelClassEventModal';
+import { useMemo, useRef, useState } from 'react';
+import useSWR from 'swr';
 
 export default function Schedule() {
-    const events = useLoaderData() as FacilityProgramClassEvent[];
+    const cancelClassEventModal = useRef<HTMLDialogElement>(null);
+    const { startDate, endDate } = useMemo(() => {
+        const start = new Date(Date.now() - 1000 * 60 * 60 * 24 * 30 * 3);
+        const end = new Date(Date.now() + 1000 * 60 * 60 * 24 * 365);
```
calisio, line 13: can we keep the comments here saying it is 3 months ago and a year from now? Might be confusing to someone coming in and looking at this randomly.

frontend/src/Pages/Schedule.tsx
```diff
@@ -88,7 +102,12 @@ export default function Schedule() {
                             </button>
                             <CancelButton
                                 onClick={() => {
-                                    console.log('cancelling event');
+                                    if (!selectedEvent) return;
+                                    if (selectedEvent.is_cancelled) {
+                                        //do nothing
+                                        return;
+                                    }
```
calisio, line 109: can we condense this
`if (!selectedEvent || selectedEvent.is_cancelled) return;`

frontend/src/Pages/Schedule.tsx
```diff
@@ -100,6 +119,15 @@ export default function Schedule() {
...
             </div>
+            {selectedEvent ? (
+                <CancelClassEventModal
+                    mutate={mutateEvents}
+                    calendarEvent={selectedEvent}
+                    ref={cancelClassEventModal}
+                />
+            ) : (
+                ''
+            )}
```
calisio, line 130: doesnt have to be a ternary, i think it just can be 
```
            {selectedEvent && (
                <CancelClassEventModal
                    mutate={mutateEvents}
                    calendarEvent={selectedEvent}
                    ref={cancelClassEventModal}
                />
            )}
```

calisio (approved):
super small comments but besides that condensing it looks good!

PR #884, opened 2025-05-20
Update program outcomes chart

corypride:
## Description of the change
Replaced static data built into the Program Outcomes chart for the Program Overview Dashboard

image


backend/src/database/program_classes.go
```diff
@@ -87,6 +87,32 @@ func (db *DB) GetProgramClassDetailsByID(id int, args *models.QueryContext) ([]m
...
+	var outcome []ProgramClassOutcome
+
+	query := db.WithContext(args.Ctx).
+		Table("(SELECT TO_CHAR(DATE_TRUNC('month', NOW()) - INTERVAL '1 month' * gs.i, 'YYYY-MM') AS month FROM generate_series(0, 5) AS gs(i)) AS months").
+		Select(`
+        months.month,
+        COALESCE(COUNT(CASE WHEN pce.enrollment_status = 'Completed' THEN 1 END), 0) AS completions,
+        COALESCE(COUNT(CASE WHEN pce.enrollment_status IN ('Incomplete: Dropped', 'Incomplete: Failed to Complete', 'Incomplete: Transferred', 'Incomplete: Withdrawn') THEN 1 END), 0) AS dropouts
+    `).
+		Joins(`LEFT JOIN program_class_enrollments pce ON TO_CHAR(DATE_TRUNC('month', pce.updated_at), 'YYYY-MM') = months.month AND pce.class_id = ?`, id).
```
jtucholski, line 106: @corypride Did you mention to filter by class_id or program id (`id`)? I'm not getting the right data back because the program id is passed in for the class id filter. 

Additionally, we should take into consideration the user's selected facility so that we show them outcomes depending on the facility-program combination.

Can you break out the `WHERE` statement into it's own clause so that it improves the readability?

jtucholski (changes requested):
**Test Cases I tried**
✅ Chart only shows the last six months (excluding current month)
✅ Chart data doesn't include data past six months
❌ Data lines up with the correct months
❌ "Dropouts" renamed to Drops
✅ Completions and drops shown separately (e.g., stacked bars)
❌ Does not include data from other programs or unrelated classes
❌ Data does not display when there are 2 or more classes in a program
❌ Data is correct depending on the selected facility



**Details**
❌ Data is displayed for the correct month
  - What the API returns looks correct. What the chart shows is off by 1 month.

❌ Does not include data from other programs or unrelated classes
  - Looks like the logic used for filtering is wrong (`class_id = program_id`)

❌ Data is correct depending on the selected facility
  - Query should limit outcomes visual to classes offered at the user's facility. Right now it's set up to filter by program id.

❌ Data does not display when there are 2 or more classes in a program
  - Couldn't test this.

backend/src/database/program_classes.go
```diff
@@ -87,6 +88,54 @@ func (db *DB) GetProgramClassDetailsByID(id int, args *models.QueryContext) ([]m
...
+			) THEN pce.class_id END), 0) AS drops
+		`).
+		Joins(fmt.Sprintf(`
+			LEFT JOIN (%s) AS pce 
+			ON TO_CHAR(DATE_TRUNC('month', pce.updated_at), 'YYYY-MM') = months.month
+		`, enrollmentsSubquery), id, facilityID).
+		Group("months.month").
+		Order(args.OrderBy)
+
+	if err := query.Debug().Find(&outcome).Error; err != nil {
```
jtucholski, line 133: @corypride Do we normally leave `Debug()` statements in?

jtucholski (changes requested):
I hadn't tested this because I noticed an issue with `GetProgramClassOutcomes` that it was including `Completed` as one of the incomplete statuses. Can you resolve that, test, and review some of the other comments to clean up the readability of the function?

**Edit:** Can you also resolve conflicts before submitting changes?

jtucholski (approved):
Tested
✅ Chart only shows the last six months (excluding current month)
✅ Chart data doesn't include data past six months
✅ Data lines up with the correct months
✅ "Dropouts" renamed to Drops
✅ Completions and drops shown separately (e.g., stacked bars)
✅ Does not include data from other programs or unrelated classes
✅ Data does not display when there are 2 or more classes in a program
✅ Data is correct depending on the selected facility

@corypride I noticed that `Incomplete: Withdrawn` was not included in the list. I realize that was missing from the original description and wasn't added once [we received a comment](https://app.asana.com/1/1201607307149189/task/1210218688496286/comment/1210250548724628?focus=true) from Lisa to include it.

PR #888, opened 2025-05-22
feat: add facility schedule to admin view

calisio:
## Description of the change

- [x]  Facility Schedule page added to side nav (below Programs, above Residents).
- [x]  Calendar component renders with Week view by default.
- [x]  Calendar starts at current hour or 6:00 AM.
- [x]  Month(s) visible at top; supports weeks that span two months.
- [x]  Selected event is visually highlighted.
- [x]  Shows all events for the current facility in visible date range.
    - note: i have defaulted the range of events pulled to be from 3 months prior to today, to 1 year out in advance. Let me know if this is preferred to change.
- [x]  Overlapping events display correctly (calendar default behavior).
- [x]  Clicking an event shows:
    - Program & Class Name
    - Instructor
    - Room
    - Start/End time
    - Frequency
    - Enrolled residents
- [x]  Events fetched via API, filtered by facility and date range.
- [x]  Timezone consistency handled.
- [x]  Switching views updates calendar correctly.

Other Notes:
- **Follow up ticket:** Cancelled events styled differently (e.g., strikethrough), but grabbing the overrides is being left for a follow up ticket. Styling is completed for this.
- **Follow up ticket:** There will be another follow up ticket to condense events in both the backend and the frontend. I have implemented separate functions/structs on purpose (although may be slightly redundant) so we can handle for this in that ticket rather than make this ticket larger.
- Updates date range parser to accept multiple formats of dates. Left it where it is because mainly relevant in this code block.
- A few UI fixes to the calendar colors to make it easier to see/read.

- **Related issues**: [Asana ticket 196](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210197344953656?focus=true).

## Screenshot(s)
Screenshot 2025-05-22 at 12 04 21 PM
Screenshot 2025-05-22 at 12 04 53 PM

### DIFF
--- a/backend/src/database/class_events.go
+++ b/backend/src/database/class_events.go
@@ -323,17 +323,17 @@ func (db *DB) getCalendarFromEvents(events []models.ProgramClassEvent, rng *mode
 
 func (db *DB) GetFacilityCalendar(args *models.QueryContext, dtRng *models.DateRange) ([]models.FacilityProgramClassEvent, error) {
 	events := []models.FacilityProgramClassEvent{}
+	// TOP DO: finish adding overrides as is_cancelled (so it renders in the frontend)
 	tx := db.WithContext(args.Ctx).Table("program_class_events pcev").
 		Select(`pcev.*,
 		c.instructor_name,
 		c.name as class_name,
-		CASE WHEN c.status = 'Cancelled' THEN TRUE ELSE FALSE END as is_cancelled,
-		ARRAY_AGG(DISTINCT CONCAT(u.id, ':', u.name_first, ' ', u.name_last)) FILTER (WHERE e.enrollment_status = 'Enrolled') AS enrolled_users`).
+        STRING_AGG(CONCAT(u.id, ':', u.name_last, ', ', u.name_first), '|' ORDER BY u.name_last) FILTER (WHERE e.enrollment_status = 'Enrolled') AS enrolled_users`).
 		Joins("JOIN program_classes c ON c.id = pcev.class_id").
 		Joins("LEFT JOIN program_class_enrollments e ON e.class_id = c.id AND e.enrollment_status = 'Enrolled'").
 		Joins("LEFT JOIN users u ON e.user_id = u.id").
 		Where("c.facility_id = ?", args.FacilityID).
-		Group("pcev.id, c.instructor_name, c.name, c.status")
+		Group("pcev.id, c.instructor_name, c.name")
 	if err := tx.Scan(&events).Error; err != nil {
 		return nil, newGetRecordsDBError(err, "program_class_events")
 	}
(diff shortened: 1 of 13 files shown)


backend/src/database/class_events.go
```diff
@@ -321,6 +321,52 @@ func (db *DB) getCalendarFromEvents(events []models.ProgramClassEvent, rng *mode
...
+				ProgramName:       event.ProgramName,
+				ClassName:         event.ClassName,
+				EnrolledUsers:     event.EnrolledUsers,
+				StartTime:         occurrence,
+				EndTime:           occurrence.Add(duration),
+				Frequency:         rRule.OrigOptions.Freq.String(),
+			}
+			facilityEvents = append(facilityEvents, facilityEvent)
+		}
+	}
```
calisio, line 366: While I know a nested `for` loop is not the best thing in terms of performance, I am not sure if there was a way to get around it in this case. I did all the calculations outside so it wouldn't repeat them, but actually building the events I think is inevitable here.

PThorpe92: I haven't looked too deeply yet but I think it should be fine because it's only ever going to be called for a limited date range :+1:

backend/src/database/class_events.go
```diff
@@ -321,6 +321,52 @@ func (db *DB) getCalendarFromEvents(events []models.ProgramClassEvent, rng *mode
 	return models.NewCalendar(days), nil
 }
 
+func (db *DB) GetFacilityCalendar(args *models.QueryContext, dtRng *models.DateRange) ([]models.FacilityProgramClassEvent, error) {
+	events := []models.FacilityProgramClassEvent{}
```
PThorpe92, line 325: nit: can we preallocate size of 10 here?

backend/src/models/class_event.go
```diff
@@ -164,3 +164,15 @@ type EventDates struct {
...
 }
+
+type FacilityProgramClassEvent struct {
+	ProgramClassEvent
+	InstructorName string    `json:"instructor_name"`
+	ProgramName    string    `json:"program_name"`
+	ClassName      string    `json:"title"`
+	IsCancelled    bool      `json:"is_cancelled"`
+	EnrolledUsers  string    `json:"enrolled_users"`
+	StartTime      time.Time `json:"start"`
```
PThorpe92, line 175: we prob need to start making sure that we are only using `*time.Time` only to avoid Go's lovely `01:01:0001:01:01` default for that type.

carddev81 (approved):
Great job!! Tested good, and looks incredible.  

TESTED Ones unchecked are unable to be tested because that logic is not within this PR. It will be included in a future enhancement.

- [x] Calendar defaults to the current week view.
- [x] Calendar first time slot either defaults to the current hour or 6:00 AM.
- [x] Month properly shown at the top of the calendar.
- [x] For weeks that span two months, both months are visible.
- [x] All events in the visible date range are shown.
- [ ] Cancelled events still appear on the calendar with differentiated styling (either strikethrough or 
- [x] Selecting an event displays the details for that event next to the calendar.
- [x] The selected evet is visibly marked in some way.
- [ ] Visible events include both one-time (overrides) and recurring (RRule) events.

The only thing I found was when I switched to dark mode I had a hard time reading the `+#` link.  see pic:

 
image

calisio:
@PThorpe92 those small fixes are pushed
@carddev81 the UI is updated for those events!

PThorpe92:
@calisio I pushed a fix to use `make([]models.ProgramClassEvent, 0, 10)` instead of  `make([]models.ProgramClassEvent, 10)` 

no biggie but just noting for future reference cuz it will add 10 default structs. other than that.. looks good, all set :+1:

PThorpe92 (approved):

PR #896, opened 2025-06-02
fix: use nullable int64 for counting login activity and use proper http code

PThorpe92:
On a new instance with no users, the Operational Insights page will error due to an attempt to scan null into an int64. we should return 0 in this case.

Also don't use 404 for a resource that doesn't return any rows, because this is always an `/api/resource/{id}` resource, we should return bad request because the ID specified doesn't exist


PThorpe92:
> @PThorpe92 It isn't clear to me that `ErrRecordNotFound` is being used in the way I would have expected. I would have thought it was because the record with a specific id or set of values could not be found. If my assumption is correct I would expect a 404 since a 400 is usually the result of a malformed request object or one that doesn't pass validation right, right?
> 
> But if the usage if this is not what I expect then I think the above could be ignored.

An HTTP 404 means "resource not found", meaning if you requested `/api/users/44`, that would signal that there is no endpoint at all with the uri of `/api/users/*`. If the user requests a specific resource by an ID `/api/users/44` and `44` is not the ID of a user, than it is a bad request, because the resource exists for you to retrieve users at that endpoint, you just put the wrong ID in the URI, if that makes sense

jtucholski (approved):

PR #897, opened 2025-06-02
fix: lock users out after 5 failed login attempts within 15 mins

CK-7vn:
## Description of the change
This pull request locks a user "out" if they have failed to login 5 times within a 15 minute period. 

The current implementation logs the first failed login attempt, and then the most recent logged in attempt, upon each attempted login the handler checks if the user is locked out, during that check we check to see if the first login was 15 minutes prior to our window duration (15 minutes) if it has been over 15 minutes then we reset the failed login attempts, if the user is locked out we render the amount of time left before they can attempt again and don't pass through the login attempt so no more counters get incremented. If a password is reset, we check if that user has any entries in the database table, and if they do, we reset the failed attempts object for that user in the table.

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1209478698507743?focus=true

## Screenshot(s)

image

### DIFF
--- a/backend/migrations/00050_add_failed_login_attempts_table.sql
+++ b/backend/migrations/00050_add_failed_login_attempts_table.sql
@@ -1,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE public.failed_login_attempts (
+    user_id INTEGER NOT NULL PRIMARY KEY, 
+    first_attempt_at timestamptz NOT NULL DEFAULT NOW(),
+    last_attempt_at timestamptz NOT NULL DEFAULT NOW(),
+    attempt_count integer NOT NULL DEFAULT 0, 
+    locked_until timestamptz NULL,
+    FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+CREATE INDEX user_failed_login_attempts_user_id_idx ON public.failed_login_attempts(user_id);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE public.failed_login_attempts CASCADE;
+-- +goose StatementEnd
--- a/backend/migrations/00050_add_failed_login_attempts_table.sql
+++ b/backend/migrations/00050_add_failed_login_attempts_table.sql
@@ -1,6 +1,6 @@
 -- +goose Up
 -- +goose StatementBegin
-CREATE TABLE public.failed_login_attempts (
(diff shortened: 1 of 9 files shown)


jtucholski (changes requested):
@CK-7vn Ran through these test cases. Looks good! 

✅ Application locks the user's account after **5 consecutive failed login attempts**.
✅ User's account is locked for 15 minutes from last failed attempt.
✅ After waiting 15 minutes, user can login again.
✅ Failure count resets to zero when logging in before the lockout threshold.
✅ When an admin resets the password, failure count rests to zero.

---

Can you make these changes? 

Clear password on failed authentication
> I realized it would be easy to spam the `enter` key when attempting to log in and quickly lock out the user's account. They may not realize something even happened because there's no screen flicker or reload. Can you clear the password field after a failed attempt to help the user realize they need to take action?

Change message with 1 minute left
> Right now it says 1 minutes. Can you make it 1 minute?

Change message with 0 minute left
> Right now it says "Account locked. Try again in 0 minutes." Can it include the seconds instead?

backend/migrations/00050_add_failed_login_attempts_table.sql
```diff
@@ -0,0 +1,17 @@
...
+    attempt_count integer NOT NULL DEFAULT 0, 
+    locked_until timestamptz NULL,
+    FOREIGN KEY (user_id) REFERENCES public.users(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+CREATE INDEX user_failed_login_attempts_user_id_idx ON public.failed_login_attempts(user_id);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+DROP TABLE public.failed_login_attempts CASCADE;
```
PThorpe92, line 16: can we add `IF NOT EXISTS` to all of these? it makes it easier in the case of a migration getting stuck on up/down if there is an error

backend/src/database/users.go
```diff
@@ -598,3 +599,64 @@ func adjustUserOrderBy(arg string) string {
...
+			if rec.AttemptCount >= models.MaxFailures {
+				lockExpiry := now.Add(models.LockDuration)
+				rec.LockedUntil = &lockExpiry
+			}
+		}
+		return tx.Save(&rec).Error
+	})
+}
+
+func (db *DB) IsAccountLocked(userID uint) (bool, time.Duration, int, error) {
```
PThorpe92, line 636: nit: for this many return values, maybe make this a struct for the named fields?

PThorpe92 (approved):
Very good code overall

PR #898, opened 2025-06-02
fix: revalidate loader data in program overview

CK-7vn:
## Description of the change
This PR makes sure that when a class is cancelled in the program overview dashboard the stat cards are updated without needing a refresh. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210238090961015?focus=true

## Screenshot(s)

https://www.loom.com/share/8b0600c65e5f49e2b64535e8a171c9cb?sid=2a119225-77b4-4d00-bc07-e75c3c1bf0e0


frontend/src/Pages/ProgramOverviewDashboard.tsx
```diff
@@ -43,6 +43,7 @@ export default function ProgramOverviewDashboard() {
     const { program_id } = useParams<{ program_id: string }>();
     const user = useAuth();
     const userFacilityId = user.user?.facility_id;
+    const revalidator = useRevalidator();
```
carddev81, line 46: I like this function! I didn't know it existed.

carddev81 (changes requested):
When a class has active enrollments and when changing the status from 'Active' to 'Complete' the Active Enrollments card is not being updated.  
- [x] Active Enrollments updates when selecting Cancelled.

frontend/src/Pages/ProgramOverviewDashboard.tsx
CK-7vn: Also though, ya, the 'active' to 'complete' is handled differently the api call is actually made in the modal for the class status' but we have the cancel call from the program overview dashboard, very un intuitive for sure. Easy fix though. Didn't even think about it.

CK-7vn: Actually...I feel like that's an issue somewhere's else...because it's not updating the field even when I refresh the page, so I don't think that has anything to do with this ticket, but i'll look into it real quick.

carddev81 (approved):
Looks good!  Tested good!

calisio (approved):

PR #899, opened 2025-06-03
feat: integration testing library

jtucholski:
## Description of the change

Introduces an integration testing framework in `/backend/tests/integration`. Run it with `go test` or `go test -v`. 

These tests that are included provide an example of how we would go about writing integration tests within the codebase. My thought is that we'd have a file per handler and a test function per handler (e.g. `GET /api/programs/{id}/classes`, `POST /api/programs/{program_id}/classes`, etc.).

Each test function would include 1 or more tests validating that the endpoint handles the provided request correctly. This includes tests that verify it succeeds and fails when expected.

Includes:
* `*_test.go` - Starting examples for writing the integration tests. We can iterate off of these. Includes: `classes`, `facilities`, and `users`.
* `client.go` - Client, Request, and Reponse wrappers to abstract some of the implementation away to call the test server.
* `testenv.go` - Sets up, cleans up, and includes some of the builder functions to create data needed for the tests.

Changes:
* `/models/*`
  * Modifications to support the many2many relationship for `facilities_programs`. Program now has `[]Facility` (for backward compatability) and `[]FacilitiesPrograms` (for handling the mapping between the two). 
* `/database/*` 
  * Dropped calls to seed test data. Handling that inside of tests.
* `/handlers/*`
  * `testingMode` used during integration testing and with the `X-Test-Claims` header to bypass auth checks in Kratos.


jtucholski:
@carddev81 Yeah, I think we will test _most scenarios._ The one you described looks like it is checking for valid/invalid usernames/names and that'll be a great candidate for one since its driven by user input. 

I omitted a lot so that I could get a few examples in sooner of varying levels of difficulty. I think I will (in a separate PR) also include an event or attendance one as well.

backend/src/handlers/user_handler.go
```diff
@@ -167,20 +166,22 @@ func (srv *Server) handleCreateUser(w http.ResponseWriter, r *http.Request, log
...
-	if err != nil {
-		log.error("error getting kolibri instance")
-		// still return 201 because user has been created in kratos,
-		// kolibri might not be set up/available
-		return writeJsonResponse(w, http.StatusCreated, response)
-	}
-	if err := srv.CreateUserInKolibri(&reqForm.User, kolibri); err != nil {
-		log.add("user_id", reqForm.User.ID)
-		log.error("error creating user in kolibri")
+	if !srv.testingMode {
```
PThorpe92, line 169: man if there was one thing that I disliked about the previous test setup, it was the in-handler branches for testing mode.

Maybe we replace the kratos client with a mock one in the case of testing mode?

backend/tests/integration/healthcheck_test.go
```diff
@@ -0,0 +1,19 @@
+package integration
+
+import (
+	"net/http"
+	"testing"
+)
+
+func TestHealthCheckEndpointHandler(t *testing.T) {
```
PThorpe92, line 8: this feels a bit excessive :stuck_out_tongue_winking_eye:

jtucholski: 😂 This was me making sure it was working initially before dealing with everything else.

backend/src/handlers/user_handler.go
jtucholski: Yeah, I know I went back and forth and I remember the frustration with mocks too. 

I'm fine with injecting a mock kratos client. Do we use mocks anywhere else that I should reference?

PThorpe92 (approved):

PR #900, opened 2025-06-03
Edit Program UI adjust

corypride:
## Description of the change
Adjusted spacing between cards and removed indent of the form on Edit Program page

## Screenshot(s)
image

---
  - https://app.asana.com/0/0/1210356415337034


jtucholski (changes requested):
Layout looks right, some of those classes that got removed were necessary for the multi-select drop-down. Either look to bring them back in or add them to the specific fields. Notice the outline cursor box in the form fields that shows up in the After image.

## After

image

## Before

image

PR #901, opened 2025-06-04
feat: add logic to reschedule event series from facility schedule

carddev81:
## Description of the change

- Added the functionality for allowing an administrator to reschedule an event series from a facility schedule. 
    - Modified/Added logic within the backend API
    - Modified/Added logic within the frontend API

- **Related issues**: Link to Asana tickets [Reschedule Series from Facility Schedule](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210197345470707?focus=true)

## Screenshot(s)
image

### DIFF
--- a/frontend/src/Components/modals/RescheduleClassEventSeriesModal.tsx
+++ b/frontend/src/Components/modals/RescheduleClassEventSeriesModal.tsx
@@ -23,7 +23,7 @@ import { RRule } from 'rrule';
 import { fromZonedTime, toZonedTime } from 'date-fns-tz';
 import { RRuleControl, RRuleFormHandle } from '../inputs/RRuleControl';
 import { addDays } from 'date-fns';
-import { parseRRuleUntiDate } from '../helperFunctions';
+import { parseRRuleUntilDate } from '../helperFunctions';
 
 export const RescheduleClassEventSeriesModal = forwardRef(function (
     {
@@ -140,7 +140,7 @@ export const RescheduleClassEventSeriesModal = forwardRef(function (
                               }
                               endDateVal={
                                   calendarEvent &&
-                                  parseRRuleUntiDate(
+                                  parseRRuleUntilDate(
                                       calendarEvent.recurrence_rule,
                                       user?.timezone
                                   )
--- a/frontend/src/Components/helperFunctions/formatting.tsx
+++ b/frontend/src/Components/helperFunctions/formatting.tsx
@@ -88,12 +88,17 @@ export function parseRRule(
             timeZone: timezone
         });
     } catch (error) {
-        console.error('error parsing rrule, error is: ', error);
+        console.error(
+            'error parsing rrule, rRule is: ',
+            rRule,
+            '; error is: ',
+            error
+        );
     }
-    return rRule;
+    return '';
 }
 
-export function parseRRuleUntiDate(rRule: string, timezone: string): string {
+export function parseRRuleUntilDate(rRule: string, timezone: string): string {
     try {
         const rule = RRule.fromString(rRule);
         if (rule.options.until) {
--- a/frontend/src/Components/helperFunctions/index.ts
+++ b/frontend/src/Components/helperFunctions/index.ts
@@ -6,5 +6,5 @@ export {
     isEndDtBeforeStartDt,
     isPastDate,
     parseRRule,
-    parseRRuleUntiDate
+    parseRRuleUntilDate
 } from './formatting';


calisio (changes requested):
I haven't taken a look at the code yet, but here are some things to consider:
- the instructor field is missing from the form. It is not in the table in asana, but instructor is listed below in the bullet points. I would double check to see if they want this field in there or not.
- I cannot click on an overrided event to edit it. Is this expected behavior?
    - ie. Computer science is scheduled for Saturday, June 21. I reschedule it for Friday, June 20. I now click on the Computer science class on june 20 on the calendar, and I am unable to click edit event, edit series, or cancel event.
- bug: if i edit the event series to the same day, but have the time as later in the day, it does not modify the session that day. Should we keep both events, even if they are on the same day? Should we assume they would like to change the event for that day to a later time, rather than repeat the class on the calendar? Not sure the answer, but something to consider.

https://github.com/user-attachments/assets/7f87aa30-2272-4b88-8744-68f6cdab3bbe

- bug: This is a tough bug to describe, but I will do my best. I modify session 1 from mondays at 3pm to tuesdays at 4pm starting june 24. the sessions before the date I listed (June 23) is still at 3pm. Now, if I modify that session, from june 23 onwards, and select only mondays, it does not remove the tuesday dates, but instead duplicates the sessions so now they are both on monday AND tuesday. If i modify the tuesday dates to a different time on tuesdays, it does not remove the monday dates. Now, i know this is a complicated bug, but let me know if you need any more explaining. An example of the change is in a video below.

https://github.com/user-attachments/assets/52c18c28-a7a3-450c-8aff-3a983bd6448d

- small bug that existed from before (not from this PR, just would prefer it fixed ASAP!)— schedule was showing archived classes. could you throw this in your PR so we can fix it along the way? edit line 439 in `class_events.go` to `Joins("JOIN program_classes c ON [c.id](http://c.id/) = pcev.class_id AND c.archived_at IS NULL").`  Thanks!

### ✅ **UI/UX**

- [x]  "Edit Series" option is available when selecting a recurring event.
- [x]  Clicking "Edit Series" opens a **modal** with:
    - [x]  Title: **"Edit Schedule (Recurring)"**
    - [x]  Body text: *"You are editing the recurring schedule for this class. Changes will apply to this session and all future occurrences."*
    - [x]  Fields:
        - [x]  **Start Time** (time picker)
        - [x]  **End Time** (time picker)
        - [x]  **Room** (textbox or dropdown)
        - [ ]  **Instructor** (dropdown)
        - [x]  **Days of week** (M/T/W/R/F toggles)
        - [x]  **Frequency** (dropdown: e.g., Weekly, Biweekly)
        - [x]  **Ends** (optional end date picker)
    - [x]  Buttons: **Cancel**, **Update Schedule**

---

### ✅ **Validation & Business Logic**

- [x]  **Start Time < End Time** (valid time range)
- [x]  **Room is required**
- [ ]  **Instructor is required**
- [x]  **Frequency and Days** are validated to prevent invalid combinations (e.g., no days selected)
- [x]  If **"Ends"** date is provided, it must be **after the current date**
- [x]  **Only future events** are affected — past events stay unchanged

---

### ✅ **Data & Backend**

- [x]  Edits apply to the **recurring rule (RRule)** starting from the selected event date forward
- [x]  Existing override or cancelled events are **not overwritten**
- [x]  Changes create a new recurrence rule and end the current one at the cutoff
- [x]  Past event data is preserved

---

### ✅ **Class History Logging**

- [x]  A log entry is created in the class history:
    
    > All future sessions rescheduled starting {effective date} by {admin username} on {action date}
    > 
- [ ]  All updated fields (time, room, days, etc.) are captured in the change history

---

### ✅ **Calendar Integration**

- [x]  All **future instances** of the event update on the calendar view
- [x]  **Past events remain unchanged**
- [ ]  New schedule appears accurately on calendar, respecting frequency and day-of-week settings

---

### ✅ **Testing & Edge Cases**

- [ ]  Changing only one field (e.g., room) still updates series correctly
- [x]  Events update properly across month/week/day views
- [x]  Attempting to set invalid times/dates triggers errors
- [ ]  Calendar reflects updates without duplication or gaps
- [x]  Edited series doesn’t override **previous overrides or cancellations**

Let me know if I am off base on something. I will begin looking over the code now, but thought I should get you this feedback sooner rather than later.

carddev81:
`I cannot click on an override event to edit it. Is this expected behavior?`
For now yes, this is the expected behavior, I held off of what to do here in the case a user wants to remove the override (cancel/reschedule) do we remove it and display the original event? (such as a cancelled date, and a cancelled date that was rescheduled to a different date). Figured we would get all the initial logic in place then come back to handle those edge cases?

`The instructor field is missing from the form. It is not in the table in asana, but instructor is listed below in the bullet points. I would double check to see if they want this field in there or not.`
Thought the instructor field was mistake since this is a field of the class not an event? I will ask the question in Asana.

Looking over bug issues and then will make the required changes.

frontend/src/Components/inputs/RRuleControl.tsx
```diff
@@ -31,9 +35,11 @@ interface RRuleControlProp {
     onChange?: (isValid: boolean) => void;
     startDateRef: string;
     endDateRef: string;
-    errors: FieldErrors<any>; // eslint-disable-line
+    formErrors?: FieldErrors<any>; // eslint-disable-line
     register: UseFormRegister<any>; // eslint-disable-line
     getValues: UseFormGetValues<any>; // eslint-disable-line
+    initialRule?: string;
+    initialDuration?: string;
```
calisio, line 42: I see these have been added, but I don't see them used anywhere in this file? I assume I may have caught this in mid edit maybe?

frontend/src/Components/modals/RescheduleClassEventSeriesModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+                <>
+                    <p className="mb-4">
+                        You are editing the recurring schedule for this class.
+                        Changes will apply to this session and{' '}
+                        <b>all future occurrences.</b>
+                    </p>
+                </>
+            )
+        },
+        ...(formDataRef
```
calisio, line 112: Would you be able to explain formDataRef to me? I am not sure I am completely following this addition into FormModal.

frontend/src/Components/ClassLayout.tsx
```diff
@@ -52,14 +69,19 @@ function ClassInfoCard({ classInfo }: { classInfo?: Class }) {
...
+                    );
+                    return classEndDate
+                        ? classEndDate.toLocaleDateString('en-US', {
+                              year: 'numeric',
+                              month: 'long',
+                              day: 'numeric',
+                              timeZone: 'UTC'
+                          })
+                        : 'No end date scheduled';
+                })()}
```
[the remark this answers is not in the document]
carddev81: are you talking about end date?

carddev81:
> @carddev81 It works really well though I noticed an issue that I don't know how to deal with.
> 
> Here's the gist:
> 
> * We have a recurring event series, for example every Tuesday.
> * I use edit series and move them all to Wednesday. **This works fine**
> * I use edit event to move just the first event back to Tuesday. **This creates an override and works fine.**
> * After that clicking edit series on the overridden event does not work. ** But I don't know the expected behavior**
> 
> I'm not sure which route to take and I'm open to making this a separate issue if it helps get this feature merged (in case its holding anything else up).
> 
> I tried to do this in Google w/my calendar and it kind of fixed the series, but it's probably an edge case.
> 
> Options:
> 
> * Don't allow edit series for events that have been overridden. We probably need to show a message if its disabled, i,e, "This event was manually edited and is not a part of the series.",
> * Detect that the user clicked an override event and prompt, "Some events in this series were manually edited. Do you want to apply this change to those as well?"
>   
>   * Clicking yes deletes the overrides.
>   * Clicking no updates the series only.
> * Assume the user is doing the prior option and automatically delete the overrides when they click edit series. Probably confusing to some users though

@jtucholski I would prefer to make this a separate issue. I would like to discuss the optional behaviors that can occur. Can we tag this one issue with [Restore Cancelled Event](https://app.asana.com/1/1201607307149189/project/1209465647700741/task/1210197345470691).

calisio (approved):
All comments addressed and updated. Looks good to me!

jtucholski:
@carddev81 -- A UX behaviors I noticed to see if you can resolve

- [ ] Always populate the fields after clicking Edit Series
- [ ] Populate the room field after clicking Edit Series

## Always populate the fields after clicking Edit Series

This might be one of the issues that I was experiencing. If I click Edit Series for an event, close the modal, and then re-click Edit Series (without clicking on another event) the fields are mostly empty (times, room etc.).

## Populate the room field after clicking Edit Series

The rest of the fields are populated when I click Edit Series, but not the Room. Can we populate that?

carddev81:
> one of the issues that I was experiencing. If I click Edit Series for an event, close the modal, and then re-click Edit Series (without clicking on another event) the fields are mostly empty (times, room

made changes/pushed

PR #903, opened 2025-06-05
Add Program name to TransferSummaryPanel text.

corypride:
## Description of the change
Included the name of the program that the class is a part of (bc it may not always match the class name).

- **Related issues**: 
[Ticket id: 268](https://app.asana.com/1/1201607307149189/project/1209457756469406/task/1210272664885981?focus=true)
## Screenshot(s)
image

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -62,15 +62,15 @@ func (db *DB) GetUsersByIDs(userIDs []uint, args *models.QueryContext) ([]models
 	return users, nil
 }
 
-func (db *DB) GetTransferProgramConflicts(ctx context.Context, id uint, transferFacilityId int) ([]string, error) {
-	var programNames []string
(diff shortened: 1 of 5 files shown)


jtucholski (approved):

carddev81 (approved):
So, I have only one finding, it doesn't specify but wanted to make sure since I seen it within the screenshot...when the program is listed alongside the class name should it be formatted like `Program: Anger Management` like within the screenshot in the ticket?

image

Overall the code looks great. Tested good.
- [x] Change to “The resident will be removed from {FACILITY}.”  
- [x] Change to “The resident will be unenrolled from all active classes and programs.”  
- [x] Change to “The following programs are not available at the new facility: {CLASS_NAME} ({PROGRAM_NAME})”  
- [x] Change to “The resident’s account will be added to {FACILITY}.”  
- [x] Change to “Staff must re-enroll resident in available programs at {FACILITY}.”  
- [x] Change to “The resident’s account history and favorites will be preserved.”

PR #904, opened 2025-06-06
Fix: Class Status Pill Issue on Program Overview Dashboard 

corypride:
A bug was identified where the Class Status pill would not render the ModifyClassModal after cancelling by esc and attempting to switch the status again.

Solution:

Added an event listener to close the modal when the escape key is pressed, ensuring the modal can be properly dismissed.
## Screenshot(s)

https://github.com/user-attachments/assets/fca8ceda-a50b-421a-ab75-be065e5ed1ae

 Related to: https://app.asana.com/1/1201607307149189/inbox/1209552815576329/item/1210272664885957/story/1210390032326801

---
  - https://app.asana.com/0/0/1210272664885957

### DIFF
--- a/frontend/src/Components/modals/TextOnlyModal.tsx
+++ b/frontend/src/Components/modals/TextOnlyModal.tsx
@@ -34,7 +34,13 @@ export const TextOnlyModal = forwardRef(function TextModal(
     ref: React.ForwardedRef<HTMLDialogElement>
 ) {
     return (
-        <dialog ref={ref} className="modal">
+        <dialog
+            ref={ref}
+            className="modal"
+            onKeyDown={(event) => {
+                if (event.key === 'Escape') onClose();
+            }}
+        >
             <div className={`modal-box ` + (width ? width : '')}>
                 <CloseX close={onClose} />
                 <div className="flex flex-col gap-6">


jtucholski:
@calisio I thought we ran into this elsewhere but outside of the Programs page and the Classes page this is the only place we use the status pill. 

I noticed that those pills also have the same issue too where if the user presses Esc to close the modal and then tries to change the pill status value after that the modal won't appear.

calisio:
> @calisio I thought we ran into this elsewhere but outside of the Programs page and the Classes page this is the only place we use the status pill.
> 
> I noticed that those pills also have the same issue too where if the user presses Esc to close the modal and then tries to change the pill status value after that the modal won't appear.

Somehow missed this comment @jtucholski ! I don't recall exactly where this was, but I believe it was maybe with the action modals on several pages (such as editing a resident, and then pressing esc). 

In this case, the modal is not opening because the `selectedModifyOption` is not changing, therefore it is not triggering the `useEffect` as seen on line 133 of `ClassStatus.tsx`. A similar implementation can be found in `ProgramStatus.tsx`. We probably should not call show modal inside of a use effect, but rather when we actually click one of the options (like how we do it in edit user, for example). 

This is likely the fix to this ticket.

jtucholski:
@calisio I definitely believe the fix should be made in `ClassStatus.tsx` as well so that whenever that component is used it automatically provides the open modal behavior. 

I could be wrong, but I think that `ProgramStatus.tsx` has the same issue too. If I click the dropdown and select Inactive, press Esc, and then try again the modal doesn't show.

frontend/src/Components/modals/ModifyClassModal.tsx
```diff
@@ -57,10 +57,20 @@ const ModifyClassModal = forwardRef(function (
         setSelectedStatus: React.Dispatch<
             React.SetStateAction<SelectedClassStatus>
         >;
-        onClose?: () => void;
+        onClose: () => void;
     },
     ref: React.ForwardedRef<HTMLDialogElement>
 ) {
+    useEffect(() => {
```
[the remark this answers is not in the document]
jtucholski: @corypride I must have misunderstood, I thought you were removing the `useEffect` here?

frontend/src/Components/ClassStatus.tsx
```diff
@@ -127,13 +127,11 @@ export default function ClassStatus({
             return;
         setDropdownOpen(false);
         setSelectedModifyOption(option);
+        setTimeout(() => showModal(modifyClassRef), 0);
+        showModal(modifyClassRef);
```
jtucholski, line 131: @corypride This looks like a code smell, calling `showModal` two lines apart, one with a setTimeout of 0ms. Is this the final solution? If it is, then can you add a comment why this is necessary?

frontend/src/Components/ProgramStatus.tsx
```diff
@@ -86,17 +86,16 @@ export default function ProgramStatus({
...
     const availableActions = programActions.get(effectiveStatus) ?? [];
 
-    function openSelectionModal(action: ProgramAction) {
+    function openSelectionModal(
+        e: React.MouseEvent<HTMLDivElement, MouseEvent>,
+        action: ProgramAction
+    ) {
         setSelectedAction(action);
         setDropdownOpen(false);
+        setTimeout(() => showModal(modifyProgramRef), 0);
```
jtucholski, line 95: @corypride Why is this solution different than the one on `ClassStatus.tsx`?

calisio (changes requested):
Hey Cory, thanks for working on this ticket. While your approach works, generally I try to avoid using timeouts, and adding event listeners is usually also advised against in React (unless its impossible to avoid). In earlier comments, I pointed towards referencing the edit modal since it would likely have similar implementation, and that modal worked on keydown of esc in still reopening the correct modal. 

I looked into the implementation on `EditUserModal.tsx`, which rather than using a `<TextOnlyModal />`, it uses a `<FormModal/>`, which I think is a key difference. I found that the issue is that the `onClose` is not being called in the component (both `ClassStatus.tsx` and `ProgramStatus.tsx` (ie `onClose={() => setSelectedModifyOption(undefined)}` was never being called), therefore the `selectedModifyOption` state was never being updated to undefined. That is why the `showModal` within the `useEffect` on line 133 of `ClassStatus.tsx` was not being called, therefore not showing the modal as expected.

The idea of calling `onClose` on keydown of the escape key is the right one, but the fix should apply to both the class status page and the program status page (this keydown listener is only applied on the modify class modal). This issue is probable to exist on other forms that also use a `<TextOnlyModal />`, therefore we should probably have the fix live in that file.

I poked around a bit, and I believe that adding the following to `<TextOnlyModal />` should do the trick to fix the bug of this ticket, along with any future issues of esc with text only modals:
```
        <dialog
            ref={ref}
            className="modal"
            onKeyDown={(event) => {
                if (event.key === 'Escape') onClose();
            }}
        >
```
This way, the modal uses React's built in logic to register a keydown event and calls `onClose()` if the key is esc, and avoids using timeouts and window event listeners. Let me know if any part needs more explaining/doesn't make sense.

carddev81 (approved):
Looks great!  Tested ESC key and Cancel button and both tests passed.

PR #905, opened 2025-06-07
fix: remove duplication bug in programs overview, create materialized view

PThorpe92:
A bug was found in staging:
image

and I discovered that the query was allowing for duplicates by joining on a table that didn't have a 1-1 relationship. After looking a little deeper I thought this query might be better expressed as a mat view, so this adds the plumbing for these 3 views:
30d
90d
all time

along with the cron job to refresh the views.

This also makes some improvements to the middleware that remove some boiler plate.

## What I tested:

1. The migrations:
  - the views will be created if they don't exist, and refreshed if they are
 
2. The query:
 - returns the expected data, free of duplicates
 - search works
 - filtering for 30/90/all_time works

3. The cron job:
 - works
 - is initiated and published hourly

4. Interaction:
 - When programs statuses are changed, the data is reflected immediately because views are refreshed when the relevant calls are made.
 - This includes updating program statuses, deleting or updating a program.
 - *otherwise* the data could be at most 59 min old/stale when a user sees it on the ProgramManagement dashboard.


backend/src/database/DB.go
```diff
@@ -195,6 +195,120 @@ func (db *DB) SeedDefaultData(isTesting bool) {
...
 		}
 	}
+	if err := initProgramMaterializedViews(db.DB); err != nil {
+		logrus.Fatalf("Failed to initialize program materialized views: %v", err)
+	}
+}
+
+func initProgramMaterializedViews(db *gorm.DB) error {
+	const sqlTemplate = `CREATE MATERIALIZED VIEW programs_overview_%s AS
+WITH recent_metrics AS (
```
carddev81, line 205: Within this MQT it is only querying all facilities data (`daily_program_facilities_history`) for dept admins+, we are missing the call for per facility data for facility admins (`daily_program_facility_history`). Previously the `GetProgramsOverviewTable()` method handled both scenarios--see removed lines 404-406 within backend/src/database/programs.go.

carddev81 (changes requested):

PR #912, opened 2025-06-09
fix: join/duplication issue on programs overview

PThorpe92:
Abandoned the mat view idea, somehow I completely missed the facilities filtering. this is just an easy fix


carddev81 (approved):
Good to go, tested and reviewed!

PR #914, opened 2025-06-10
test: creating handlers

JamesUnlocked:
## test fails if facility has invalid data

A handler was added to ensure the facility was created correctly.

### DIFF
--- a/backend/tests/integration/facilities_test.go
+++ b/backend/tests/integration/facilities_test.go
@@ -76,7 +76,34 @@ func TestCreateFacilityHandler(t *testing.T) {
 func TestUpdateFacilityHandler(t *testing.T) {
 	env = SetupTestEnv(t)
 	defer env.CleanupTestEnv()
+	// Begining of code from James
+	t.Run("Update facility fails with invalid data", func(t *testing.T) {
+		facility, err := env.CreateTestFacility("Test Facility") // Creating a test facility to update
+		require.NoError(t, err)                                  // An assertion to ensure the facility was created successfully
 
+		// Attempt to update with invalid data (e.g., empty name)
+		facility.Name = "" // Set an invalid name
+		NewRequest[models.Facility](env.Client, t, http.MethodPatch, fmt.Sprintf("/api/facilities/%d", facility.ID), facility).
+			WithTestClaims(&handlers.Claims{Role: models.SystemAdmin}).
+			Do().
+			ExpectStatus(http.StatusBadRequest).
+			ExpectMessage("facility name is required") // Expect a specific error message
+	})
+	t.Run("Update facility fails with non-existent ID", func(t *testing.T) {
+		// Attempt to update a facility that does not exist
+		nonExistentID := 9999 // Assuming this ID does not exist
+		facility := models.Facility{
+			Name:     "Non-existent Facility",
+			Timezone: "America/Chicago",
+		}
+		NewRequest[models.Facility](env.Client, t, http.MethodPatch, fmt.Sprintf("/api/facilities/%d", nonExistentID), facility).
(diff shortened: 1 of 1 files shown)


backend/tests/integration/facilities_test.go
```diff
@@ -76,7 +76,34 @@ func TestCreateFacilityHandler(t *testing.T) {
 func TestUpdateFacilityHandler(t *testing.T) {
 	env = SetupTestEnv(t)
 	defer env.CleanupTestEnv()
+	// Begining of code from James
+	t.Run("Update facility fails with invalid data", func(t *testing.T) {
+		facility, err := env.CreateTestFacility("Test Facility") // Creating a test facility to update
+		require.NoError(t, err)                                  // An assertion to ensure the facility was created successfully
 
+		// Attempt to update with invalid data (e.g., empty name)
+		facility.Name = "" // Set an invalid name
```
PThorpe92, line 85: When you deserialize json into a Go struct, Go will automatically add 'default' values for fields that are not present. 

If we look at the handler: 
```go
	var facility models.Facility
	err = json.NewDecoder(r.Body).Decode(&facility)
	if err != nil {
	}
```
we see that it is deserializing into a models.Facility, so if the `name` field is not included at all, it will be an empty string anyway.

image
 
But since the UI enforces that there is a Name when you edit a facility, we have to send a curl request to actually test this out. As you can see, it will successfully update when there is no name provided.
 
but when we curl again, we can see that the name was not updated. :thinking: 
```json
{
  "message": "Data fetched successfully",
  "data": {
    "id": 1,
    "created_at": "2024-11-19T03:23:39.967529Z",
    "updated_at": "2025-06-13T14:55:07.801134Z",
    "name": "mvcf",
    "timezone": "America/Chicago",
    "programs": null,
    "login_activity": null
  }
}
```

This demonstrates how `gorm`'s `db.Updates` method works.. Since Go has this weird behavior of just filling in whatever it has decided are the default values for struct fields, it will not explicitly update a field when the field is in it's default value state. In order to do that, you would need to use gorm's `Update` method that takes the field and the new value 
```go
db.Model(&models.Enrollment{}).Update("enrollment_status", enrollmentStatus).Error
```

The TLDR here, is that setting the name to an empty string will not actually update it _most_ of the time.

And the handler itself doesn't actually check that the name is required 
```go
func (srv *Server) handleUpdateFacility(w http.ResponseWriter, r *http.Request, log sLog) error {
	if !userIsSystemAdmin(r) {
		return newUnauthorizedServiceError()
	}
	id, err := strconv.Atoi(r.PathValue("id"))
	if err != nil {
		return newInvalidIdServiceError(err, "facility ID")
	}
       // we aren't doing any manual checks for the facility name here
	log.add("facilty_id", id)
	var facility models.Facility
	err = json.NewDecoder(r.Body).Decode(&facility)
	if err != nil {
	} 
	err = srv.Db.UpdateFacility(&facility, uint(id))
	if err != nil {
		log.add("facilityName", facility.Name)
	}
	return writeJsonResponse(w, http.StatusOK, "facility updated successfully")
}
```

and the relevant `UpdateFacility` method:
```go
func (db *DB) UpdateFacility(facility *models.Facility, id uint) error {
      // we would need to add `Validate().Struct(facility)` here, and add a struct tag to the facility name field in models/facility, in order to return that BadRequest that you are testing for 
	if err := db.Model(models.Facility{}).Where("id = ?", id).Updates(&facility).Error; err != nil {
		log.WithField("facility_id", facility.ID).Error("error updating facility name database/UpdateFacility")
		return newUpdateDBError(err, "facilities")
	}
	return nil
}
```

I hope this explains enough

corypride: @PThorpe92 a struct tag as relates to validation like alphnum, min, max, ..?

PR #915, opened 2025-06-10
fix: update completion details modal 

CK-7vn:
Updates the Completion details modal with new verbage, and includes a check that guarantees that the date shown is either 1. The start date of the class if the user was enrolled prior to class tart 2. The date the user was enrolled if enrolled post start of the class

Outside of that, there was a couple bugs due to changeStatusValue not being set in time for renders and PATCH calls. Prior to this you were unable to bulk update or singularly update "Completed" status due to this issue. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210481100839405?focus=true

## Screenshot(s)
image

### DIFF
--- a/backend/migrations/00051_add_enrolled_on_dt_to_completion.sql
+++ b/backend/migrations/00051_add_enrolled_on_dt_to_completion.sql
@@ -1,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE public.program_completions ADD COLUMN IF NOT EXISTS enrolled_on_dt TIMESTAMP WITHOUT TIME ZONE NULL;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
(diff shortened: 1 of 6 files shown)


carddev81 (changes requested):
The enrolled on date isn't matching the enrolled date within the listing. I'm wondering if this is due to the column meta being 'without timezone'? The program_classes_enrollment created_at column is created 'with timezone'. I'm not sure this is the reason just a thought. Also, after seeing this issue I notice that the other dates seem off too, I know this doesn't have to do with this particular ticket, but just wanted to note this for future reference. For example the program_class_start_dt of 06/13 displays as 14 on the screen which is actually correct, but is being inserted into the table as an incorrect value. Timezone/UTC issue? Thoughts?

EnrolledDtNotMatching

TimestampsAreNotReflectiveOfActualDates

PR #916, opened 2025-06-10
Show Reason Add Class is Disabled & Block classes at facilities its not eligible for

corypride:
## Front-end
- [x] **Verify** that the “Add Class” button is disabled for **inactive** programs  
  *Tooltip:* “This program is inactive and cannot accept new classes.”
- [x] **Verify** that the “Add Class” button is disabled for **archived** programs  
  *Tooltip:* “This program has been archived and cannot be modified.”
- [x] **Implement** disabling of “Add Class” when the program is **not offered** at the selected facility  
  *Tooltip:* “This program isn’t available at the selected facility.”
- [x] Ensure Department Admins and Super Admins are **subject to the same** disabling logic

## Back-end
- [x] In the **CreateClass** handler, reject creation if the program is **inactive**  
  • Respond HTTP 400 with message “Program is inactive and cannot accept new classes.”
- [x] Reject if the program is **archived**  
  • Respond HTTP 400 with message “Program has been archived and cannot be modified.”
- [X] Reject if the program is **not offered** at the specified facility  
  • Respond HTTP 400 with message “Program isn’t available at the selected facility.”
- [x] Enforce these checks **regardless of user role** (including Admin/Super Admin)

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210488775054989

Note from @jtucholski: Edits to `auth.go` to make sure that the resolver always gets run, even in testing mode. Prior to that, it was only run in production.

## Screenshot(s)

https://github.com/user-attachments/assets/6053a62f-3c70-432d-83bc-f6ed50c42db8

### DIFF
--- a/backend/src/models/program.go
+++ b/backend/src/models/program.go
@@ -52,15 +52,6 @@ type Program struct {
 
 func (Program) TableName() string { return "programs" }
 
-func (p *Program) OfferedAtFacility(facilityID uint) bool {
-	for _, f := range p.Facilities {
-		if f.ID == facilityID {
-			return true
-		}
-	}
-	return false
-}
-
 type ProgramType struct {
 	ProgramType ProgType `json:"program_type" gorm:"primaryKey;type:program_type" validate:"required"`
 	ProgramID   uint     `json:"program_id" gorm:"primaryKey;not null" validate:"required"`
--- a/frontend/src/Pages/ProgramOverviewDashboard.tsx
+++ b/frontend/src/Pages/ProgramOverviewDashboard.tsx
@@ -223,11 +223,11 @@ export default function ProgramOverviewDashboard() {
             archivedProgram:
                 'This program has been archived and cannot be modified.',
             notOfferedAtFacility:
-                'This program isn’t available at the selected facility.'
+                'This program is not available at the selected facility.'
         };
         return tooltipMap[status];
     }
-    //comment
+
     return (
         <div className="p-4 px-5">
             <div className="flex flex-col gap-4">


jtucholski (commented):
@corypride I just have one question about the `getProgramStatus` before approving it. Looks like it became more complicated and I want to be careful before we add it in and block the wrong person from creating classes.

calisio:
Just an update @corypride @carddev81 -- I dropped the eslint build file changes and will put it in a different PR. It didn't feel too relevant here, and there is more research I want to do but don't want to hold up this PR. I left the commit that fixed the linting error, though.
@corypride Could you make sure the commit history is cleaned up? Ideally I'd like to remove the several commits we added throughout that attempted to fix the linting error so they don't litter our commit history. Let me know if you need any help with that, I am happy to assist since I was part of the problem 🤣

corypride:
Hi Carolina,

Yes, I will squash the commits and ensure the commit history is clean. If I
run into any issues during the process, I will certainly let you know, and
I appreciate your offer to assist.

Thanks,
Cory

jtucholski (changes requested):
@corypride It works well. I commented on a few sections of code that are no longer needed.

I also noticed I can navigate directly to `programs/{id}/classes/new` and create a new program as long as I am a department or super admin. I don't think we should allow for that because that means those users can circumvent the front-end validation we are applying.

I _believe_ to fix that you'll need to remove the following code in the `validateFacility` resolver. **Please validate it first though before assuming this is the only fix.**

```
if r.Context().Value(ClaimsKey).(*Claims).canSwitchFacility() {
	return true
}

```

jtucholski (approved):

PR #918, opened 2025-06-12
feat: add label to display last updated on programs management screen per ticket 288

carddev81:
## Description of the change

Added last updated label below the currently existing one below the programs table. This work consisted of adding an SQL query to get the max date from within the daily programs table (last updated date) and parsing the cron expression that is used for scheduling the job.

- **Related issues**: Link to Asana tickets [Show last updated on Programs Management](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210481100839397?focus=true)

## Screenshot(s)

image

### DIFF
--- a/backend/src/handlers/programs_handler.go
+++ b/backend/src/handlers/programs_handler.go
@@ -7,7 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"strconv"
+
+	"github.com/go-co-op/gocron/v2"
 )
 
 func (srv *Server) registerProgramsRoutes() []routeDef {
@@ -19,6 +22,7 @@ func (srv *Server) registerProgramsRoutes() []routeDef {
 		/* admin */
 		adminFeatureRoute("GET /api/programs/detailed-list", srv.handleIndexProgramsOverviewTable, axx),
 		adminFeatureRoute("GET /api/programs/stats", srv.handleIndexProgramsFacilitiesStats, axx),
+		adminFeatureRoute("GET /api/programs/stats/lastRun", srv.handleGetProgramOverviewLastUpdatedAt, axx),
 		adminFeatureRoute("GET /api/programs/{id}/history", srv.handleGetProgramHistory, axx),
 		adminFeatureRoute("POST /api/programs", srv.handleCreateProgram, axx),
 		adminFeatureRoute("DELETE /api/programs/{id}", srv.handleDeleteProgram, axx),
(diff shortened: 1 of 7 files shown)


backend/src/handlers/programs_handler.go
```diff
@@ -275,6 +279,31 @@ func (srv *Server) handleFavoriteProgram(w http.ResponseWriter, r *http.Request,
...
+		job, err := scheduler.NewJob(gocron.CronJob(cronExpression, false), gocron.NewTask(func() {}))
+		if err != nil {
+			return newBadRequestServiceError(err, "unable to create new generic job using the cron schedule: "+cronExpression)
+		}
+		next, err := job.NextRun()
+		if err != nil {
+			return newBadRequestServiceError(err, "unble to get the next run of job")
+		}
+		lastUpatedAt.LastRanTime = next.Format("3:04 PM")
+	}
```
PThorpe92, line 303: The schedules can vary, they are not all set to that schedule but either way this should prob just be a query to the job in the database and use the last run timestamp for the job.

frontend/src/common.ts
```diff
@@ -724,6 +724,15 @@ export interface ProgramOverview extends Program {
...
 }
 
+export interface RunnableTask {
+    id: number;
+    job_id: string;
+    last_run: Date;
+    provider_platform_id: number;
+    open_content_provider_id: number;
+    status: string;
+}
```
jtucholski, line 734: @carddev81 I don't have a good reason not to, but it sticks out to me that we're exposing this over the API. 

Is there a reason it couldn't come back in the `ProgramsFacilitiesStats` response to `/api/programs/stats`?

carddev81: @jtucholski The only reason I made a handler for it was that it would only be called once during page load (loader) versus every time a user selects an option from the time filter on that page--less db calls, but I can move this into the `/api/programs/stats` handler? Please just confirm this and I'll make the change ASAP.

jtucholski: I'd rather see it in `api/program/stats` than send anything about runnable tasks down to the client.

PThorpe92: @carddev81 I think he wants to avoid sending such an internal implementation detail. we can use a couple fields from it but we probably don't want to have this struct represented on the client.

jtucholski: @PThorpe92 Thank you for clarifying. Yes, that was exactly what I was thinking.

carddev81: Made the change/push and thanks!

jtucholski (approved):
Works great and I like this approach.

PR #923, opened 2025-06-15
fix: remove loader call and replace logic with facilities per ticket 310

carddev81:
## Description of the change

Fixed operation insights bug so that facilities display in dropdown and also made sure that switching between the facilities was working. This required removing the useLoaderData call along with a bit of frontend refactor.

- **Related issues**: Link to Asana tickets [BUG: Operational Insights Doesn't Update When Switching Facilities](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210535405607307?focus=true)

### DIFF
--- a/frontend/src/Components/OperationalInsightsCharts.tsx
+++ b/frontend/src/Components/OperationalInsightsCharts.tsx
@@ -1,21 +1,13 @@
 import { useEffect, useState } from 'react';
 import useSWR from 'swr';
-import {
-    Facility,
-    FilterPastTime,
-    LoginMetrics,
-    ServerResponseOne
-} from '@/common';
+import { FilterPastTime, LoginMetrics, ServerResponseOne } from '@/common';
 import StatsCard from './StatsCard';
 import { ResponsiveContainer } from 'recharts';
 import EngagementRateGraph from './EngagementRateGraph';
 import { useAuth, canSwitchFacility } from '@/useAuth';
 import DropdownControl from './inputs/DropdownControl';
-interface Props {
-    facilities: Facility[];
-}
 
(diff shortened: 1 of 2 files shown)


PThorpe92 (approved):

frontend/src/Components/OperationalInsightsCharts.tsx
```diff
@@ -86,7 +78,7 @@ const OperationalInsights = ({ facilities }: Props) => {
                                         <option key={'all'} value={'all'}>
                                             All
                                         </option>
-                                        {facilities?.map((facility) => (
+                                        {user?.facilities?.map((facility) => (
```
jtucholski, line 81: @carddev81 I would test this but for some reason my environment isn't starting up. Will `facilities` load for facility admins? I'm looking at `auth.go` and it feels to me like the facilities map wouldn't load.

```
if claims.canSwitchFacility() {
		// if the user can see all the facilities, preload them in the claims so we
		// aren't fetching them repeatedly in every other route loader.
		if _, facilities, err := srv.Db.GetAllFacilities(1, 100); err == nil {
			traits["facilities"] = facilities
		}
	} else if !claims.canSwitchFacility() && claims.FacilityID != user.FacilityID {
		// user isn't an admin, and has alternate facility_id in the claims
		log.error("user viewing context for different facility. this should never happen")
		return newUnauthorizedServiceError()
}
```

calisio: For a facility admin, this dropdown would not be rendered. Therefore, the facility admin does not need access to the facilities.

calisio (approved):

calisio:
Will wait for @jtucholski on this merge in case he still has any hesitations.

PR #924, opened 2025-06-17
fix: remove validation for allowing past dates per ticket 305

carddev81:
## Description of the change

Removed validation from ClassManagementForm that doesn't allow past dates. 

NOTES:  
- BTW: Calendar only displays 3 months in the past 
- Potential Issue Discovered: Daylight savings time causes scheduled times to shift. I think we should fix this and it could be added this to 'Restore Event' ticket

- **Related issues**: Link to Asana tickets [Allow for classes to be created with start dates in the past](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210495081491621?focus=true)

### DIFF
--- a/frontend/src/Components/inputs/RRuleControl.tsx
+++ b/frontend/src/Components/inputs/RRuleControl.tsx
@@ -11,7 +11,6 @@ import { useAuth } from '@/useAuth';
 import {
     formatDuration,
     isEndDtBeforeStartDt,
-    isPastDate,
     timeToMinutes
 } from '../helperFunctions';
 export interface RRuleFormHandle {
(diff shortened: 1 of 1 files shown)


calisio:
Ya, calendar only shows 3 months in the past. I think we need to have a discussion if this is best or what amount of time would be best. Tagging @jtucholski since we've mentioned this before.

carddev81:
> I think there will be more places where this is relevant, such as editing an event or a series? Right now that behavior is restricted for me. Also, if I remember correctly, there might be some checks in the backend that restrict this behavior? I'm not 100% positive on that one though, but I would double check. Screenshot 2025-06-18 at 11 18 28 AM

@calisio Made change and pushed

frontend/src/Components/inputs/RRuleControl.tsx
```diff
@@ -11,7 +11,6 @@ import { useAuth } from '@/useAuth';
 import {
     formatDuration,
     isEndDtBeforeStartDt,
-    isPastDate,
```
calisio, line 14: Do you think if we don't use `isPastDate` anymore (this was the only place it was used) we should remove it from `helperFunctions` so it doesn't leave any debt, and if we need it in the future we can add it in again?

calisio (approved):
Looks good, just left one more comment !

frontend/src/Components/inputs/RRuleControl.tsx
carddev81: @calisio made the requested change and pushed 🙌

PR #927, opened 2025-06-22
feat: add Role column, disabled icons w/tooltip to user admin pg

carddev81:
## Description of the change

Added the Role column to user table and added the action icons for delete, edit, and reset password. Also added the tooltips requested when disabled. Modified the SQL query to pull all department admin users no matter the facility selected in the facility dropdown.

- [x] Add new **Role** column to the Admins page
- [x] Implement **Facility switcher** to only affect which Facility Admins are shown (not Department Admins)
- [x] Ensure **Department Admins** are the only users who see the Admins page
- [x] Department Admins should see all **Department Admins** (regardless of selected facility)
- [x] (SuperAdmin also sees all Department Admins, confirm with @jtucholski if this is an issue)
- [x] Add **action buttons** (Edit, Reset Password, Delete) and tooltips on Admins page
    - [x] Disabled Edit: Tooltip — "Edit not allowed"
    - [x] Disabled Reset Password: Tooltip — "Reset password not allowed"
    - [x] Disabled Delete: Tooltip — "Delete not allowed"

- **Related issues**: Link to Asana ticket [Clean up and address confusion on the Administrators Page](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210089681287799?focus=true)

## Screenshot(s)

image

### DIFF
--- a/frontend/src/Pages/AdminManagement.tsx
+++ b/frontend/src/Pages/AdminManagement.tsx
@@ -136,6 +136,26 @@ export default function AdminManagement() {
         return;
     };
 
+    function getUserRoleDesc(role: UserRole): string {
+        let roleDescription;
+        switch (role) {
+            case UserRole.SystemAdmin:
+                roleDescription = 'System Admin';
+                break;
+            case UserRole.DepartmentAdmin:
+                roleDescription = 'Department Admin';
+                break;
+            case UserRole.FacilityAdmin:
+                roleDescription = 'Facility Admin';
+                break;
+            case UserRole.Student:
+                roleDescription = 'Student';
+                break;
+            default:
+                roleDescription = 'Unknown';
+        }
+        return roleDescription;
+    }
     return (
         <div>
             <div className="flex flex-col space-y-6 overflow-x-auto rounded-lg p-4 px-5">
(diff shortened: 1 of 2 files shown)


frontend/src/Pages/AdminManagement.tsx
```diff
@@ -206,6 +226,7 @@ export default function AdminManagement() {
                             <tr className="grid grid-cols-5 px-4">
```
CK-7vn, line 226: Should we add grid-cols-6 here since we have 6 columns now? Seems like the column headers are off by just a tiny tiny bit. 
image

CK-7vn: Other one for the table body is on line 250, for some reason it won't let me comment on that line.

CK-7vn (commented):
Works quite well, just the headers off very very slightly, if everyones kosher with those then this is all set.

frontend/src/Pages/AdminManagement.tsx
carddev81: Thanks for catching this, didn't even notice the css. 🙌 Made change/pushed

jtucholski (approved):

PR #928, opened 2025-06-23
Update: submit button text based on create/edit mode

JamesUnlocked:
## **Submit buttons need to reflect actions and state**
* Button components now track state
* State determines text on buttons

- **Related issues**: Link to Asana ticket [Submit Button text for User and Facility modals](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210608461992709?focus=true)

## Screenshot(s)
ScreenShot 1

Screenshot 2

## Additional context
I know that we don't use forks but I don't have permissions ( as far as I can tell)  yet to just push

---
  - https://app.asana.com/0/0/1210603606220523

### DIFF
--- a/frontend/src/Pages/ClassManagementForm.tsx
+++ b/frontend/src/Pages/ClassManagementForm.tsx
@@ -337,9 +337,7 @@ export default function ClassManagementForm() {
                         <div className="w-32 pt-4">
                             <SubmitButton
                                 label={
-                                    class_id && class_id !== 'new'
-                                        ? 'Save Changes'
-                                        : 'Create Class'
+                                    isNewClass ? 'Create Class' : 'Save Changes'
                                 }
                                 errorMessage={errorMessage}
                             />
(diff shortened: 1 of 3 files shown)


frontend/src/Pages/ClassManagementForm.tsx
```diff
@@ -316,7 +316,7 @@ export default function ClassManagementForm() {
                             </label>
                         </div>
                         <div className="w-32 pt-4">
-                            <SubmitButton errorMessage={errorMessage} />
+                            <SubmitButton  errorMessage={errorMessage} label ={isNewClass ? "Create Class " : "Save Changes "}/>
```
carddev81, line 319: delete the extra space after Class and Changes

jtucholski, line 319: @JamesUnlocked Small nitpick, can you remove the extra spacing between `label` and the expression?

`<SubmitButton errorMessage={errorMessage} label={isNewClass ? "Create Class" : "Save Changes"} />`

**Also,** the submit button and cancel button are extra-wide when compared to other forms. I think that we just need to get rid of  the `w-32` class on it.

jtucholski (changes requested):
@JamesUnlocked Looks good, just two small changes on the Class Management form.

JamesUnlocked:
Thank you

PThorpe92 (approved):

PR #940, opened 2025-06-30
fix: total programs and enrollments for facility admin pull from latest date

calisio:
## Description of the change
Bug was found where the facility admins total programs was not reflected correctly in their program management page. This was due to a bug that totaled all programs for all days within the time filter. The solution was to extract the total programs and enrollments into its own query, only pulling the most recent date for these numbers.

- **Related issues**: Fixes [Asana Ticket ID-298](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210487095566664?focus=true). This should also address the issues we had with [ID-329](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210615552064401?focus=true), where the facility admin numbers were off (they are the same bug I believe, but correct if I misinterpreted the ticket(s)).

## Screenshot(s)

Below is what a super admin will see (all totals across all facilities)
Screenshot 2025-06-30 at 10 51 08 AM

Below is what a facility admin will see (total programs for their facility specifically.
Screenshot 2025-06-30 at 10 51 25 AM

## Additional context

Bug was originally thought to be displaying the totals of all facilities, but it was actually just over-calculating the number of programs across a greater time frame (rather than just pulling the most recent totals).

### DIFF
--- a/backend/src/database/programs.go
+++ b/backend/src/database/programs.go
@@ -370,10 +370,29 @@ func (db *DB) GetProgramsFacilitiesStats(args *models.QueryContext, timeFilter i
 
 func (db *DB) GetProgramsFacilityStats(args *models.QueryContext, timeFilter int) (models.ProgramsFacilitiesStats, error) {
 	var programsFacilityStats models.ProgramsFacilitiesStats
-	tx := db.WithContext(args.Ctx).Model(&models.DailyProgramFacilityHistory{}).
+	var mostRecentDate time.Time
+	if err := db.WithContext(args.Ctx).
+		Model(&models.DailyProgramFacilityHistory{}).
+		Select("MAX(date)").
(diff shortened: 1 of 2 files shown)


CK-7vn (commented):
I may be missing something, but I think we may still need to update the tooltip on the frontend, when logged in as a facility admin, it's still giving me "across all facilities" 
image

Also, another thing that I may be over looking, because I don't see the check in the db method for getting programs across all facilities, but are we supposed to be filtering archived programs out? 

Other than those 2 questions it works good!

calisio:
Both great catches and complete oversight on my part. 
I actually had noted down the archived thing (for classes & hadn't realized it for programs), but I decided it will take a bit more to fix that, so I am going to put it into a different PR. 
The tooltip should be updated on this one now though👍

PR #943, opened 2025-07-03
feat: added more descriptive text with segregated option

JamesUnlocked:
# Description of the change
## Add an Incomplete: Segregated status
### The Enrollment feature now has a Segregated option that also informs client modal that the feature counts as a "Dropped" class. This is also asserted in the DB record
https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210605185296656
## Screenshot(s)
image

image

image

## Additional context

### DIFF
--- a/backend/src/models/program_classes.go
+++ b/backend/src/models/program_classes.go
@@ -78,6 +78,7 @@ const (
 	EnrollmentIncompleteDropped          ProgramEnrollmentStatus = "Incomplete: Dropped"
 	EnrollmentIncompleteFailedToComplete ProgramEnrollmentStatus = "Incomplete: Failed to Complete"
 	EnrollmentIncompleteTransfered       ProgramEnrollmentStatus = "Incomplete: Transfered"
+	EnrollmentIncompleteSegregated       ProgramEnrollmentStatus = "Incomplete: Segregated"
 )
 
 type ProgramCompletion struct {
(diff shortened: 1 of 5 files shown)


backend/src/models/program_classes.go
```diff
@@ -78,8 +78,11 @@ const (
 	EnrollmentIncompleteDropped          ProgramEnrollmentStatus = "Incomplete: Dropped"
 	EnrollmentIncompleteFailedToComplete ProgramEnrollmentStatus = "Incomplete: Failed to Complete"
 	EnrollmentIncompleteTransfered       ProgramEnrollmentStatus = "Incomplete: Transfered"
+	EnrollmentIncompleteSegregated      ProgramEnrollmentStatus = "Incomplete: Segregated"
 )
 
+
```
CK-7vn, line 84: Super nit but can we remove these added lines here.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.20-alpine AS build
```
CK-7vn, line 1: Few things here, we essentially never mess with any of the container stuff, we don't want to delete the .env.example, and we don't want to merge the csv's either. I would do a soft reset, back to the head, this will get you back to the state of main, without reverting your changes, and then I'd just run ```git add backend/ frontend/``` I did very similar things to this when I was getting used to git, usually comes from running a ```git add .``` un-intentionally. And then once that's all updated, just force push to this branch. Once it's just outcomes.go, program_classes.go, ClassEnrollmentDetails.tsx, and common.ts, this should be just about good.

backend/src/models/outcomes.go
```diff
@@ -2,6 +2,7 @@ package models
 
 type Outcome struct {
 	DatabaseFields
+    ProgramName string `gorm:"size:255" json:"program_name"`
```
CK-7vn, line 5: You're using vs code, I think? Right? I'd try to set up "go fmt" on save in vs code, not sure how to do it, i've never used vs code, but that will help with things like this ProgramName being out of the column of the rest of the struct fields.

JamesUnlocked:
thank you for the help, I'll get right to it

jtucholski (approved):

PR #944, opened 2025-07-03
fix: user circle icon wasn't centered on resident profile

jtucholski:
## Description of the change

This change centers the user profile icon on the resident profile page

**Screenshots:**
Before
image

After
image


carddev81 (approved):
Tested good on two versions of chromium browsers and also tested good on firefox browser.

PR #947, opened 2025-07-09
Replace the Mark Attendance button button on Attendance page for dates in the future

corypride:
## Description of the change

Attendance date is in the future -  Label (muted text + Available after MM/DD)
Attendance date is in present/past and class active - Clipboard w/check mark icon + "Mark attendance" (on hover, underline)
Class is cancelled or completed - Eye icon + "View attendance" (on hover, underline)

**Related vidz or screenshot: 

https://github.com/user-attachments/assets/d9d6aa8b-991e-4430-8aaf-bf4c6c046213

---
  - https://app.asana.com/0/0/1210251192355414

### DIFF
--- a/frontend/src/Pages/ClassEvents.tsx
+++ b/frontend/src/Pages/ClassEvents.tsx
@@ -4,6 +4,7 @@ import { useUrlPagination } from '@/Hooks/paginationUrlSync';
 import {
     Attendance,
     Class,
+    SelectedClassStatus,
     ServerResponseMany,
     ServerResponseOne
 } from '@/common';
@@ -21,6 +22,7 @@ import {
     EyeIcon
 } from '@heroicons/react/24/outline';
 import moment from 'moment';
+import ULIComponent from '@/Components/ULIComponent';
 
 function toLocalMidnight(dateOnly: string): Date {
     const [year, month, day] = dateOnly.split('-').map(Number);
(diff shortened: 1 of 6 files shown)


frontend/src/Pages/ClassEvents.tsx
```diff
@@ -57,6 +63,14 @@ export default function ClassEvents() {
     const meta = data?.meta;
     const events = data?.data ?? [];
 
+    function isFutureDate(date: string): boolean {
+        const day = parseLocalDay(date);
+        console.log(day);
```
carddev81, line 68: remove this console.log here

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -13,6 +13,12 @@ import Pagination from '@/Components/Pagination';
 import { useForm } from 'react-hook-form';
 import { DateInput } from '@/Components/inputs/DateInput';
 import { isCompletedCancelledOrArchived } from './ProgramOverviewDashboard';
+import { parseLocalDay } from '@/Components/helperFunctions/formatting';
+import {
+    ClipboardDocumentCheckIcon,
+    EyeIcon
+} from '@heroicons/react/24/outline';
+import { format, parseISO } from 'date-fns';
```
carddev81, line 21: @corypride After talking with @calisio there was a decision made to use just the moment library and in the near future we are going to be removing the `date-fns`. Basically they do the same thing, but we planned on just keeping the moment library since it is a dependency to the big-react-calander.  The equivalent using moment is `moment(event.date).format('M/D')`.

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -146,38 +163,61 @@ export default function ClassEvents() {
...
+                                                blockEdits ||
+                                                event.is_cancelled;
+                                            const isFuture = isFutureDate(
+                                                event.date
+                                            );
+
+                                            if (isViewOnly) {
+                                                return (
+                                                    <>
+                                                        <div
```
carddev81, line 178: Noticed that when I click the link to go to the Event Attendance page and click 'Cancel' I get an error message. I don't think this should happen. Cancel should just take the user back to the page they came from? Also, after I cancel and go back to the page the attendance no longer exists. See video:

https://github.com/user-attachments/assets/4b89509d-41e5-4674-a893-ef9fb9020234

carddev81 (changes requested):
Noticed that when changing the status of the class to 'complete' or 'cancelled' and was attendance marked the 'Status' column no longer displays 'Marked' see pic:

Image

jtucholski:
@corypride Looks good. I refactored the logic into its own helper component.

carddev81 (commented):
@corypride I marked attendance and then set the class status as 'complete'. I would have expected that the status of the Attendance to stay as 'Marked' but it shows as 'Not Marked'? Is this correct logic.

Pic 1
Image


Pic 2
Image

calisio:
just pushed a small change that fixes alignment on the table & makes sure to use ULIComponent. **Did not change any functionality**

jtucholski:
@corypride There was a breaking change with the `completed` field that was added to the `ProgramClasses` type. It was missing a `gorm:-`. I added that in and committed but I question if we need to make all of these changes at this point or if we could have handled them in a separate issue. 

I know that the scope of this ticket increased once we started dealing with **Marked** and **Unmarked**. **Can you try and explain what approach you were taking to solve** for? The current one, adding `enrolled` plus `completed` doesn't work when someone is Incomplete.

carddev81 (approved):
@corypride and @jtucholski `events_attendance.go:94:` I've been having issues seeing cached results when I should be seeing live data. In order to correct the issue I suggest to get a new session versus a used still open session?  This will fix the current issue but I believe we should still be displaying the completed enrollments? Is this true? When clicking the View Attendance link the page is blank? 
 	
`.Session(&gorm.Session{NewDB: true})`

frontend/src/Pages/ClassEvents.tsx
```diff
@@ -81,16 +96,58 @@ export default function ClassEvents() {
...
+        if (
+            program_class?.status === SelectedClassStatus.Cancelled ||
+            event.is_cancelled
+        ) {
+            return 'Cancelled';
+        }
+        if (eventDate > today) {
+            return 'Scheduled';
+        }
+        const expectedCount =
```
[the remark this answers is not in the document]
jtucholski: @corypride & @carddev81 We can talk about this later but I want us to reconsider why we're doing this at all...or at least help me understand what it's aiming to achieve.

We're solving for an issue with the Status column so that it correclty displays logic **Marked** or **Unmarked.** I think the issue existed _before_ this pull request. In order for it to work correctly, it needs to know how many were enrolled _on each specific date_. 

The number of enrollments will change over time as students enroll/unenroll (Complete early, Withdraw/Drop/Fail/etc.). In order to know if all students were marked we'd need to know how many were enrolled on a given date (PR #959 supports this).

Maybe since this issue already existed we can remove this commit and revisit the logic for how the Status column works?

jtucholski (approved):

PR #953, opened 2025-07-12
chore: remove unused ProgramCard component per ticket 314

carddev81:
## Description of the change

Removed ProgramCard along with any other attached unused resource.

Removed the following:
- ProgramCard.tsx
- API route handler `/api/programs/{id}/save`
- Database call ToggleProgramFavorite

- **Related issues**: Link to asana ticket [Clear out ProgramCard component](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210545182051642?focus=true)


jtucholski:
@carddev81 This wasn't in the original ticket, but do you think we should remove the `program_favorites` table, the link to `[]ProgramFavorite` and the `ProgramFavorite` struct (`dashboard.go`)?

I'm not aware of future functional need but I also know that we generalized our approach toward handling favorites too.

What do you think?

carddev81:
> @carddev81 This wasn't in the original ticket, but do you think we should remove the `program_favorites` table, the link to `[]ProgramFavorite` and the `ProgramFavorite` struct (`dashboard.go`)?
> 
> I'm not aware of future functional need but I also know that we generalized our approach toward handling favorites too.
> 
> What do you think?

@jtucholski Yeah, I think that we should remove the all of the above since it is no longer used and I don't see any need in the future for it as well. I will make the necessary changes.

CK-7vn (approved):
I can't find anywhere else where this is referenced.

carddev81:
> > @carddev81 This wasn't in the original ticket, but do you think we should remove the `program_favorites` table, the link to `[]ProgramFavorite` and the `ProgramFavorite` struct (`dashboard.go`)?
> > I'm not aware of future functional need but I also know that we generalized our approach toward handling favorites too.
> > What do you think?
> 
> @jtucholski Yeah, I think that we should remove the all of the above since it is no longer used and I don't see any need in the future for it as well. I will make the necessary changes.

Made this change and pushed

backend/migrations/00053_drop_program_favorites.sql
```diff
@@ -0,0 +1,14 @@
+-- +goose Up
+-- +goose StatementBegin
+DROP TABLE IF EXISTS public.program_favorites CASCADE;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+CREATE TABLE public.program_favorites (id SERIAL NOT NULL,
```
CK-7vn, line 8: Probably should add an IF NOT EXISTS when we bring this table back.

CK-7vn (commented):
Just that IF NOT EXISTS statement.

backend/migrations/00053_drop_program_favorites.sql
carddev81: Good call, I will add this quick.

CK-7vn (approved):
Looks good!

jtucholski (approved):

PR #956, opened 2025-07-13
fix: modified PageNav component to reset pagination issue

JamesUnlocked:
# Bug: Facility Switch Retains Pagination State and Causes "No Residents" Error

## Modified the PageNav component to deal with the page reset when changing facility.
(https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210707681181866)

## Screenshot(s)

[Video walk through of process.
](https://drive.google.com/file/d/1tYKwXQmDYBEZxVb9amiPnKiutvkOZHJ5/view?usp=sharing)

## Additional context

### DIFF
--- a/frontend/src/Components/PageNav.tsx
+++ b/frontend/src/Components/PageNav.tsx
@@ -55,8 +55,11 @@ export default function PageNav({
         );
         if (resp.success) {
             const params = new URLSearchParams(window.location.search);
-            params.set('page', '1');
-            window.location.href = `${window.location.pathname}?${params.toString()}`;
+            if (params.get('page') !== null) {
+                params.set('page', '1');
+            }
+            const paramsString = params.size > 0 ? '?' + params.toString() : '';
+            window.location.href = window.location.pathname + paramsString;
         }
     };
 
--- a/frontend/src/Components/PageNav.tsx
+++ b/frontend/src/Components/PageNav.tsx
@@ -55,9 +55,8 @@ export default function PageNav({
             {}
         );
         if (resp.success) {
-            // Navigate to the first page of the residents list
             navigate('/residents?page=1');
-            // If you need to reload to ensure all data is fresh, you can do so after a short delay
+
             setTimeout(() => {
                 window.location.reload();
             }, 100);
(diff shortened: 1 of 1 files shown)


calisio (changes requested):
Unfortunately, with this change, every time we use the facility dropdown in the top right corner, we will be navigated to the `/residents` page, even if we change facilities in a completely irrelevant page. For example, try changing the facility on the `/knowledge-insights` page, and you will see the issue.

JamesUnlocked:
Damn, you're right. Thought I had something fixed

JamesUnlocked:
I also used the timeout F(x) to avoid race conditions that may spring up whenever using asynchronous F(x)'s

calisio:
I think ideally we would want to stay away from changing the `navigate` behavior on this ticket, especially because it is dealing with a hyper-specific case (only doing the reset navigate when its on the residents page) in a component that has little to do with the page.

I think the core of the problem has to do with the URL being set to `/residents?per_page=10&page=3`, and that query param needs to be reset to page=1 when you switch facilities. The problem probably has to do with how keep the history navigation, which was introduced in [this PR](https://github.com/UnlockedLabs/UnlockEdv2/pull/815/files#diff-24f1deb599a56ce4ef67e2d357be7d29d1258b57ff919e3fbdbaf5582e85e329). 

I think the issue lies in the following block from `StudentManagement.tsx`:
```
const {
        page: pageQuery,
        perPage,
        setPage: setPageQuery,
        setPerPage
    } = useUrlPagination(1, 20);
```
Where `useUrlPagination()`, found in `paginationUrlSync.tsx` is saving the history of the page query. Remember, ideally we don't want to change the general-ness of components to handle for one-off cases. But, in this case, I think that we would experience a similar issue anywhere else we use the hook `useUrlPagination();` so, it is probably worth fixing within this component. We would probably need to handle the scenario "if the facility changes, reset the pagination to the first page".

JamesUnlocked:
I understand about maintaining a general component and agree. Can you
please clarify which component you mean when you say "it is probably worth
fixing within this component." ( StudentManagement.tsx: ||
paginationUrlSync.tsx).
 I apologize for my confusion.

calisio:
No apologies needed! I meant inside of `paginationUrlSync.tsx`

JamesUnlocked:
I have to say that although I pushed this, this was the work of Carolina. She was extremely helpful with this process and I would be remiss if tried to take any credit for this fix.

carddev81 (approved):
Looks/Tested Great!

calisio (approved):

PR #957, opened 2025-07-14
chore: move cal and event related types to their own file per 262

carddev81:
## Description of the change

Refactored the Calendar and Event related types by moving them into their own file for better organization within the code base.

Removed/Refactored the following:
- Removed the following unused event/calendar related types
    - EventCalendar
    - Month
    - Day
    - DayData
    - StudentCalendar
    - EventDate
    - Event
- Refactored these types into their own file:
    - FacilityProgramClassEvent
    - ProgramClassEvent
    - ProgramClassEventOverride
    - ProgramClassEventAttendance
    - ClassEventInstance
    - ShortCalendarEvent
    - CalendarEvent

- **Related issues**: Link to asana ticket [Refactor Calendar and Event-Related types in common.ts](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210329763082319?focus=true)

- Removed the following components - Added to [Wiki](https://github.com/UnlockedLabs/UnlockEdv2/wiki/Deprecated-Features-and-Components):
    - WeeklyScheduleTable
    - EditEventForm


calisio (changes requested):
I think ideally this ticket would go a step further and would refactor the types you have moved into `events.ts` and condense them down. There are many interfaces that are quite repetitive, such as `Event` and `CalendarEvent` that are essentially the same +/- some fields (those two are actually identical). There should only be a handful of types in this file. I'd suggest exploring having some of the fields be optional and adding in checks where they are used if needed, or if needed using `extends` rather than repeating the same fields several times.

carddev81:
> I think ideally this ticket would go a step further and would refactor the types you have moved into `events.ts` and condense them down. There are many interfaces that are quite repetitive, such as `Event` and `CalendarEvent` that are essentially the same +/- some fields (those two are actually identical, just written in a different order). There should only be a handful of types in this file. I'd suggest exploring having some of the fields be optional and adding in checks where they are used if needed, or if needed using `extends` rather than repeating the same fields several times.

I'll go back and rename/condense.

backend/src/database/class_events.go
```diff
@@ -526,7 +527,7 @@ func (db *DB) GetFacilityCalendar(args *models.QueryContext, dtRng *models.DateR
 		}
 
 		linkedOverridesMap := getLinkedOverrideEventMap(event, overrides)
-
+		fmt.Println(">>>>>>>>>>>>>>>>>>>>>linkedOverridesMap", linkedOverridesMap)
```
CK-7vn, line 530: Just tagging this so you don't miss it.

CK-7vn (commented):
I know you're doing a refactor, so i'll redo this as soon as you push that.

carddev81:
> ng a refactor, so i'll redo this as soon as you push tha



> I know you're doing a refactor, so i'll redo this as soon as you push that.

fixing to push now

carddev81:
> > I think ideally this ticket would go a step further and would refactor the types you have moved into `events.ts` and condense them down. There are many interfaces that are quite repetitive, such as `Event` and `CalendarEvent` that are essentially the same +/- some fields (those two are actually identical, just written in a different order). There should only be a handful of types in this file. I'd suggest exploring having some of the fields be optional and adding in checks where they are used if needed, or if needed using `extends` rather than repeating the same fields several times.
> 
> I'll go back and rename/condense.

Made changes/pushed

carddev81:
> I am happy with this PR and would be okay with this getting merged in as it stands now.
> 
> I am still struggling to recognize when we would use one interface over another, which I think is why this became an issue in the first place. I would love to hear how people feel if we were to include a brief on-line description on top of each interface, explaining when it is used. For example, I am struggling to understand the difference between `FacilityProgramClassEvent` and `ProgramClassEvent`? Not sure if it's just me.

@calisio I added a one line description to the top of each interface to help describe what each is used for.

jtucholski:
I'm good with these changes. 

Adding the TS comment above the type helps no matter what and forces us to always ask if it's worth extending a type, updating an existing one, or creating a new one.

jtucholski (approved):

PR #962, opened 2025-07-23
fix: knowledge center search modal sizing

CK-7vn:
## Description of the change
This PR fixes the issue with the search functionality of the knowledge center and how the user, particularly on a secure book is/was unable to reach the bottom of the search results, or modal, disallowing them the ability to click on subsequent pages of results. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210825285656046?focus=true

## Screenshot(s)
27" Monitor
image

Laptop Monitor Chrome
image

Chrome Securebook dimensions
image

Firefox Securebook dimensions
image

## Additional context
Prior to this PR the LibrarySearchModal had a set modal height, and was placed inside of a dialog element, which as a result stopped the modal from being able to scroll relative to it's parent page, essentially being constrained by the viewport. With a header of approximately 125px, and a footer of approximately 80px, together this leaves like 500px left for the actual flexible content area, and on a 768px screen (secure book) it just doesn't add up to being enough room, particularly on these odd, and weird resolutions. Second to that, my-auto was stopping the flex items from shrinking smaller than the content they were wrapping, so even though we were using auto in terms of overflow and height, the container was trying to be as tall as all of the content combined.  What this PR essentially does, is locks in the header and footer dimensions, but allows the content area to be dynamic, making sure that the overflow doesn't cause the modal to reach outside of appropriate dimensions.

### DIFF
--- a/frontend/src/Components/LibrarySearchResultsModal.tsx
+++ b/frontend/src/Components/LibrarySearchResultsModal.tsx
@@ -253,8 +253,8 @@ const LibrarySearchResultsModal = forwardRef<
 
     return (
         <dialog ref={ref} className="modal" onClose={handleCloseModal}>
-            <div className="w-[1200px] h-[700px] bg-background rounded-lg shadow-lg flex flex-col">
-                <div className="sticky top-0 bg-background z-50 p-4 border-b border-grey-2 rounded-t-lg">
+            <div className="w-[min(90vw,1200px)] h-[min(85vh,700px)] bg-background rounded-lg shadow-lg flex flex-col">
+                <div className="flex-shrink-0 bg-background z-50 p-4 border-b border-grey-2 rounded-t-lg">
                     <div className="flex items-center gap-4">
                         <div className="flex flex-col">
                             <h2 className="text-2xl">{searchResults.title}</h2>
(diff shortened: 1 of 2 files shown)


jtucholski:
@CK-7vn This is already looking better. I like the responsive adjustment.

I noticed a few things while looking at this and wanted to see if we also try incorporating some tweaks:
* The `X` to close the dialog is now in the top-right corner of the screen. Can we move it back into the window?
* What do you think about bumping the height to `-[min(90vh,900px)]`? I think for larger screens they can have the extra vertical space and for smaller screens the dialog behavior will make it clear that this window is _on top_. 

I also looked at Google search results a bit and then tried experimenting with the `LibrarySearchResultCard` component. Here is what I tried doing with it
* reduced the icon size
* made the main page and library name more visually distinct
* tried to truncate the description to 60 words while preserving the bolded parts

I thought truncating might help fit more results on the screen but then it also felt `body-small` was too small for the font size so I bumped that up. Ultimately it doesn't fit more on screen but does use the space a bit more.

Here's what it might look like on the full size display and the smaller 1366 x 768 display. What do you think? Do you think it can go with the PR? 

image

image

calisio (changes requested):
Quick first review on this:
1. I dont _love_ that on default if there is no search in the input, the modal is super small. I think it should be a fixed width for this & we adjust this based on screen size? If we didn't want the entire modal as a fixed height, we alternatively could just make the "Enter a Search" a fixed height so it takes up more room (adjusted based on screen size. [Tailwind has this built in pretty easily](https://tailwindcss.com/docs/responsive-design#overview))
Screenshot 2025-07-28 at 2 27 44 PM

2. Because we no longer have a fixed height, my modal is expanding off the page, so I am unable to see the pagination. Are other people also experiencing this? I think this might be part of the original issue?
Screenshot 2025-07-28 at 2 28 09 PM

calisio:
Ah, I checked out the commit right before `commit 821192c42b6735d086e760c5ef21ac98a3bd6989` (your last commit that adds in `relative`), and I see how you originally fixed it. It seems like "relative" might be the part that messed it up?

CK-7vn:
@calisio I think both of those issues were a result of an extra "]"........ 😢 I removed it..and literally golden now..
image

jtucholski:
@CK-7vn ADo you need to push your fix still? I can't scroll the search window either. 

image

CK-7vn:
I think it's good, everything is working fine for me today. I think the issue you were having was related to the non-alpha character in your search, and it must have have been an issue with how search makes suggestions when theres special characters in it, and the request getting messed up somehow. That's at least on the surface what I would assume, because I can't get it to crash out like that for me, it just says there's an error whenever I search with a special character.

PR #965, opened 2025-07-28
fix: fix stuck tour on step 3 & other misc tour bugs

calisio:
## Description of the change

Fixed the tour getting stuck on step 3. I made sure to go both forward and backwards on all of the steps, but let me know if I missed something & its not working as expected again. There were some other random bugs that kept appearing with the tour, and I realized it was likely because it was setting tour states without checking if the tour was active in the first place. I made sure to wrap all tour things with a "if tourState is active" to only apply changes when necessary.

- **Related issues**: [Closes asana ID-365](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210825285656036?focus=true)

### DIFF
--- a/frontend/src/Components/LibraryLayout.tsx
+++ b/frontend/src/Components/LibraryLayout.tsx
@@ -6,7 +6,7 @@ import useSWR from 'swr';
 import Pagination from './Pagination';
 import { useLocation, useOutletContext } from 'react-router-dom';
 import LibrarySearchResultsModal from '@/Components/LibrarySearchResultsModal';
-import { initialTourState, useTourContext } from '@/Context/TourContext';
+import { useTourContext } from '@/Context/TourContext';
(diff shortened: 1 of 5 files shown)


corypride (changes requested):
Looks good overall, saw a slight bug , a UX skip between step 6 and step 11, even though your steps array is sequential. 

https://github.com/user-attachments/assets/4980d2a0-bbe2-47bc-822d-94edffe23aee

corypride (approved):
Thanks for the clarification, @calisio. I went back and tested it again after adding libraries into my local instance, and the full step flow now appears correctly without skipping. Confirmed that the UX skip between step 6 and 11 only occurs when no libraries are available, which makes sense. Good to go from my side

jtucholski (approved):
Looks good, and went through the whole flow. Didn't run into issues.

PR #966, opened 2025-07-30
fix: modified the reason field in the UI for attendance to be required

JamesUnlocked:
## Description of the change

## The reason textbox is now required and clears whenever one switches back to "Present"
[Require the reason field in the UI for Excused/Unexcused attendance](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210842346686462) &
[Clear out notes text input when user selects present on attendance page ](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210801792172853)
## Screenshot(s)

https://github.com/user-attachments/assets/68e8d68a-cd3a-4304-9437-d2e5b7984073

## Additional context

### DIFF
--- a/frontend/src/Components/inputs/TextInput.tsx
+++ b/frontend/src/Components/inputs/TextInput.tsx
@@ -39,7 +39,7 @@ export function TextInput({
     onChange,
     placeholder,
     inputClassName,
-    errorTextAlign = 'center'
+    errorTextAlign
 }: TextProps) {
     const options = {
         required: {
@@ -85,7 +85,7 @@ export function TextInput({
                         ? 'text-center'
                         : errorTextAlign === 'right'
                           ? 'text-right'
-                          : ''
+                          : 'text-left'
                 }`}
             >
                 {errors[interfaceRef]?.message as string}
--- a/frontend/src/Components/inputs/TextInput.tsx
+++ b/frontend/src/Components/inputs/TextInput.tsx
@@ -39,7 +39,7 @@ export function TextInput({
     onChange,
     placeholder,
     inputClassName,
-    errorTextAlign = 'left'
+    errorTextAlign = 'center'
 }: TextProps) {
     const options = {
         required: {
--- a/frontend/src/Components/inputs/TextInput.tsx
+++ b/frontend/src/Components/inputs/TextInput.tsx
@@ -77,7 +77,7 @@ export function TextInput({
                 }}
                 placeholder={placeholder}
             />
-            <div className="text-error text-sm">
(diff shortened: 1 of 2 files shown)


frontend/src/Pages/EventAttendance.tsx
```diff
@@ -358,8 +351,10 @@ export default function EventAttendance() {
                                                         : ''
                                                 }
                                                 interfaceRef={`note_${row.user_id}`}
-                                                required={false}
-                                                length={500}
+                                                required={
+                                                    row.attendance_status !==
```
corypride, line 355: https://github.com/user-attachments/assets/7c858c87-8755-47aa-9fed-1717e802eb35

@JamesUnlocked ,
I dont think that we should be forced to add a note for rows you didn’t actually interact with.

right now  the form is rendering note fields as **required** based solely on:

```tsx
required={row.attendance_status !== Attendance.Present}
```

…but this doesn’t account for whether the user *intended* to change that row.

---

I believe that  you only want to enforce validation on `note_${user_id}` **if**:

- The row has been **modified** (e.g. `selected: true`)
- **AND** `attendance_status !== Attendance.Present`

---

### 💡 Suggested Change:

Update the `required` condition to:

```tsx
required={row.selected && row.attendance_status !== Attendance.Present}
```

This ensures that validation is only applied to fields the user has actively interacted with, preventing unnecessary errors on untouched rows.

corypride (changes requested):
looks fine overall, I left a comment on a small logic bug, and once that is resolved this should be good to go.

JamesUnlocked:
# I just force updated this PR.

frontend/src/Components/inputs/TextInput.tsx
```diff
@@ -77,7 +77,7 @@ export function TextInput({
                 }}
                 placeholder={placeholder}
             />
-            <div className="text-error text-sm">
+            <div className="text-error text-sm text-center">
```
corypride, line 80: "Great approach! However, we should consider whether this component is used elsewhere in the application and if we want it to be centered in all instances. If we prefer not to apply this globally, we can try something like adding an optional prop and using a template literal with a ternary expression to conditionally apply it to the className."

carddev81 (changes requested):
Scenario: Select Excused | Click Save Attendance (is required is displayed) | Select Present 

In this scenario the 'is required' message is still visible. When selecting Present the is required should clear out. I believe you had this logic in place before? clearErrors?

corypride (approved):
@JamesUnlocked works well.

PR #968, opened 2025-07-31
Correctly dereference pointer values in change log generation

corypride:
## Description of the change

Fixes an issue in the change log audit system where pointer values (e.g., *string, *int) were being logged as memory addresses instead of actual values.

What changed: 

Added derefToString() helper function to safely dereference reflect.Value fields during audit comparison.

Updated GenerateChangeLogEntries to use derefToString instead of fmt.Sprintf("%v", ...).

Why it matters
Previously, audit entries for pointer fields would show raw pointer addresses like 0xc0004d3080. This change ensures consistent, human-readable logs for all supported field types.

Asana Task: 400
🔗 [View in Asana](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210839425550614)

### DIFF
--- a/backend/src/models/change_log_entry.go
+++ b/backend/src/models/change_log_entry.go
@@ -36,14 +36,16 @@ func NewChangeLogEntry(tableName, fieldName string, oldValue, newValue *string, 
 }
 
 func derefToString(v reflect.Value) string {
-	if v.Kind() == reflect.Ptr && !v.IsNil() {
-		return fmt.Sprint(v.Elem().Interface())
-	}
-	if v.Kind() == reflect.Ptr && v.IsNil() {
-		return ""
+
+	for v.Kind() == reflect.Ptr {
+		if v.IsNil() {
+			return ""
+		}
+		v = v.Elem()
 	}
 	return fmt.Sprint(v.Interface())
 }
+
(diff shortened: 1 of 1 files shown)


backend/src/models/change_log_entry.go
```diff
@@ -35,6 +35,15 @@ func NewChangeLogEntry(tableName, fieldName string, oldValue, newValue *string,
...
 
+func derefToString(v reflect.Value) string {
+	if v.Kind() == reflect.Ptr && !v.IsNil() {
+		return fmt.Sprint(v.Elem().Interface())
+	}
+	if v.Kind() == reflect.Ptr && v.IsNil() {
+		return ""
+	}
+	return fmt.Sprint(v.Interface())
+}
```
CK-7vn, line 46: We could use a for loop here something like this 
```
	for v.Kind() == reflect.Ptr {
		if v.IsNil() {
			return ""
		}
		v = v.Elem()
	}
	return fmt.Sprint(v.Interface())
```
And that would de-reference multiple pointer layers, while avoiding 2 if statements, slightly more concise, but also not entirely necessary. Since we're only using this function right here, and the pointer level is explicit, we don't necessarily need the multiple layer redundancy.

CK-7vn (approved):
Sweet deal my friend!

carddev81 (approved):
Works great! Looks great!

PR #972, opened 2025-08-05
feat: add warning banner on non offered facility programs

calisio:
## Description of the change

On a program that is not offered at the facility, there is a warning banner at the top of the screen notifying the user.

- **Related issues**: Link to related Asana ticket that this closes: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210752575902065?focus=true

## Screenshot(s)
Screenshot 2025-08-05 at 3 24 55 PM
Screenshot 2025-08-05 at 3 25 46 PM


carddev81 (approved):
@calisio Looks great and functions as expected. The only question is the color of the banner doesn't match up with the banner color within the screenshot in Asana ticket?

calisio:
That was complete oversight on my end, I did not see that follow up comment with the design!
Let me fix it and the icon also, and then get that pushed. I reviewed with Josh about coloring but I like the light yellow better (the ticket said bg-warning, so that's what I used, but I am going to go back and use pale yellow).

calisio:
@carddev81 Fixed :) (except I made the icon the color `text-body` for accessibility standards sake to make sure it has enough color contrast).

PR #973, opened 2025-08-05
fix: add logic to display enrollment status options based on class status per ticket 394

carddev81:
## Description of the change

Added the necessary frontend logic to do the following:
- allow the selection of 'Cancelled' when class status is 'Scheduled'
- when class status is 'Active' the status 'Cancelled' will not be an enrollment option

Added the necessary backend logic to do the following:
- rejects the user's request if there is an attempt to set any status that is not allowed.

### Front-end
- [X] UI filters the status list based on the current class status.
- [X] It is impossible for the user to select or submit an invalid enrollment status through the interface.

### Back-end
- [X] The backend (`handleUpdateProgramClassEnrollments`) rejects any request that attempts to set an enrollment status not allowed for the current class status.

- **Related issues**: Link to related Asana ticket is: [Limit Enrollment Status Options for Scheduled & Active Classes](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210839726305327?focus=true)

## Screenshot(s)

[statuses.webm](https://github.com/user-attachments/assets/f1a26a59-a062-4cce-a158-71416895886b)

## Additional context
NAVIGATION: Login (Admin) | Click Programs | Click Program | Select Class (Active/Scheduled) | Select Enrollment Tab | Make sure when a class is Scheduled you cannot see any other selection other than 'Cancelled' for a resident when the class is Active make sure you can't select 'Cancelled'

### DIFF
--- a/backend/src/handlers/class_enrollments.go
+++ b/backend/src/handlers/class_enrollments.go
@@ -114,13 +114,6 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r 
 	if err != nil {
 		return newInvalidIdServiceError(err, "class enrollment ID")
 	}
-	class, err := srv.Db.GetClassByID(classId)
-	if err != nil {
-		return newDatabaseServiceError(err)
-	}
-	if class.CannotUpdateClass() {
-		return newBadRequestServiceError(err, "cannot perform action on class that is completed cancelled or archived")
-	}
 	enrollment := struct {
 		EnrollmentStatus string  `json:"enrollment_status"`
 		UserIDs          []int   `json:"user_ids"`
@@ -131,6 +124,13 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r 
 	}
 	if enrollment.EnrollmentStatus == "" {
 		return newInvalidIdServiceError(errors.New("enrollment status is required"), "enrollment status")
+	}
+	class, err := srv.Db.GetClassByID(classId)
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	if class.CannotUpdateClassWithEnrollment(enrollment.EnrollmentStatus) {
+		return newBadRequestServiceError(err, "cannot perform action on class that is completed, cancelled, archived, class is scheduled and enrollment is not cancelled, or class is active and enrollment is cancelled")
 	}
 	switch enrollment.EnrollmentStatus {
 	case "Completed":
--- a/backend/src/handlers/class_enrollments.go
+++ b/backend/src/handlers/class_enrollments.go
@@ -115,9 +115,9 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r 
 		return newInvalidIdServiceError(err, "class enrollment ID")
 	}
 	enrollment := struct {
-		EnrollmentStatus string  `json:"enrollment_status"`
-		UserIDs          []int   `json:"user_ids"`
-		ChangeReason     *string `json:"change_reason,omitempty"`
+		EnrollmentStatus models.ProgramEnrollmentStatus `json:"enrollment_status"`
+		UserIDs          []int                          `json:"user_ids"`
+		ChangeReason     *string                        `json:"change_reason,omitempty"`
 	}{}
 	if err := json.NewDecoder(r.Body).Decode(&enrollment); err != nil {
 		return newJSONReqBodyServiceError(err)
--- a/backend/src/handlers/class_enrollments.go
+++ b/backend/src/handlers/class_enrollments.go
@@ -130,7 +130,7 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r 
 		return newDatabaseServiceError(err)
 	}
 	if class.CannotUpdateClassWithEnrollment(enrollment.EnrollmentStatus) {
-		return newBadRequestServiceError(err, "cannot perform action on class that is completed, cancelled, archived, class is scheduled and enrollment is not cancelled, or class is active and enrollment is cancelled")
+		return newBadRequestServiceError(err, "cannot perform action due to invalid class or enrollment status")
 	}
 	switch enrollment.EnrollmentStatus {
 	case "Completed":
--- a/frontend/src/Pages/ClassEnrollmentDetails.tsx
+++ b/frontend/src/Pages/ClassEnrollmentDetails.tsx
@@ -160,7 +160,7 @@ export default function ClassEnrollmentDetails() {
                 // If one or more users are selected with the check-boxes, then they are going to be
                 // 'Graduated'. If selectedResidents is empty, that means the Dropdown is being used
                 // to change an individual status inline.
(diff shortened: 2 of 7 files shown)


backend/src/handlers/class_enrollments.go
```diff
@@ -132,6 +125,13 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r
 	if enrollment.EnrollmentStatus == "" {
 		return newInvalidIdServiceError(errors.New("enrollment status is required"), "enrollment status")
 	}
+	class, err := srv.Db.GetClassByID(classId)
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	if class.CannotUpdateClassWithEnrollment(enrollment.EnrollmentStatus) {
+		return newBadRequestServiceError(err, "cannot perform action on class that is completed, cancelled, archived, class is scheduled and enrollment is not cancelled, or class is active and enrollment is cancelled")
```
jtucholski, line 133: @carddev81 This seems verbose. Do you think this full message is necessary? What about, "Cannot perform action due to invalid class or enrollment status."

jtucholski (approved):

backend/src/handlers/class_enrollments.go
carddev81: I agree it is verbose :).  Just changed/pushed these up

backend/src/handlers/class_enrollments.go
```diff
@@ -114,13 +114,6 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r
...
 	}
-	class, err := srv.Db.GetClassByID(classId)
-	if err != nil {
-		return newDatabaseServiceError(err)
-	}
-	if class.CannotUpdateClass() {
-		return newBadRequestServiceError(err, "cannot perform action on class that is completed cancelled or archived")
-	}
 	enrollment := struct {
 		EnrollmentStatus string  `json:"enrollment_status"`
```
CK-7vn, line 118: I know you didn't mess with this, but what's your opinion on switching this from the string to "models.ProgramEnrollmentStatus" ? I think we should move away from stringly typed stuff whenever possible and try to keep as much of the codebase as type safe as we can.

frontend/src/Pages/ClassEnrollmentDetails.tsx
CK-7vn, line 1: On line 286, 
```    
 {selectedResidents.length > 0 && (
                        <button
                            disabled={blockEdits}
                            className="button"
                            onClick={handleOpenModalGraduate}
                        >
                            Graduate Selected
                        </button>
                    )}
```
We should just add a check to be safe and make sure the class is active. so essentially
```    
  {selectedResidents.length > 0 && clsInfo?.status === SelectedClassStatus.Active && (
                        <button
                            disabled={blockEdits}
                            className="button"
                            onClick={handleOpenModalGraduate}
                        >
                            Graduate Selected
                        </button>
                    )}
```

frontend/src/Pages/ClassEnrollmentDetails.tsx
```diff
@@ -150,21 +154,33 @@ export default function ClassEnrollmentDetails() {
...
-                change_reason: reasonText?.trim()
-            })
-        });
+        const resp = await API.patch(
+            `program-classes/${class_id}/enrollments`,
+            {
+                // If one or more users are selected with the check-boxes, then they are going to be
+                // 'Graduated'. If selectedResidents is empty, that means the Dropdown is being used
+                // to change an individual status inline.
+                enrollment_status: changeStatusValue?.status ?? 'completed',
```
CK-7vn, line 163: this 'completed' isn't case correct, I think we can just remove it, but at the very least we should make sure its uppercase.

frontend/src/Pages/ClassEnrollmentDetails.tsx
CK-7vn, line 1: Line 275: Cancelled is spelled with one l, Canceled. Just for reference. Which, I did know, is also technically a correct spelling. 
```
"Canceled" and "cancelled" are both correct past tense forms of the verb "cancel," but they are used in different regions. "Canceled" (with one "l") is more common in American English, while "cancelled" (with two "l's") is more common in British English. Both spellings are acceptable in their respective regions. 
``` 
Very interesting, very interesting, so I suppose either spelling is fine, we should just make sure they line up (also not on you, just noting it here).

CK-7vn (commented):
Just trivial stuff really, I think we should try and move away from any string to string comparison and try to keep what we can type safe. Works as planned though! Also, it makes me un-easy that we're duplicating the EnrollmentStatusOptions in ClassEnrollmentDetails on line 203, I feel like there's a better way that we can inherit from the EnrollmentStatus enum in common. I think we could deduplicate the EnrollmentStatusOptions, and also clean up getEnrollmentStatusOptions with something like this 
```
const enrollmentStatusLabel: Record<EnrollmentStatus, string> = {
  [EnrollmentStatus.Enrolled]: 'Enrolled',
  [EnrollmentStatus.Cancelled]: 'Cancelled',
  [EnrollmentStatus.Completed]: 'Completed',
  [EnrollmentStatus.Withdrawn]: 'Incomplete: Withdrawn',
  [EnrollmentStatus.Dropped]: 'Incomplete: Dropped',
  [EnrollmentStatus.Segregated]: 'Incomplete: Segregated',
  [EnrollmentStatus['Failed To Complete']]: 'Incomplete: Failed to Complete',
  [EnrollmentStatus.Transfered]: 'Incomplete: Transfered',
}

const SCHEDULED_ALLOWED: EnrollmentStatus[] = [EnrollmentStatus.Cancelled]
const ACTIVE_ALLOWED: EnrollmentStatus[] = Object.values(EnrollmentStatus).filter(
  s => s !== EnrollmentStatus.Cancelled
)

const getEnrollmentStatusOptions = (): Record<string, string> => {
  const allowed =
    clsInfo?.status === SelectedClassStatus.Scheduled
      ? SCHEDULED_ALLOWED
      : clsInfo?.status === SelectedClassStatus.Active
      ? ACTIVE_ALLOWED
      : (Object.values(EnrollmentStatus) as EnrollmentStatus[])

  return Object.fromEntries(allowed.map(s => [enrollmentStatusLabel[s] ?? s, s]))
}
```

But, Typescript is far from being my friend, so maybe I'm thinking of this the wrong way, don't hesitate to tell me otherwise if I am.

frontend/src/Pages/ClassEnrollmentDetails.tsx
carddev81: @jtucholski & @CK-7vn This makes sense, but is this a bug that is already being worked on, just want to be sure? Or can I fix this here?

carddev81: @CK-7vn The line I changed was, just double checking this is the one you were referencing

`Cancelled: 'incomplete: cancelled'`

carddev81:
@CK-7vn Made requested changes.

frontend/src/Pages/ClassEnrollmentDetails.tsx
carddev81: made change/pushed

PR #977, opened 2025-08-07
fix: add proper validation to attendance endpoints

CK-7vn:
## Description of the change
This PR makes sure there is proper validation for the endpoints related to attendance, in particular when adding and or deleting attendance. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210871477181002?focus=true

## Additional context
I'm not overly comfortable with the so much happening inside of a loop in the add attendance method, but I currently can't think of a clever way around it. I think we could probably get away with just checking one of the attendances date, and as long as it's not cancelled safely assume that none of the rest of the array is cancelled, but then that begs the question, "what happens if someone figures that out, gets an admin cookie, and just starts sending curl requests, they could manipulate all the attendance they want". Which is true, it's highly unlikely, but it's true, they could. I think given that it's highly unlikely this array will ever be much larger than 20 residents/users at a time, I think the load is manageable, and nothing to worry about. 
- Added support in the testing client for validating response bodies. 
- Fixed JSON parsing for 204 no content responses 
- Decided to perform cancelled date validation before enrollment validation because a cancelled date will automatically apply to everyone where as enrollment validation can be case by case 
- 409 for cancelled date as this indicates that there is a resource state conflict 
- 400 for enrollment issues as this indicates a bad request /  client issue, and I felt as though a non-enrolled user would be a client issue, or more than likely a malicious actor or bug as we take steps to prevent un-enrolled users from appearing in the attendance tracking event page.

### DIFF
--- a/backend/src/handlers/events_attendance_handler.go
+++ b/backend/src/handlers/events_attendance_handler.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"UnlockEdv2/src/models"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strconv"
 	"time"
@@ -46,6 +47,12 @@ func (srv *Server) handleAddAttendanceForEvent(w http.ResponseWriter, r *http.Re
 	now := time.Now().In(time.Local)
 	endOfToday := time.Date(now.Year(), now.Month(), now.Day(), 23, 59, 59, 999999999, time.Local)
 
+	args := srv.getQueryContext(r)
+	overrides, err := srv.Db.GetProgramClassEventOverrides(&args, []uint{uint(eventID)}...)
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+
 	for i := range attendances {
 		if attendances[i].Date == "" {
 			attendances[i].Date = time.Now().Format("2006-01-02")
@@ -59,6 +66,20 @@ func (srv *Server) handleAddAttendanceForEvent(w http.ResponseWriter, r *http.Re
 		if parsed.After(endOfToday) {
 			return writeJsonResponse(w, http.StatusUnprocessableEntity, "attempted attendance date in future")
 		}
+
+		if isDateCancelled(overrides, attendances[i].Date) {
+			return writeJsonResponse(w, http.StatusConflict, "cannot record attendance for cancelled class date")
+		}
+
+		var enrollment models.ProgramClassEnrollment
+		enrollmentErr := srv.Db.Where("class_id = ? AND user_id = ? AND enrollment_status = ?",
+			classId, attendances[i].UserID, models.Enrolled).
+			First(&enrollment).Error
+		if enrollmentErr != nil {
+			return writeJsonResponse(w, http.StatusBadRequest,
+				fmt.Sprintf("user %d is not enrolled in class %d", attendances[i].UserID, classId))
+		}
+
 		attendances[i].EventID = uint(eventID)
 	}
 	if err := srv.Db.LogUserAttendance(attendances); err != nil {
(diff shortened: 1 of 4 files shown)


backend/src/handlers/events_attendance_handler.go
```diff
@@ -59,6 +66,20 @@ func (srv *Server) handleAddAttendanceForEvent(w http.ResponseWriter, r *http.Re
...
+		}
+
+		var enrollment models.ProgramClassEnrollment
+		enrollmentErr := srv.Db.Where("class_id = ? AND user_id = ? AND enrollment_status = ?",
+			classId, attendances[i].UserID, models.Enrolled).
+			First(&enrollment).Error
+		if enrollmentErr != nil {
+			return writeJsonResponse(w, http.StatusBadRequest,
+				fmt.Sprintf("user %d is not enrolled in class %d", attendances[i].UserID, classId))
+		}
```
jtucholski, line 81: @CK-7vn You could extract the user IDs and run a query against the enrollments table to get the count of records with the provided class ID and list of user IDs and compare that to the incoming request. That would allow you to give one query up front to check for enrollments and save having to run a query for every record being recorded. 

Something that sticks out here that I don't see often in the codebase is the usage of `.Where` within the handler. Most of the time we push that down into the `database` layer. It's a layer of complexity but maybe it brings some consistency?

backend/src/handlers/events_attendance_handler.go
jtucholski, line 1: More of a discussion to make sure I'm not missing something now that I realize the date is potentially included with every incoming attendance record. 

Given the changes you recently made in #976 where we check to see if the user has `enrolled_at` I _think_ we want to keep looking to see if they have a status of `Enrolled` (which you are)  because then we would be allowed to change their attendance record. If we relied on `enrolled_at` and `enrollment_ended_at` then we might be opening ourselves up to alter attendance records _after_ they complete the class, which doesn't seem like it should be possible.

If the above is true, then do we need to address that scenario on the front-end where if we open attendance for an event that has taken place and allow the admin to still:
* modify attendance for students enrolled in the class AND 
* block attendance for students who've finished

I'm not sure and am looking for your thoughts on this.

carddev81: @jtucholski & @CK-7vn I agree to check both the “Enrolled” status and the enrollment dates. That way, we don’t accidentally allow attendance edits for residents who have already finished the class. It also makes sense to restrict admin editing on the front-end, so only current students’ (enrolled) attendance can be modified.

backend/src/handlers/events_attendance_handler.go
CK-7vn: Agreed on the handler. Wasn't even thinking. I will move that down to the database layer.

backend/src/handlers/events_attendance_handler.go
jtucholski: @CK-7vn No, not necessary. I can make a separate ticket once we finish this since that was the scope of this one.

jtucholski (approved):

backend/src/handlers/events_attendance_handler.go
```diff
@@ -59,6 +68,16 @@ func (srv *Server) handleAddAttendanceForEvent(w http.ResponseWriter, r *http.Re
 		if parsed.After(endOfToday) {
 			return writeJsonResponse(w, http.StatusUnprocessableEntity, "attempted attendance date in future")
 		}
+
+		if isDateCancelled(overrides, attendances[i].Date) {
+			return writeJsonResponse(w, http.StatusConflict, "cannot record attendance for cancelled class date")
+		}
+
+		if err := srv.Db.IsUserEnrolledInClass(uint(classId), attendances[i].UserID); err != nil {
```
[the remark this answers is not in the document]
CK-7vn: That's clever. I like it.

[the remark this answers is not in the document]
CK-7vn: I think you'll like the fix, sending it out now.

carddev81 (approved):
Good work!

PR #981, opened 2025-08-11
fix: add attendance cell for historic attendance calculations

CK-7vn:
## Description of the change

This PR Fixes attendance and enrollment calculations on the events page so that historical attendance and enrollments are reflected accurately. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210958323931668?focus=true

## Additional context

The issue we were having when rendering historical attendance data was that prior to commit `91aee47dd1531c2bdab5523c4e4a44e05d6475c7`, we were primarily using the attendance_status field to determine who was actually enrolled in a class. This becomes an issue quickly though as students can only be "enrolled" for so long. The commit referenced above added the "enrolled_at" and "enrollment_ended_at" fields which allows us to keep track of when the resident was actually involved in an active class. This PR essentially switches the logic located in on the events page to look for users that were actually enrolled in the class on that given date, using the two previously mentioned fields. To do this I created first a method (GetHistoricalEnrollmentForDate) that would calculate the enrollment status of a user in a class based on the date of the event by comparing their enrolled_at and enrollment_ended_at fields to the date of the event. Then I created an adminValidatedFeatureRoute (handleGetHistoricalEnrollmentForDate) that would call the DB method, and return the historical enrollment count for that date. I decided upon a new endpoint because it helps keep separation of concerns, instead of relying on the existing endpoint running through potentially 30 different dates on a single events page, we call the endpoint for the specific date and get the necessary count. The seperate endpoint also helps avoid N + 1 query issues that would almost definitely arise if we had tried to do it all in the existing endpoint. Also, I felt that separating this endpoint helps keep the logic cleaner, more maintanable, and easier to use from other places if needed, which I assumed is likely. After I had made sure that the actual enrollment was being calculated correctly I then switched my focus to attendance. We ran into a similar problem with attendance calculations as a result of checking against the enrollment_status like before, this would cause "phantom" status changes, where attendance rates for historical events would actually retroactively change if a residents enrollment_status changed from anything but enrolled. So, essentially entirely broken, that was fixed by updating the sql query to reflect appropriate enrolled_at time and dates.

## Warning
This can be difficult to test with seeded data, so when testing please pay special attention to actual enrolled_at dates, and the data that is being reflected in the UI.


backend/src/database/events_attendance.go
```diff
@@ -126,15 +126,24 @@ func (db *DB) GetEnrollmentsWithAttendanceForEvent(qryCtx *models.QueryContext,
...
-			and pce.id = ?`
-	if err := db.WithContext(ctx).Raw(sql, eventID, classID).Scan(&attendanceRate).Error; err != nil {
+	sql := `
+	select coalesce(
+		count(*) filter (where pcea.attendance_status = 'present') * 100.0 /
+		nullif(
+			(select count(*) from program_class_enrollments e 
+			 where e.class_id = ? 
+			   and e.enrolled_at <= ?
+			   and (e.enrollment_ended_at IS NULL OR e.enrollment_ended_at > ?)),
```
jtucholski, line 138: @CK-7vn Do we need to use `e.enrollment_ended_at >= ?` 

I'm horrible with date comparison but this would account for when the date in question is the last day of the class, right?

If I'm wrong then just close this comment.

CK-7vn: Ah, good catch, on line 138 I forgot to add the =.

CK-7vn:
@jtucholski Fixed it.

carddev81 (approved):
Looks good, tested good. Good work! I initially questioned whether the denominator could be 0 as in 14 / 0.

CK-7vn:
@carddev81 Hey....we can in fact divide by zero here at Unlocked Labs, that's just the truth of our world.

PR #988, opened 2025-09-03
fix: update cancel and completed class with until

CK-7vn:
## Description of the change

This ticket fixes the issue of classes showing up after being cancelled or completed if it was a class without an end date. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210881980228465?focus=true

### DIFF
--- a/backend/src/database/class_enrollments.go
+++ b/backend/src/database/class_enrollments.go
@@ -190,6 +190,7 @@ func (db *DB) UpdateProgramClasses(ctx context.Context, classIDs []int, classMap
 	if tx.Error != nil {
 		return newUpdateDBError(tx.Error, "begin transaction")
 	}
+	defer tx.Rollback()
 
 	var (
 		toBeCompletedEnrollments []models.ProgramClassEnrollment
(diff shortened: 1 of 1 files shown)


backend/src/database/class_events.go
```diff
@@ -867,3 +869,49 @@ func (db *DB) GetProgramClassEventOverrides(qryCtx *models.QueryContext, eventID
 	}
 	return overrides, nil
 }
+
+func (db *DB) UpdateClassEventRRuleUntilDate(classIDs []int, completionTime time.Time) error {
+	if len(classIDs) == 0 {
+		return NewDBError(errors.New("no class IDs provided"), "no class IDs provided")
+	}
+	tx := db.Begin()
+	defer tx.Rollback()
```
jtucholski, line 878: @CK-7vn How does Go handle this nested transaction?  What happens if this commits (successfully) but the outer transaction in `UpdateProgramClasses` fails?

jtucholski (commented):
Looks and works well. Makes sense to have some of the code moved around to be better shared.

I only have one comment and I'm not sure if it needs to be changed. Interested to hear what you think.

backend/src/database/class_events.go
CK-7vn: None of the transaction actually commits until we run the commit method on 913, at least update wise.

jtucholski (approved):

PR #990, opened 2025-09-06
Feat: cross-browser class event scheduling feature

JamesUnlocked:
# Class event scheduling feature that works across browsers

**Our date picker did not port well into Firefox. We now have a consistent feature across Chrome, Edge and Firefox. Instead of using a library with suspect dependencies I just  "hand rolled" one.**

[Implement a cross-browser compatible month picker for <DateInput />](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210803595523745)

## Screenshot(s)
* In Chrome 
https://github.com/user-attachments/assets/d2540945-862e-4c7b-afa3-11ce270daba1
* In Edge

https://github.com/user-attachments/assets/40b776fc-f90b-42db-a585-0c1bd2873b26

* In Firefox

https://github.com/user-attachments/assets/e150b27d-b688-4c62-93b3-ba212b2029bf

If the PR includes changes to the UI: 
- Please add screenshots or a short screengrab of the change 
- Include images at 1366 x 768 resolution for any Resident-facing changes.

## A thing to point out. I don't know why the calendar modal does not show in the screenshots but it is there !!!
**The Firefox version requires the user to select a date before triggering a event. The difference we're experiencing between Firefox and Chrome/Edge is due to how these browsers handle the initial state and default behavior of <input type="date"> elements.**

### Browser Differences:
**Chrome/Edge**:

When an <input type="date"> loads, it often has an implicit "today's date" or shows a placeholder that still triggers onChange events
* These browsers are more "eager" about firing onChange events, sometimes even on focus/blur
*They may automatically populate or suggest the current date, which can trigger the form hooks
 Firefox:

 ##  Firefox is more strict about <input type="date"> - it starts truly empty with no implicit value
* It only fires onChange when the user explicitly selects a date
* No automatic population or implicit current date behavior
* More conservative about when to trigger events
## Why This Affects our Form Integration:
Since  SimpleCalendar relies on the [onChange](vscode-file://vscode-app/c:/Users/Hudson%20Link/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) event to:

* Update react-hook-form state via [registerProps.onChange(e)](vscode-file://vscode-app/c:/Users/Hudson%20Link/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
* Extract the month and call setValue('selectedMonth', monthValue)
* Trigger the table data refresh
**Firefox requires an explicit user interaction to start this chain, while Chrome/Edge may trigger it automatically through their more permissive date input behavior.**

This is actually expected behavior - Firefox is being more standards-compliant by requiring explicit user input, while Chrome/Edge are being more user-friendly by providing implicit interactions.

The solution is working correctly - it's just highlighting the different philosophies these browsers have about form input initialization!

### This solution allows us to remove the `DateInput.tsx `component, hence it's removal from my instance of `ClassEvents.tsx`

### DIFF
--- a/frontend/src/Pages/ClassEvents.tsx
+++ b/frontend/src/Pages/ClassEvents.tsx
@@ -8,8 +8,7 @@ import {
     ServerResponseOne
 } from '@/common';
 import Pagination from '@/Components/Pagination';
-import { useForm } from 'react-hook-form';
-import { DateInput } from '@/Components/inputs/DateInput';
+import { useForm, UseFormRegister, UseFormSetValue } from 'react-hook-form';
 import { isCompletedCancelledOrArchived } from './ProgramOverviewDashboard';
 import { ClassEventInstance } from '@/types/events';
 import AttendanceCell from '@/Components/AttendanceCell';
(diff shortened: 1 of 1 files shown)


JamesUnlocked:
The updates address the following issues:

1.  The interface only displays Month name and year.
2.  The Datetime JavaScript bug is fixed ( Selecting the first day of the Month no longer makes displays the previous month)

JamesUnlocked:
The components are now browser specific. We get the old functionality for Chrome or Edge and the slightly ugly but functional Firefox version

PR #993, opened 2025-09-10
feat: add logic for downloading resident usage report per ticket 269

carddev81:
## Description of the change

Added the functionality to the Resident Profile screen that allows an authorized user to download a shareable PDF of the resident’s usage with safe defaults.

 for an Administrator to be able to download a resident's usage report based upon the criteria found within the Asana ticket.

Checklist:
- [X] Action **appears in Resident Profile Actions menu** for authorized roles (Admins) only.
- [X] Clicking the action **generates and downloads a PDF** with the sections above and *All-time* range.
- [X] **Program Participation section omitted** if Program Management is disabled for the tenant.
- [X] PDF is **legible when printed** (Letter size, simple header, table borders).
- [X] PDF includes **Facility name**.
- [x] PDF includes **UnlockEd logo** (if available). NOTE: Chris chose the new logo?
- [X] No sensitive **content history** included in MVP.

- **Related issues**: Link to related Asana ticket that this closes: [FEAT: Download a Resident Usage / Transcript from Updated Resident Usage View](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210383025477738?focus=true)

## Screenshot(s)
Screenshot from 2025-09-10 11-54-15

Buttons

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -410,7 +410,6 @@ func (db *DB) GetUserSessionEngagement(userID int, days int) ([]models.SessionEn
 			Select(`ust.user_id, TO_CHAR(DATE(ust.session_start_ts), 'YYYY-MM-DD"T"HH24:MI:SS"Z"') AS time_interval,
 			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 3600) AS total_hours,
 			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 60) AS total_minutes`).
-			Joins("JOIN users u ON ust.user_id = u.id").
 			Where("ust.user_id = ? AND ust.session_start_ts >= ?", userID, daysAgo).
 			Group("ust.user_id, time_interval").
 			Order("ust.user_id, time_interval")
@@ -420,7 +419,6 @@ func (db *DB) GetUserSessionEngagement(userID int, days int) ([]models.SessionEn
 			Select(`ust.user_id,
 			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 3600) AS total_hours,
 			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 60) AS total_minutes`).
-			Joins("JOIN users u ON ust.user_id = u.id").
 			Where("ust.user_id = ?", userID).
 			Group("ust.user_id")
 	}


backend/src/database/users.go
```diff
@@ -398,27 +398,36 @@ func (db *DB) GetLoginActivity(args *models.QueryContext, days int, facilityID *
...
+			Where("ust.user_id = ? AND ust.session_start_ts >= ?", userID, daysAgo).
+			Group("ust.user_id, time_interval").
+			Order("ust.user_id, time_interval")
+	} else {
+		//wide open query for total time spent in Unlocked
+		query = db.Table("user_session_tracking as ust").
+			Select(`ust.user_id,
+			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 3600) AS total_hours,
+			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 60) AS total_minutes`).
+			Joins("JOIN users u ON ust.user_id = u.id").
```
calisio, line 423: I don't think you need to join on users here, do you? I don't see it used anywhere in this query?

backend/src/database/users.go
```diff
@@ -398,27 +398,36 @@ func (db *DB) GetLoginActivity(args *models.QueryContext, days int, facilityID *
...
-		Where("ust.user_id = ? AND ust.session_start_ts >= ?", userID, daysAgo).
-		Group("ust.user_id, time_interval").
-		Order("ust.user_id, time_interval")
+	if days >= 0 {
+		//specific amount of days for calculating total time spent in Unlocked
+		daysAgo := time.Now().AddDate(0, 0, -days)
+		query = db.Table("user_session_tracking as ust").
+			Select(`ust.user_id, TO_CHAR(DATE(ust.session_start_ts), 'YYYY-MM-DD"T"HH24:MI:SS"Z"') AS time_interval,
+			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 3600) AS total_hours,
+			SUM(EXTRACT(EPOCH FROM ust.session_duration) / 60) AS total_minutes`).
```
calisio, line 412: just out of curiosity (i know this was in the code before), but is it more efficient to do this calculation in the backend, or does it make a difference? I thought we had a helper function in the frontend that converted seconds to hours and minutes. not a necessary fix but just a thought

calisio (approved):
two super small comments but it looks good overall!

backend/src/database/users.go
carddev81: Yes you are right you do not need that table included in the query since nothing is being accessed and I believe there is a constraint on user id so there should never be any orphaned user_ids within the user_session_tracking table.

carddev81: made changes/pushed

PR #1000, opened 2025-09-16
James unlocked/improved attendance page nav

JamesUnlocked:
## Description of the change

Users were confused by having both calendar dropdown AND pagination controls on the attendance page. The pagination only worked within the selected month, leading users to think earlier classes were missing and attendance percentages were incorrect.

## Solution
- **Removed** calendar dropdown component
- **Removed** pagination controls that only worked within months  
- **Added** clean month navigation: "← Previous | August 2025 | Next →"
- **Added** "This Month" button for quick navigation
- **Fixed** daily classes to show all events per month (31 max)
- **Added** smart button states based on class date ranges

## Benefits
- Eliminates user confusion about dual navigation
- All events for a month are visible at once
- Clear visual month navigation that users understand
- Accurate attendance tracking across all months
- Works reliably for daily, weekly, and monthly classes

- [x] Daily programs show all events per month
- [x] Navigation buttons work correctly within class date ranges  
- [x] "This Month" button functions properly

[Improve Navigation on Attendance Page](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1210958323931663)
## Screenshot(s)

https://github.com/user-attachments/assets/46c2d8b0-f480-46e0-b125-9f3ab191a2ab

### DIFF
--- a/frontend/src/Components/MonthNavigation.tsx
+++ b/frontend/src/Components/MonthNavigation.tsx
@@ -1,51 +1,36 @@
 import { ChevronLeftIcon, ChevronRightIcon } from '@heroicons/react/24/outline';
+import {
+    getPreviousMonth,
+    getNextMonth,
+    formatMonthYear,
+    getCurrentMonth
+} from '@/Components/helperFunctions/formatting';
 
 interface MonthNavigationProps {
-    currentMonth: string; // YYYY-MM format
+    currentMonth: string;
     onMonthChange: (month: string) => void;
-    hasEarlierClasses?: boolean; // Whether earlier months have classes
-    hasLaterClasses?: boolean; // Whether later months have classes
+    showPrevious?: boolean;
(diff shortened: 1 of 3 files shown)


calisio (changes requested):
1. Just testing this out briefly, I modified the schedule of this class to go back to August and through December (as seen in the schedule tab), but these events don't appear in the attendance page. Do you know if this was introduced in this PR? If it was not, then do not worry about fixing it in this PR, it can be logged as a different ticket. Let me know! 

https://github.com/user-attachments/assets/f6ae5656-b639-4e55-b207-01ef5b04d1dd


2. I also noticed that the "this month" button disappears/reappears which shifts where the "next" button is/the width of that section. Would it be possible to keep the button there, but just when we are on this month disable it?

calisio (approved):
looks good!

PR #1005, opened 2025-09-19
Issue when transferring resident facility

corypride:
## Description of the change

updated the transferResident function and facility.id value to be facility_id to rid application related console error.

- **Related issues**: [Asana task](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1211224372188163)

## Screenshot(s)

https://github.com/user-attachments/assets/b21016ef-715f-466d-8a08-c436a7ba69ae

### DIFF
--- a/frontend/src/Pages/ResidentProfile.tsx
+++ b/frontend/src/Pages/ResidentProfile.tsx
@@ -176,7 +176,7 @@ const ResidentProfile = () => {
         const response = await API.patch(`users/resident-transfer`, {
             trans_facility_id: resident?.trans_facility_id,
             user_id: resident?.user.id,
-            curr_facility_id: resident?.user.facility.id
+            curr_facility_id: resident?.user.facility_id
         });
         checkResponse(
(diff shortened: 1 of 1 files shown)


corypride:
@jtucholski As it pertains to this issue, I believe that the modifying the variable facility.id -> facility_id solves for the original issue on the [asana ticket](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1211224372188163?focus=true).
image

See image and Note a total 6 residents, 1 added and 5 imported. (
image
)
 Of those, the user Test User (created by adding resident) was inserted into the kratos.identities table and the 5 users that i bulk created were not. With that being said, if the user that you were trying to transfer was created using the bulk create, you will get an error. 
image 
image

 Looking at the way the the HandleBulkCreate function is written and does not call HandleCreateUserKratos(), I believe that it was designed like this?

carddev81 (approved):
Nice! This looks good and works good.

PR #1006, opened 2025-09-19
feat: adds posthog

calisio:
## Description of the change

Adds posthog analytics. We'll need to add the env variables to make it work. You'll also have to be invited to the posthog project in order to see the dashboards. Reach out to me and I can send them over!

- **Related issues**: [Asana Posthog integration ticket](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1211328741842500?focus=true)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,9 @@ services:
       - node_modules:/app/node_modules
     networks:
       - intranet
+    environment:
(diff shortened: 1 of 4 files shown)


calisio:
@jtucholski Yes, I thought we had decided to not identify users in this ticket. I treated the ticket as simply just connect to posthog (which is why I didnt even touch into identifying users/resetting). I can handle that in this ticket if you'd like?

PR #1008, opened 2025-09-21
fix(proxy): preserve encoded paths and correct forwarded proto

cole-ul:
## Description of the change

Updates the Kiwix proxy handler to correctly preserve double‑encoded path segments and reflect the actual request scheme while keeping the diff minimal.

Key changes:
- Use `r.URL.RawPath` (fallback to `EscapedPath`) so multi‑encoded sequences (e.g. `%252C`) are not normalized away.
- Replace `url.JoinPath` with manual concatenation to avoid decoding/normalization.
- Set both `req.URL.Path` (decoded) and `req.URL.RawPath` (original encoded) when forwarding.
- Use actual `scheme` (`http` in dev, `https` otherwise) for `X-Forwarded-Proto`.
- Simplify redirect rewriting (string concat instead of a second `url.JoinPath`).
- Net delta: +2 lines (23 insertions / 21 deletions).

Behavior / impact:
- Fixes missed assets when paths contain double-encoded characters.
- More accurate protocol reporting for services behind the proxy in dev.
- No change to existing cache headers beyond formatting.

**Related issues**: Closes ASANA-ID-463

## Additional context

Focus areas for review:
- Correctness of `finalEncodedPath` construction for libraries with or without trailing slashes.
- Redirect rewriting still preserves intended proxy prefix (query + fragment still intentionally omitted).
- Ensures no regression when `RawPath` is empty.

Known omissions (intentional):
- Redirect `Location` query parameters and fragments are still dropped (matches prior behavior).
- No new integration test for encoded path cases (follow-up candidate).
- No env flag to force HTTPS in dev (kept scope minimal).

Decisions deferred:
- Adding query/fragment preservation (will do separately).
- Larger refactor to avoid reparsing full URL (kept structure close to original).
- Adding tests in this PR (kept patch lean).
- Restoring verbose cache warning comment (removed as noise).

Edge cases considered:
- Library path of `/` (still produces valid absolute path).
- Asset paths already beginning with `/` (guard prevents double slashes).
- Missing `RawPath` (fallback maintains previous baseline behavior).
- Dev behind TLS terminator now sees `X-Forwarded-Proto: http` (accurate for plain dev; optional override could be added).

Risk:
- Low. Localized to one handler; easy rollback (revert commit `608584b5`).

Follow-up recommendations:
- Preserve query + fragment in redirects.
- Add integration test covering `%252C` or similar.
- Add `FORCE_HTTPS_PROXY` env override.
- Trace-level logging of `finalEncodedPath` in dev for debugging.

Removed / deprecated features:
- None.

### DIFF
--- a/backend/src/handlers/proxy_handler.go
+++ b/backend/src/handlers/proxy_handler.go
@@ -26,15 +26,24 @@ func (srv *Server) handleForwardKiwixProxy(w http.ResponseWriter, r *http.Reques
 		srv.errorResponse(w, http.StatusNotFound, "Library context not found")
 		return
 	}
-	assetPath := strings.TrimPrefix(r.URL.Path, fmt.Sprintf("/api/proxy/libraries/%d", library.ID))
+	// Use RawPath if present to preserve double-encoded segments
+	rawPath := r.URL.RawPath
+	if rawPath == "" {
+		rawPath = r.URL.EscapedPath()
+	}
+	assetPath := strings.TrimPrefix(rawPath, fmt.Sprintf("/api/proxy/libraries/%d", library.ID))
 	if assetPath != "" && assetPath != "/" {
 		assetPath = strings.TrimPrefix(assetPath, library.Path)
 	}
-	targetURL, err := url.JoinPath(library.BaseUrl, library.Path, assetPath)
-	if err != nil {
-		srv.errorResponse(w, http.StatusBadRequest, "Malformed request to build targetURL")
-		return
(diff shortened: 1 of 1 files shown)


calisio (commented):
For some reason I am now unable to enter into any of the libraries without getting an error. I confirmed this is the case only on this branch, as it works as expected on my `main` branch. Is there something special I should be doing? The error I get is
`server-1            | 2025/09/22 17:36:19 http: proxy error: EOF`

Any idea of the issue?

PR #1014, opened 2025-09-25
feat: add first initial config that allows devs to debug the backend api within vscode, will work in other IDEs

carddev81:
## Description of the change

This PR introduces minimal configuration changes to enable running a debugger against the backend API.

- Added the go-delve library to the dev backend Docker image, which starts a debugger instance.
- The debugger can be remotely attached from VS Code or any other IDE (you may need to configure your own connection settings).

# Steps to Use
1. Run `make init` or `make dev`
2. In VS Code, open **Debug View** | Click **Go Remote Debug**
image

3. Set a breakpoint in your code and trigger the corresponding flow. Execution will pause at the breakpoint, and you can begin debugging as normal.
image

# Notes
- This is intended to support local development and debugging.
- If this is approved I will then move onto adding a debugger to the provider-middleware

- **Related issues**: No related issues; added purely to streamline the developer workflow.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,7 +11,13 @@
       "host": "127.0.0.1",
       "showLog": true,
       "apiVersion": 2,
-      "trace": "verbose"
+      "trace": "verbose",
+      "substitutePath": [
+        {
+          "from": "${workspaceFolder}",
+          "to": "/app"
+        }
+      ]
     }
   ]
 }


corypride (changes requested):
The configuration looks right overall - just seeing something with the connection. on my end I'm curious if this might be environment-specific.

I'm wondering if it could be:
- VS Code Go extension version compatibility
- Docker networking differences
- Go/Delve version timing issues

Could you  have another team member test on different os's and maybe add some debugging logs to see what's happening when and if the connection fails?

The setup seems correct, so it might just need some environment-specific adjustments.
## Vscode
vscode
## Cursor
cursor

corypride (approved):
Works Great!
image

PR #1015, opened 2025-09-26
feat: editable enrollment date component

CK-7vn:
## Description of the change
This pull request adds the ability for admins to update a residents enrolled date retroactively so that the date reflects the actual date appropriately. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209457756469406/task/1211317866761100?focus=true

## Screenshot(s)
image
image
image
image

### DIFF
--- a/backend/src/handlers/class_enrollments.go
+++ b/backend/src/handlers/class_enrollments.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"time"
 )
 
 func (srv *Server) registerProgramClassEnrollmentsRoutes() []routeDef {
@@ -16,6 +17,7 @@ func (srv *Server) registerProgramClassEnrollmentsRoutes() []routeDef {
 		adminFeatureRoute("GET /api/program-classes/{class_id}/enrollments", srv.handleGetEnrollmentsForProgram, axx),
 		adminValidatedFeatureRoute("POST /api/program-classes/{class_id}/enrollments", srv.handleEnrollUsersInClass, axx, resolve),
 		adminValidatedFeatureRoute("PATCH /api/program-classes/{class_id}/enrollments", srv.handleUpdateProgramClassEnrollments, axx, resolve),
+		adminValidatedFeatureRoute("PATCH /api/program-classes/{class_id}/enrollments/{enrollment_id}/date", srv.handleUpdateEnrollmentDate, axx, resolve),
 		adminValidatedFeatureRoute("DELETE /api/programs/{id}/classes/{class_id}/enrollments", srv.handleDeleteProgramClassEnrollments, axx, resolve),
 		adminValidatedFeatureRoute("GET /api/programs/{id}/classes/{class_id}/enrollments/{enrollment_id}/attendance", srv.handleGetProgramClassEnrollmentsAttendance, axx, resolve),
 		validatedFeatureRoute("GET /api/users/{id}/program-completions", srv.handleGetUserProgramCompletions, axx, UserRoleResolver("id")),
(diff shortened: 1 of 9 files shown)


backend/src/handlers/class_enrollments.go
```diff
@@ -144,6 +146,44 @@ func (srv *Server) handleUpdateProgramClassEnrollments(w http.ResponseWriter, r
...
+		return newInvalidIdServiceError(err, "enrollment ID")
+	}
+	enrollmentUpdate := struct {
+		EnrolledDate string `json:"enrolled_date"`
+	}{}
+	if err := json.NewDecoder(r.Body).Decode(&enrollmentUpdate); err != nil {
+		return newJSONReqBodyServiceError(err)
+	}
+	if enrollmentUpdate.EnrolledDate == "" {
+		return newInvalidIdServiceError(errors.New("enrolled date is required"), "enrolled date")
```
carddev81, line 165: i think this newInvalidIdServiceError should be changed to newBadRequestServiceError?

frontend/src/Components/modals/FormModal.tsx
```diff
@@ -192,6 +192,8 @@ export const FormModal = forwardRef(function FormModal<T extends FieldValues>(
                                             getValues={input.getValues}
                                             disabled={input.disabled}
                                             allowPastDate={input.allowPastDate}
+                                            minDate={input.minDate}
```
carddev81, line 195: Are these inputs being used? I can't see where they are being accessed or being validated against? Maybe they just got left behind during an iteration of coding?

frontend/src/Components/modals/index.ts
```diff
@@ -61,6 +61,8 @@ export interface Input {
     disabled?: boolean;
     allowPastDate?: boolean;
     monthOnly?: boolean;
+    minDate?: string;
```
carddev81, line 64: same here as before

frontend/src/Components/inputs/DateInput.tsx
```diff
@@ -20,6 +20,8 @@ interface DateProps {
         | Record<string, Validate<any, any>>; // eslint-disable-line @typescript-eslint/no-explicit-any
     disabled?: boolean;
     monthOnly?: boolean;
+    minDate?: string;
```
carddev81, line 23: i left a comment about these in the other comments, just making sure if you remove them these aren't forgotten.

frontend/src/Components/EditableEnrollmentDate.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const defaultValues = useMemo(
+        () => ({
+            enrolled_date: enrollment.enrolled_at
+                ? toDateOnly(enrollment.enrolled_at).toISOString().split('T')[0]
+                : toDateOnly(enrollment.created_at).toISOString().split('T')[0]
+        }),
+        [enrollment.enrolled_at, enrollment.created_at]
+    );
+
+    const dateConstraints = useMemo(() => {
```
carddev81, line 96: not sure this function is needed any either if you are not using it? I could be completely wrong, but this goes along with the other comment about removing the input fields because they might not be used.

frontend/src/Components/modals/FormModal.tsx
CK-7vn: Oh crap, no, I tried graying out dates that aren't eligible to be used, and forgot to take those out, because I couldn't get it how I liked it.

CK-7vn:
@jtucholski @carddev81 Per the ticket: The enrollment start date must allow values within the bounds of the class start and end dates. 

Which is what I believe it was doing? No? We don't want to allow enrollment in the future, because you can't enroll a student in the future. I think the point of this is so that if a user has been involved in a class that is currently going on, or a student was missed in an original enrollment, the admin is able to retroactively enroll them on the appropriate date.

CK-7vn:
> 1. The strange scenario when someone is trying to update the enrollment date, but the class hasn't started yet (so they cant change enrollment date to the future, but also cant update it to a past date because it is before the start date of the class). I dont think we should give them the option to edit the enrollment date at this point (aka block the button or something)? I think this would mean that we only allow them to edit once it is past the start date. Thoughts? Open to push back
> 
>     2. would we be able to fix the small glitchy behavior where the enrolled button keeps switching back and forth in the background? see vid
> 
> 
> Screen.Recording.2025-10-17.at.1.56.45.PM.mov
> 
> in the normal scenario, works as expected. just a few things I caught to make it not seem as buggy!

Ya, should be able to pretty easily just hide the button until the class has started, that'd probably be the best bet.

calisio (approved):
looks great, thanks!!

PR #1079, opened 2026-02-10
feat: select all on facility comparison

CK-7vn:

## Description of the change

Adds a select all to the facility comparison report export. 

- **Related issues**: https://app.asana.com/1/1201607307149189/task/1213058997016756

## Screenshot(s)

image


corypride (changes requested):
@CK-7vn  this works as intended, but i think that when it was mentioned in the [video](https://drive.google.com/file/d/12Q_C0Fc_TKpXLVGKuzdKKyUM9R4Q0XoA/view) at around minute marker 8:31, while in an effort to click check Program types and then Funding types, that "Maybe there could be a 'Select All'. This leads me to believe that this feature was intended to be used to "Select All" across each of the groups and not just facility. With that being said, I would like to ask that we create a true "Select All" here or at least a "Select All Program Types, Funding Types as well. Does that make any sense?

Image

PR #1107, opened 2026-03-11
Ck 7vn/total refactor residents 3

CK-7vn:

## Description of the change
Adds bulk selection and bulk action dialogs to the Residents table. Residents can be selected across pages using checkboxes, with a floating action bar that appears at the bottom of the screen. Three new bulk operations are supported: reset passwords, deactivate,
  and delete.

  Backend:
  - 3 new POST endpoints under /api/users/bulk/ — reset-password, deactivate, delete
  - All endpoints batch-fetch users in a single query, filter by facility for non-system/dept admins, and loop sequentially with skip-on-error per user
  - Reset password returns [{user_id, username, name, doc_id, temp_password}] for client-side CSV generation
  - Deactivate and delete return {success_count, failed_count}

  Frontend:
  - Header checkbox selects/deselects all non-deactivated residents on the current page
  - Individual row checkboxes toggle without triggering row navigation
  - Deactivated rows have no checkbox
  - Selection persists across page changes, clears on search/filter/sort changes
  - Floating action bar (centered, matches prototype) with Clear Selection, Reset Passwords, Deactivate, and Delete buttons
  - BulkResetPasswordDialog: 2-step flow (confirm with count -> results with CSV download)
  - BulkDeactivateDialog: orange warning with resident list, count confirmation
  - BulkDeleteDialog: red warning with resident list, count confirmation
  - After any bulk action, selection clears and table refreshes via SWR mutate

## This PR
image

## prototype
image

## This PR
image

## Prototype 
image

## This PR
image

## Prototype 
image

## This PR
image

## Prototype 
image

## This PR
image

## Prototype 
image

## Additional context
Chose to center the bulk actions bar and not drop "selected" to the row below, it looks cleaner in my personal opinion and I was unable to get it to center/match the prototype exactly without adjusting the navbar.

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -109,17 +109,13 @@ func (db *DB) GetUserByDocIDAndID(ctx context.Context, docID string, userID int)
 	return &user, nil
 }
 
-func (db *DB) GetUsersByIDs(userIDs []uint, args *models.QueryContext) ([]models.User, error) {
+func (db *DB) GetUsersByIDs(ctx context.Context, userIDs []uint, facID uint) ([]models.User, error) {
 	users := make([]models.User, 0, len(userIDs))
-	if err := db.WithContext(args.Ctx).Where("id IN (?)", userIDs).Find(&users).Error; err != nil {
-		return nil, newGetRecordsDBError(err, "users")
+	tx := db.WithContext(ctx).Where("id IN ? AND role = ?", userIDs, models.Student)
+	if facID != 0 {
+		tx = tx.Where("facility_id = ?", facID)
 	}
-	return users, nil
-}
-
-func (db *DB) FindUsersByIDs(userIDs []uint) ([]models.User, error) {
-	var users []models.User
-	if err := db.Where("id IN ?", userIDs).Find(&users).Error; err != nil {
+	if err := tx.Find(&users).Error; err != nil {
 		return nil, newGetRecordsDBError(err, "users")
 	}
 	return users, nil
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -111,7 +111,15 @@ func (db *DB) GetUserByDocIDAndID(ctx context.Context, docID string, userID int)
 
 func (db *DB) GetUsersByIDs(userIDs []uint, args *models.QueryContext) ([]models.User, error) {
 	users := make([]models.User, 0, len(userIDs))
-	if err := db.WithContext(args.Ctx).Find(&users).Where("id IN (?)", userIDs).Error; err != nil {
+	if err := db.WithContext(args.Ctx).Where("id IN (?)", userIDs).Find(&users).Error; err != nil {
+		return nil, newGetRecordsDBError(err, "users")
+	}
+	return users, nil
+}
+
+func (db *DB) FindUsersByIDs(userIDs []uint) ([]models.User, error) {
+	var users []models.User
+	if err := db.Where("id IN ?", userIDs).Find(&users).Error; err != nil {
 		return nil, newGetRecordsDBError(err, "users")
 	}
 	return users, nil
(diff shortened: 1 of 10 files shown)


carddev81 (changes requested):
@CK-7vn here are my findings
 
1. Resident listing page spans the entire width of the viewable area and should be centered and padded with grey on all sides.

### Target Figma
Image

### Current
Image
 
2. Facility sorting not sorting alphabetically 
3. When creating a user and the user name already exists the error message is 'userexists', should this be 'Username already exists'? Figma does not show this so went back to the original and copied that message.
4. Resetting passwords	
    - When clicking 'Download Password File' the modal does not auto close after download completes
    - No toast message pops up after downloading completes

### REVIEWER NOTES ONLY @CK-7vn @calisio @corypride 
Noticed that the deactivated accounts do not have checkboxes on the current redesign UI. To me this makes sense (do we want to allow the deletion) but want @calisio to be aware that this does not match figma design where every row has a checkbox regardless.
 
### Target Figma
Image

### Current
Image

backend/src/handlers/user_handler.go
```diff
@@ -751,3 +754,173 @@ func (srv *Server) handleExportResidentAttendanceCSV(w http.ResponseWriter, r *h
...
+		UserIDs []uint `json:"user_ids"`
+	}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return newJSONReqBodyServiceError(err)
+	}
+	if len(req.UserIDs) == 0 {
+		return newBadRequestServiceError(errors.New("no user IDs provided"), "user_ids is required")
+	}
+	var users []models.User
+	if err := srv.Db.Where("id IN ?", req.UserIDs).Find(&users).Error; err != nil {
```
carddev81: @CK-7vn Direct database calls should be within the database layer when possible. Please fix all where possible

backend/src/database/users.go
```diff
@@ -111,7 +111,15 @@ func (db *DB) GetUserByDocIDAndID(ctx context.Context, docID string, userID int)
...
 func (db *DB) GetUsersByIDs(userIDs []uint, args *models.QueryContext) ([]models.User, error) {
 	users := make([]models.User, 0, len(userIDs))
-	if err := db.WithContext(args.Ctx).Find(&users).Where("id IN (?)", userIDs).Error; err != nil {
+	if err := db.WithContext(args.Ctx).Where("id IN (?)", userIDs).Find(&users).Error; err != nil {
+		return nil, newGetRecordsDBError(err, "users")
+	}
+	return users, nil
+}
+
+func (db *DB) FindUsersByIDs(userIDs []uint) ([]models.User, error) {
```
corypride: has no facility context here.

backend/src/handlers/user_handler.go
```diff
@@ -751,3 +753,173 @@ func (srv *Server) handleExportResidentAttendanceCSV(w http.ResponseWriter, r *h
...
+		}
+		successCount++
+	}
+	return writeJsonResponse(w, http.StatusOK, map[string]int{
+		"success_count": successCount,
+		"failed_count":  failedCount,
+	})
+}
+
+func (srv *Server) handleBulkDeleteUsers(w http.ResponseWriter, r *http.Request, log sLog) error {
```
corypride, line 894: Bulk endpoint should probably validate that the target users are actually Students maybe?

corypride (changes requested):
UI matches are far as looking like the Figma. Great job on that!

-  I did however, notice and issue with all bulk actions on the Resident page, If residents were spread across multiple pages. 

https://github.com/user-attachments/assets/7f7e3415-fde2-4b39-8db5-0000dd8fd578

Image
Image


- user_handler.go, Missing Role Validation, to make sure that target users are actually  Students

-  And we  need error reporting for bulk actions, currently we are returning only an array of successful resets only. Would like to probably see invalid_rows as well in a similar way to that of Bulk Upload.

corypride (approved):
Tested good.

PR #1112, opened 2026-04-06
Ck 7vn/knowledge center

CK-7vn:

## Description of the change
Brings over the knowledge center from the new design, removes knowledge insights, redirects users to dashboard and home on login, and migrates the guided tour feature from the original frontend/ to frontend-v2

## Screenshot(s)
image
image
image
image
image
image
image

## Additional context
  - Tour steps 6-7 (#knowledge-center-search-lib, #knowledge-center-fav-lib) reference per-card search/favorite buttons from the old LibraryCard design. The new ResidentKnowledgeCenter uses a unified ContentCard without those buttons, so these steps will be skipped by
   Joyride's TARGET_NOT_FOUND handling. These steps should be updated or removed in a follow-up once the tour content is reviewed against the new KC design.
  - #library-viewer-favorite -- the redesigned LibraryViewer on this branch removed the visible bookmark toggle button from the header. The tour useEffect and ID are in place but the button element doesn't exist yet. This step will also be skipped via
  TARGET_NOT_FOUND.
  - The knowledge-insights page has been removed in a separate branch. The admin post-login redirect was updated here to point to /dashboard instead.
  - The getAdminLink function was simplified since knowledge-insights no longer exists. If feature-based admin routing is still desired, it should be revisited.


provider-middleware/video_dl.go
```diff
@@ -349,7 +349,12 @@ func (vs *VideoService) retrySingleVideo(ctx context.Context, videoId int) error
 
 func (yt *VideoService) addVideos(ctx context.Context) error {
 	params := yt.Body
-	urls := params["video_urls"].([]any)
+	rawUrls, ok := params["video_urls"].([]any)
+	if !ok || len(rawUrls) == 0 {
+		logger().Warn("addVideos called with no video_urls")
+		return nil
+	}
+	urls := rawUrls
```
carddev81: is urls needed here? I suggest to change `rawUrls` on line 352 and 353 `len(rawUrls)` to use the name `urls` then you won't have to create another variable.

provider-middleware/provider-middleware
carddev81, line 1: what is this binary for? it is new? Can we remove it?

carddev81 (changes requested):
@CK-7vn here are my findings:

I was unable to test Helpful Links due to `Method Not Allowed` see below:

Image

### PR Review Findings
1. Dept. Admin clicking star--change toast message to match 'Library featured' and 'Library unfeatured' (since figma doesn't show anything I would stick with the messages from the old version)
2. Also for the toggle of visibility toast messages--match old version since figma doesn't state it 'Library is now hidden' or 'Library is now visible'
3. Pressing back button on step 13 of tour gives 404; you may want to check all others too, I just tested this one step
4. Good work on the Home page, the only thing I see that may need some work is the Featured content images--they are extremely pixelated? maybe come up with a slightly different design? @calisio's design eye might help. 

Image
 
5. Cards are sized different when text wraps--Card should stay the same size and when text reaches a certain limit (let it span 2 lines for consistency) then truncated to include the '...' at the end

Image

Also please see other comments, I found a new binary was added?

provider-middleware/provider-middleware
CK-7vn: No, must have just accidentally committed it, i'll remove it.

carddev81: Great! Thanks! I thought that this is what you may have created for the youtube downloader to work. Let me know once you get these changes completed.

carddev81 (changes requested):
Found the following

1. _rollover issue from last review, just copied issue here_ Cards are sized different when text wraps--Card should stay the same size and when text reaches a certain limit (let it span 2 lines for consistency) then truncated to include the '...' at the end

Image

2. Category tags are missing in the associated cards

Image

carddev81:
> Found the following
> 
> 1. _rollover issue from last review, just copied issue here_ Cards are sized different when text wraps--Card should stay the same size and when text reaches a certain limit (let it span 2 lines for consistency) then truncated to include the '...' at the end
> 
> Image
> 2. Category tags are missing in the associated cards
> 
> Image
> 3. You may or may not have to add the categories from the redesign (@calisio please let @CK-7vn know if these should be added or not as soon as possible), to see these just compare the ones between your PR and the redesign.

REVIEWERS NOTE: I manually ran SQL statements to bind the records to the specific categories, also noted that each parent record can be associated to mutiple categories and the cards should display a plus symbol, ect...just look at the redesign

CK-7vn:
Card texts are fixed. 


Super weird though, I explicitly remember making sure that was taken care of on my original push/pr but I must have messed it up after I checked it originally.

carddev81 (approved):
Made the following changes to Knowledge Center Management/Resident
1. Filter 
    - Facility Admin - When selecting Featured Only no cards would display when there are libraries featured
    - Department Admin - When selecting Featured Only all cards would still display for videos
2. Cards were not sized correctly
    - title of video cards was not being clipped which caused the card to grow larger than expected
3. Resident view UNABLE to test due to no scrollbar on page
    - added scroll bar
4. Featured content
    - missing feature flag on top of card

PR #1115, opened 2026-04-23
feat: add schedule and refactor for code reuse and efficiency

carddev81:

## Description of the change

This PR delivers the class schedule feature along with a refactor of scheduling-related components for improved code reuse and maintainability.

# Schedule feature (frontend)                                                                          
- Rewrote Schedule.tsx with full support for viewing, managing, and filtering class events         
- Added new modals: ChangeInstructorModal, ChangeRoomModal, and RescheduleSessionModal               
- Refactored CancelEventModal, RescheduleEventModal, RescheduleSeriesModal, and RestoreEventModal to use a shared useChangeEventField hook, eliminating duplicated submit logic across all event-change flows
- Removed CancelSessionModal from class-detail (consolidated into shared schedule components)
- Updated SessionDetailSheet, SessionsTab, ScheduleTab, and ClassHeader to align with the new modal architecture                                                                                         
- Added RRuleControl enhancements for recurrence rule editing                                        
- Extended lib/formatters.ts with new date/time formatting utilities                                 
                                                                                                       
Backend                                                                                              
- Added DB migration (00067) with new columns on program_class_events (e.g. instructor override, room override, change reason)
- Extended class_events.go, program_classes.go, and programs.go with queries supporting per-event overrides and series changes
- Added facilities instructor endpoint (facilities_handler.go)
- Updated models (class_event.go, program_classes.go) to reflect new fields
- Minor integration test cleanup 

- **Related issues**: Link to related Asana ticket that this closes: N/A

## Screenshot(s)

image

image

NOTE TO REVIEWERS: PLEASE TEST AS MANY EDGE CASES AS POSSIBLE. Also there are some differences within the modals, i believe the only difference is that i took out the apply to all future sessions when rescheduling a single session due to it making sense too.


backend/migrations/00067_add_columns_to_program_class_events.sql
```diff
@@ -0,0 +1,45 @@
...
+    ADD COLUMN is_cancelled BOOLEAN NOT NULL DEFAULT FALSE;
+
+UPDATE public.program_class_events pce
+SET instructor_id = pc.instructor_id
+FROM public.program_classes pc
+WHERE pce.class_id = pc.id
+  AND pc.instructor_id IS NOT NULL;
+
+DROP INDEX IF EXISTS idx_program_classes_instructor_id;
+ALTER TABLE public.program_classes DROP COLUMN IF EXISTS instructor_id;
```
CK-7vn, line 15: I think this is causing problems. After migration I'm unable to create classes, and do some other things, it looks like from my fuzzy search that there's still live queries in: 
1. database/program_classes.go:544
2. /database/program_classes.go:557
3. database/program_classes.go:699
4. database/program_classes.go:704
Just need to redirect those to program_class_events.instructor_id since that's the new source of truth.

CK-7vn (commented):
Not sure how we can shorten this cancellation line through the Non-DST Class but I would assume we don't want it extending across the section

Image

-------------------------------------------------
Off by one on DST in sidebar note time :( 

Image

** After DST switch ** 

Image

------------------------------------
Should we be able to change the instructor for a single session? Noticed the Reschedule single session modal doesn't have instructor, but if its this and recurring, it does have it. 

Image 
Wasn't sure. 

------------ 
Notice the black ring? I'm not sure if that's something we can change or if its on purpose, but I figured since it's not that consistent gray I'd note it. 

Image

--------------
Focus ring on tab is black throughout the calendar, I'm assuming this would be an editable class in react-big-calendar but unsure; 

Image
------- 

I'm not sure if this is just a wording thing or if i'm thinking into it too much; But when we cancel a session/class, we have the option to "apply this change to all future sessions"; The button we have to restore sessions says "Restore all future sessions" Should we include a "Restore this and all future sessions" or something just so the user has clarification that they are also restoring the session thney are in/clicked on? 

------

I noticed the highlight in the schedule length/split is not the same as it is in the figma; In the figma, after your cursor clicks on it, it is gray until you unclick, and then it turns green (green is the correct selected color); In this code; it goes to a different shade of green during click and then once cursor is released/not clicked, it is back to the matching green. 

Image

figma: 

Image

--------------------------------------------------------
There is a highlight on the pagination? component in the figma when it is clicked
Figma: 

Image

This: 

Image

---------------------------------------------------------
I can't remember if we decided if the header was going to be sticky across the application or not, so I'm adding this here, but, It could be correct. In the figma the header is not sticky on the schedule page; On this it is. 
This: 

Image

Figma; 

Image

--------------------------------------------------------------
The X button is slightly off in the Reschedule class modal it is moved to the upper right corner farther than in the figma;
This: 

Image
Figma: 

Image
-------------------------------------------------
Even after clicking "Apply to this session" in the reschedule modal of the figma, it still includes an "Apply this change to all future sessions" checkbox, this checkbox is not in this code after "Apply to this session" 
Figma: 

Image

This: 

Image
----------------------------------------------
Same issue with the "x" button as in the reschedule modal above; it is farther to the upper right in the figma; This is also the case in the Change Instructor modal 
This: 
Image
Figma; 
Image

Change Instructor Modal: 
This: 

Image

Figma: 

Image

--------------------------------------------------------

I also noticed, that there is no room-conflict detection; I'm not sure if that is out of the scope for this PR; if it is then just disregard this comment, but when changing the room of a class/session in direct conflict with another session on the Schedule it did not say anything, or reject the request; 

Note the room and times in these pictures; 

Image

Image


This gets even more weird....When changing the room of the class that starts first to the same room as the class that starts second, it shows the room conflict modal; but, if I go to the second class that starts after the start of the first class to the same room as the first class; it accepts the request without issue. Recorded a short clip...there's just a ton of weird stuff going on I guess because that same scenario just behaved in another completely different way; this has to be some type of RRULE issue or something...may not even be worth fixing in this PR and creating a whole other ticket for it...
https://www.loom.com/share/6ef54f2e597b4146a2184c6c61e79bc0

Image


----------------------------------------------------
When changing the instructor of class x to the instructor of class y at the same time it returned a toast of "room already booked at this time" which seems weird

Image

---------------------------------
I went to go try and cancel a class and it intermittently would return "room already booked at this time" even when I wasn't doing anything with the room, I was just trying to cancel it. After cancelling a different class + session entirely, I went back to the prior class and it allowed me to cancel. 

----------------------------------------
The same issue with the "x" button in the Reschedule series modal; 

--------------------------------------------------------------
The spacing on the bottom buttons of the reschedule modal are spaced differently than the figma.
This: 

Image
Figma: 

Image
---------------------------------------------------
In the agenda, i can se that the bottom of the agenda isn't viewable;

Image
--------------------------------------------------------
Should we have an option not just in the agenda, but on the schedule as well to undo a specific cancellation instead of JUST restoring ALL of the future cancelled sessions? 

Image

------------------------------------------------------
The Take Attendance icon is different 
Figma: 
image

This:
image

-------------------------------------------------------
Couple questions/wonders
1. Should there be a way on the schedule to indicate to users whether or not attendance has been taken? 
2. Looking at the prototype code, if we COULD reschedule a class in the figma/prototype, then I believe the cell would/should be blue in the schedule for the rescheduled class, are we matching this in our schedule? 
3. Should we include "View Full Class Details" on a cancelled class? Or make the user go find the class in the classes list
4. Should we find a way to control overlap/overflow? My worry is we'll run into the same issue that we had in v2 where when there's a lot of events they begin overlapping eachother and just looking bad, I know i did some research a couple months ago and we can do a lot of customization in terms of step, min, max, etc in react-big-calendar; just an idea and probably something to talk about. Maybe the calendar should be dynamic, maybe it's totally cool how it is, i'm probably not the one to answer that question with my eye. 
5. 

Definitely a big PR, with a lot of moving parts, but, that being said, great job man. Tried to be thorough like you asked.

carddev81:
@CK-7vn lol, i looked at this one below, but that line is actually the current time line, if that makes sense? I'll screenshot it here, but that line represents the current time, at least that's the line I think you mean?

image

image

carddev81:
@CK-7vn Very good point here, I will put a ticket in for this, want to make sure with Carolina before making this change and do not want to hold this one up because we are unsure if we want to follow along with the figma or not on this one. (Schedule Tab/Schedule)
image

CK-7vn:
> @CK-7vn lol, i looked at this one below, but that line is actually the current time line, if that makes sense? I'll screenshot it here, but that line represents the current time, at least that's the line I think you mean?
> 
> image image

Lol, I think you're right. Pretty sure it is the current date/time. Well, that saves some time then!

carddev81:
@CK-7vn these will stay black, issue with getting it to change color due to browser handling internal icons differently. Did research and now I understand why @calisio kept as is. So at least this is exact like figma.
image

corypride (changes requested):
# ✅ PR #1115 Manual Testing Checklist

## 🧪 Test Data Setup
- [X] Run seed scripts  
```

docker compose exec -T postgres psql -U unlocked -d unlocked -f database_seed.sql
docker compose exec -T postgres psql -U unlocked -d unlocked -f seed_script_extended.sql

```

---

## 🐛 Bugs Caused by PR #1115

### Seeder Compilation Failure — `make seed` Broken
- [X] Attempt `make seed` on DB with migration 00067 applied — ❌ compile error
- **Root cause**: `backend/seeder/main.go:646` sets `InstructorName: faker.Name()` on a `models.ProgramClass` struct. Migration 00067 removed the `instructor_name` column from `program_classes` and dropped the `InstructorName` field from the model. The seeder was not updated.
- **Impact**: `make seed` / `go build ./backend/seeder/...` fails to compile. Cannot use seeder for test environment setup after applying this migration.
- **File**: `backend/seeder/main.go:646`

---

## 🐛 Discovered Bugs — Unrelated to PR #1115

### Bulk Upload: Students Cannot Log In After Import
- [X] Navigate to Residents page → click "Bulk Import"
- [X] Upload `bulk_upload_test.csv` (3 rows: atestbulk/B1001, btestbulk/B1002, ctestbulk/B1003)
- [X] Validate CSV — 3 valid, 0 invalid ✓
- [X] Click "Create 3 Accounts" — success toast shown, 3 users appear in residents table ✓
- [X] Check DB: `SELECT kratos_id FROM users WHERE username IN ('atestbulk','btestbulk','ctestbulk')` — **all `kratos_id` values are empty** ❌
- [X] Attempt login as `atestbulk` / `ChangeMe!` — **"Incorrect username or password."** ❌ Login fails

**Root cause**: `handleBulkCreate` (`backend/src/handlers/user_handler.go:689`) calls `CreateUsersBulk` to insert users into the DB but never calls `HandleCreateUserKratos`. Single user creation (`handleCreateUser:256`) does call it. Bulk-created students have no Kratos identity and cannot log in with any password.

**Screenshots**: bulk_upload_validated.png, bulk_upload_created.png, bulk_upload_login_fail.png
**Seed data**: `bulk_upload_test.csv`

---

## 🗓️ Testing completed on:
- [X] Date: 2026-04-29

---

## 📊 Overall Results

**15 of 15 tests completed**

| Test | Status | Notes | Screenshot |
|------|--------|-------|------------|
| **Test 8.8** Cross-facility instructor assignment | ❌ FAILED — Critical Security | `handlePatchEventOverride` assigns `instructor_id` with no facility check. fadmin2 (fac 2) assigned dadmin (fac 1) — HTTP 200, written to DB. | *(API-only)* |
| **Test 3.1** Cross-facility API access | ❌ FAILED — Critical Security | `GET /api/admin-calendar?facility_id=N` not validated against user's allowed facilities. fadmin2 read 57 facility-1 events. | test_3_1_fadmin2_schedule.png, test_3_1_facility_id_bypass.png |
| **Test 6.1** DST transition | ❌ FAILED — Bug | RRule expansion uses current PDT offset for all dates. Pre-spring-forward events (Mar 8–9) display 1 hour early. | test_6_1_dst_week.png |
| **Test 6.19** Long reason text | ❌ FAILED — Bug | No client-side `maxLength` + no server-side validation → HTTP 500 on >255 char reason. | test_6_19_after_submit.png |
| **Test 4.11** Undo reschedule | ⚠️ UX Issue | No confirmation dialog — one click immediately and irreversibly reverts the event. | test_4_11_undo_button.png, test_4_11_after_undo.png |
| **Test 2.1** Calendar loading | ✅ PASSED | Events loaded, 0 console errors, API 200 OK. | test_2_1_schedule.png |
| **Test 2.3** Room conflict detection | ✅ PASSED | "Room Scheduling Conflict" dialog shown, HTTP 409. Re-confirmed with Facility 2 events. | test_2_3_room_conflict.png, test_2_3_fac2_conflict.png |
| **Test 3.2** Student vs admin access | ✅ PASSED | Student nav scoped correctly; `/schedule` redirects to `/home`; unauthenticated API → 401. | test_3_2_admin_actions.png |
| **Test 3.3** Same-facility instructor | ✅ PASSED | Toast "Instructor changed to mike jones", HTTP 200, dropdown scoped to same-facility users. | *(none)* |
| **Test 4.5** Cancel event modal | ✅ PASSED | Grey "Cancelled" badge, strikethrough styling, reason displayed, undo button shown. | test_4_5_cancel_ready.png, test_4_5_after_cancel.png |
| **Test 4.6** Reschedule single session | ✅ PASSED | Event moved to 1:00 PM–2:00 PM, "Undo Reschedule" button present. | test_4_6_reschedule_modal.png, test_4_6_after_reschedule.png |
| **Test 4.8** Uncancel series | ✅ PASSED (with note) | All 7 future events restored. No date picker — modal auto-selected the clicked event's date. | test_4_8_restore_modal.png, test_4_8_after_restore.png |
| **Test 5.5** Concurrent modifications | ✅ PASSED (with note) | Last write wins, no data corruption, consistent state after refresh. | test_5_5_tab1_cancelled.png, test_5_5_tab2_after_roomchange.png |
| **Test 6.20** Special characters | ✅ PASSED | No XSS, no SQL injection. Reason stored verbatim, rendered as literal text. | test_6_20_after_special_chars.png |
| **Test 1.2** Migration rollback | ✅ PASSED | `goose down` restored `program_classes.instructor_id` + removed 3 columns from `program_class_events`. `goose up` re-applied cleanly. Both directions clean in <11ms. | *(schema check only)* |
| **BUG: Seeder broken** `make seed` does not compile after migration 00067 | ❌ FAILED — PR-Caused Bug | `backend/seeder/main.go:646` references `InstructorName` removed by this PR's migration. `make seed` fails to compile on any DB with migration 00067 applied. | *(N/A — compile error)* |

---

## 🔍 Pre-Test Code Analysis

Findings from static analysis before testing began. Status reflects whether the issue still exists in the current code.

### 1. `frontend-v2/src/main.tsx:30` — Unsafe `Promise<any>` return
- **Issue**: `return res.json()` in the SWR fetcher returns `Promise<any>` with no type annotation
- **Status**: ⚠️ Still present — TypeScript lint warning, not a runtime bug

### 2. Migration 00067 — No validation before DROP
- **File**: `backend/migrations/00067_add_columns_to_program_class_events.sql`
- **Issue**: Migration copies `instructor_id` from `program_classes` to `program_class_events` via UPDATE, then immediately runs `ALTER TABLE program_classes DROP COLUMN IF EXISTS instructor_id` with no intermediate row-count or validation check. If the UPDATE silently affects 0 rows (e.g. no existing instructors), the DROP proceeds and data is unrecoverable.
- **Recommendation**: Add a validation query between the UPDATE and DROP to assert expected row counts match before committing the destructive step
- **Status**: ⚠️ Still present

### 3. ON DELETE SET NULL — UI handling of NULL instructor
- **Issue**: If an instructor user is deleted, `program_class_events.instructor_id` becomes NULL. Does the UI handle this gracefully?
- **Findings**: ✅ Handled correctly
  - `GetFacilityCalendar` uses `COALESCE(iu.name_first || ' ' || iu.name_last, '')` — returns empty string for NULL instructor (`class_events.go:682`)
  - `Schedule.tsx:498` guards with `{selectedEvent.instructor_name && (...)}` — instructor row simply does not render when blank
  - `SessionDetailSheet.tsx:83` initializes `instructor_name: ''` as safe fallback
- **Status**: ✅ Not an issue — resolved by existing code

### 4. Seeder compilation error — `InstructorName` field removed
- **File**: `backend/seeder/main.go:646`
- **Issue**: `InstructorName: faker.Name()` references a field that no longer exists on `models.ProgramClass`. Migration 00067 moved instructor tracking to the events layer; the seeder was not updated.
- **Status**: ❌ PR-caused bug — see "Bugs Caused by PR #1115" section above

### 5. `handleGetAdminCalendar` — No result limit (performance risk)
- **File**: `backend/src/database/class_events.go:674` (`GetFacilityCalendar`)
- **Issue**: No `LIMIT` clause. Query returns all events for a facility across the full requested date range. Frontend passes a ~15-month window. For a facility with many classes and long recurrence rules this could return an unbounded result set.
- **Recommendation**: Add pagination or a hard cap on returned events
- **Status**: ⚠️ Still present

### 6. Cross-facility instructor assignment — no facility check
- **File**: `backend/src/handlers/class_events.go:325`
- **Issue**: `handlePatchEventOverride` assigns `instructor_id` with no facility validation
- **Status**: ❌ Confirmed as live vulnerability — see Test 8.8 above

carddev81:
@CK-7vn this is fixed now, but note I couldn't get it to match exactly. Took a bit of tweaking but I think we can use it, please let me know if you do not like this and I can try changing it to match exactly. It is still a tiny bit off. I actually think how I changed it looks better than the figma:

image

carddev81:
@CK-7vn This is now a ticket for discussion:

I'm not sure if this is just a wording thing or if i'm thinking into it too much; But when we cancel a session/class, we have the option to "apply this change to all future sessions"; The button we have to restore sessions says "Restore all future sessions" Should we include a "Restore this and all future sessions" or something just so the user has clarification that they are also restoring the session thney are in/clicked on?

carddev81:
@CK-7vn Made another ticket for this, we need to be consistent so if we just change it here then all other pages will be sticky. I think we should talk about this one
image

carddev81:
@CK-7vn made the css changes requested here and also ticketed the other :

image

carddev81:
@CK-7vn This was intentional, since we now have the reschedule series addition this was missed on the figma, it should have been taken out:

image

carddev81:
@CK-7vn Ticketed this one ([here](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214430813078159?focus=true)):


**I also noticed, that there is no room-conflict detection; I'm not sure if that is out of the scope for this PR; if it is then just disregard this comment, but when changing the room of a class/session in direct conflict with another session on the Schedule it did not say anything, or reject the request;**

carddev81:
@CK-7vn the following is fixed:

The same issue with the "x" button in the Reschedule series modal;

carddev81:
@CK-7vn Nice catch, this was fixed
The spacing on the bottom buttons of the reschedule modal are spaced differently than the figma.

carddev81:
@CK-7vn good eyes, this was fixed
The Take Attendance icon is different

carddev81:
@CK-7vn Some of your issues that were found in this PR were moved over into tickets, here are the list of ticket links that were created based on your concerns. Good findings!!

- [Restore single instance versus all future sessions (schedule/schedule tab)](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214434555944310)
- [Room Conflict Detection (Issue or No Issue)](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214430813078159)
- [Schedule ?'s and Wonders](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214434555944314)
- [Bottom of Agenda isn't Viewable (Schedule) Figma/Current Code](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214434555944306)
- [Change Instructor Issues?](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214430813078163)
- [Sticky Header Or No Sticky Header?](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214430813078157)
- [Add restore all sessions -- too much? or should we do this](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214430813078155)
- [add instructor to reschedule modal > schedule tab/schedule](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214367464623386)

CK-7vn (approved):
Looks good on my end man; Only thing I will note 

Since we now have the pre-reschedule modal that asks if the user wants to reschedule that sessions only or all future sessions as well, we need to probably make a ticket for the Class Details rescheduling, that currently does not have that modal implemented, but now no longer has the required checkbox to ask if we want to reschedule future sessions as well. 

I also did not review what Cory noted/mentioned just for reference.

corypride (approved):
Good job with this one. A small fix and this one is good to go for me! The bug was not related to your PR I don't believe, so it is not a blocker for me. 
# PR #1115 — Re-Review Test Results

## Bugs

### Bug 1: Concurrent override returns 500 instead of 409
- **Test**: 5.5
- **Severity**: Medium
- **Related to PR**: Yes
- **Description**: Two concurrent PATCH requests for the same `(event_id, date)` both pass the `First()` existence check before either commits. The second `Create()` hits the unique constraint on `program_class_event_overrides(event_id, override_rrule)` and surfaces as a generic 500 instead of 409 Conflict.
- **Repro**: Fire two simultaneous `PATCH /api/program-classes/{id}/events/{id}` with identical `date` and `is_cancelled: true`. One will return 500.

### Bug 2: Reason field silently truncates — no length validation
- **Test**: 6.19
- **Severity**: Low
**- **Related to PR**: No (pre-existing)**
- **Description**: `patchEventOverrideRequest.Reason` has no length guard. The `program_class_event_overrides.reason` column is `VARCHAR(255)` but the handler passes the value through without validation. 310 chars in → 254 chars stored, HTTP 200 returned — silent data loss with no error to the caller.
- **Repro**: Submit a cancel override with `reason` of 256+ characters. DB shows truncated value; response is 200 OK.

---

**Completed**: 2026-05-06
**Summary**: 13 passed · 1 skipped · 2 failed · 2 bugs (1 PR-related, 1 pre-existing)

carddev81:
@corypride fixed bug #2, bug 1--I don't think that 2 users would do the fire off the same endpoint for the same resident at the same time. With that said we could just create a ticket for this one to create a block for this

PR #1121, opened 2026-05-04
fix: logic in program_classes.go hook that prevented enrolled_ended_a…

corypride:
## Problem

When enrollment status was changed to a terminal state (`Completed`, `Cancelled`, `Incomplete:*`, etc.), the `enrollment_ended_at` field should have been automatically set to track when the enrollment ended.

However, it was never being set, leaving the field as `NULL` even after enrollments were completed, cancelled, or withdrawn.

---

## Root Cause

The `BeforeUpdate` GORM hook in `ProgramClassEnrollment` (`backend/src/models/program_classes.go`) had a buggy early return condition that prevented `enrollment_ended_at` from ever being set:

```go
// BUGGY CODE (line 180):
if !tx.Statement.Changed("enrollment_status") ||
    tx.Statement.Changed("enrolled_at") ||
    tx.Statement.Changed("enrollment_ended_at") { // ← BUG
    return nil
}
```

### Original intent

The third condition was meant to allow calling code to explicitly override `enrollment_ended_at` if it wanted to set a custom timestamp.

### The reality

No code ever explicitly sets `enrollment_ended_at` — the hook is the **only** place that sets it.

So every time enrollment status changed:

1. The hook detected that `enrollment_ended_at` would change (`NULL → timestamp`)
2. The hook assumed calling code was explicitly setting it and returned early to avoid overriding
3. `enrollment_ended_at` was never set

---

## How We Confirmed This

Here is the code evidence showing that no database handler or API function ever explicitly sets `enrollment_ended_at`.

### 1. `GraduateEnrollments()` — when changing status to `Completed`

```go
if err = tx.Model(&models.ProgramClassEnrollment{}).
    Where("user_id IN (?) AND class_id = ?", userIds, classId).
    Set("class_id", classId).
    Update("enrollment_status", models.EnrollmentCompleted).Error; err != nil {
    // Only updates enrollment_status
    // enrollment_ended_at NOT set here
}
```

### 2. `UpdateProgramClassEnrollments()` — when changing status to other values (`Cancelled`, `Incomplete:*`, etc.)

```go
updates := map[string]any{
    "enrollment_status": status,
    // change_reason and update_user_id also updated
    // enrollment_ended_at NOT in this map
}

if err := db.Model(&models.ProgramClassEnrollment{}).
    Where("class_id = ? AND user_id IN (?)", classId, userIds).
    Set("class_id", classId).
    Updates(updates).Error; err != nil {
    // Only updates the map values
    // enrollment_ended_at NOT set here
}
```

### 3. `UpdateProgramClasses()` — when class status changes

```go
if err := tx.
    Model(&models.ProgramClassEnrollment{}).
    Where("class_id = ? AND enrollment_status = ?", classID, models.Enrolled).
    Set("class_id", classID).
    Update("enrollment_status", enrollmentStatus).
    Error; err != nil {
    // Only updates enrollment_status
    // enrollment_ended_at NOT set here
}
```

### 4. The only place `enrollment_ended_at` is set — the hook (`program_classes.go`)

```go
// Clear enrollment_ended_at when status becomes Enrolled (reactivation)
if newEnrollmentStatus == Enrolled && (classStatus == Active || classStatus == Paused) {
    tx.Statement.SetColumn("enrollment_ended_at", nil)
}

// Set enrollment_ended_at for terminal statuses
if IsTerminalEnrollment(newEnrollmentStatus) && (classStatus == Active || classStatus == Paused) {
    tx.Statement.SetColumn("enrollment_ended_at", time.Now().UTC())
}
```

---

## Conclusion

✅ Confirmed: Across **all** database functions that update enrollment status, **none** explicitly set `enrollment_ended_at`.

The GORM `BeforeUpdate` hook is the **only** place in the codebase that sets this field.

That is why the buggy condition:

```go
tx.Statement.Changed("enrollment_ended_at")
```

was always triggering.

The hook detected that `enrollment_ended_at` would change (`NULL → timestamp`) and incorrectly assumed calling code was setting it explicitly, so it returned early without actually setting it.

---

## Solution

- Removed the buggy `tx.Statement.Changed("enrollment_ended_at")` condition from line 180
- Added logic to clear `enrollment_ended_at` when status changes from terminal back to `"Enrolled"` (reactivation scenario)
- Fixed the hook to handle both `int` and `uint` types for the `class_id` context value

---

### `backend/src/models/program_classes.go`

Updated the `BeforeUpdate` hook to:

- Remove the buggy early return condition
- Clear `enrollment_ended_at` on reactivation to `"Enrolled"`
- Fix type handling for the `class_id` context value

---

## Testing

- ✅ All existing enrollment integration tests pass (`TestUpdateProgramClassEnrollments`)
- ✅ Hook correctly sets `enrollment_ended_at` for all terminal statuses:
  - `Completed`
  - `Cancelled`
  - `Incomplete: Withdrawn`
  - `Incomplete: Dropped`
  - `Incomplete: Failed to Complete`
  - `Incomplete: Transfered`
  - `Incomplete: Segregated`
- ✅ Hook clears `enrollment_ended_at` when status changes back to `"Enrolled"` (reactivation)
- ✅ All integration tests pass (`2.209s` total)

---

## Affected Areas

This fix ensures `enrollment_ended_at` is correctly set in all scenarios where enrollment status changes via:

- `PATCH /api/program-classes/{class_id}/enrollments` — individual and bulk status changes
- `GraduateEnrollments()` — completion flow
- `UpdateProgramClasses()` — class-level status changes affecting all enrollments

All UI enrollment status changes now correctly populate `enrollment_ended_at` when an enrollment moves to a terminal state.

---
  - https://app.asana.com/0/0/1213790719514971


carddev81:
@coderabbitai review

carddev81 (approved):
Looks great and tested great.

NOTE: Asana ticket will be created for when class is set to Active the date of the active setting will be the date set for enrolled_at

PR #1130, opened 2026-05-14
Cpride/pagination

corypride:

- [x] Functions according to ticket specifications

## Description of the change

Centralizes pagination state management site-wide by migrating all paginated pages from local `useState` to the `useUrlPagination` hook, so pagination state is preserved in the URL and survives browser refresh and back navigation.

- Deleted the degraded `components/shared/Pagination.tsx` duplicate; re-exported `components/Pagination.tsx` from the shared index so all consumers get the canonical component
- Added an optional `prefix` param to `useUrlPagination` for pages with multiple independent paginators (e.g. `activity_page`, `history_page`, `audit_page`)
- Added `instanceId` prop to `Pagination.tsx` to keep select element IDs unique when multiple instances appear on one page
- Replaced bare prev/next buttons in `ClassEnrollmentDetails` and `AddClassEnrollments` with the full `Pagination` component

**Related issues**: [Asana — Split 4: Pagination](https://app.asana.com/...)

## Screenshot(s)

test_3_7_history_page2_persists
test_3_8_audit_page2_persists
test_3_4_kcm_tab_switching
test_2_4_add_class_enrollments_pagination
test_2_3_class_enrollment_details_pagination
test_2_2_dark_mode_active_page
test_2_1_admin_management_pagination

## Additional context

**Bug found and fixed during testing — `per_page` dropped from URL when changing items-per-page:** The `Pagination.tsx` items-per-page `onChange` handler was calling both `onItemsPerPageChange(value)` and `onPageChange(1)`. Both closures captured stale `searchParams`, so the second `setSearchParams` overwrote the first, silently dropping `per_page` from the URL. Fixed by removing the redundant `onPageChange(1)` — `setPerPage` in `useUrlPagination` already resets page to 1 internally.

**Bug found and fixed during testing — `activity_page` resets to 1 on every mount:** `ActivityHistoryCard` had a `useEffect(() => setPage(1), [filterQuery])` that fired on every mount. Before this PR it reset a local `useState` value harmlessly; after migration it called `setSearchParams`, overwriting `activity_page` back to 1 on every page load. Fixed by adding a `useRef(true)` first-render guard so the reset only fires when `filterQuery` actually changes from a user action.

---
  - https://app.asana.com/0/0/1214782085182660


carddev81 (changes requested):
When selecting to display more on the page it does not display what is selected. See page below, holding off on the review till this is corrected since this will reflect all the changes made.

To reproduce, just go to the residents listing and select 40 from the items to display

corypride:
https://github.com/user-attachments/assets/3b89dccd-ed77-4a28-923f-1b353f7fa0e6

@carddev81 Could not reproduce as you stated. Send me a video

carddev81 (approved):
@corypride I don\t understand why I am not getting that error I posted, but here are the only findings I could spot. Great Job.

Audit History (Please check all locations) - See video--After selecting to display more than 20 records at a time the pagination controls disappear.

https://github.com/user-attachments/assets/f3bbaf09-4474-491d-b686-5033c75a8284

## Linting errors:
1. ClassesPage.tsx (Line 219): error is -> React Hook useEffect has a missing dependency: 'setCurrentPage'. Either include it or remove the dependency array
2. KnowledgeCenterManagement.tsx (344): error is -> React Hook useEffect has a missing dependency: 'setCurrentPage'. Either include it or remove the dependency array.

## Inconsistency
The ClassesPage.tsxk (line 50) imports Pagination from @/components/shared while every other location it is needed imports from @/components/Pagination. The re-export in shared/index.ts makes it work, but it's inconsistent. Either update ClassesPage or delete the re-export. I say delete the re-export because it's not even being used by any other page.

## Please fix these issues and merge your code into the total refactor branch please.

corypride:
🫡


*Cory Pride*
Associate Engineer
unlockedlabs.org

PR #1142, opened 2026-05-27
Ck 7vn/serial patch

CK-7vn:

## Description of the change
Two issues in asana ticket ID-675, 
Issue A - sequential PATCH loops, all four bulk-action modals were iterating for..of and awaited a single PATCH per session, so with a 30 session patch, you're making 30 round trips. We fix this by introducing a small bulkPatchEvents helper that fires all patches in parallell via Promise.allSettled and returns {ok, fail} counts. 

Issue B - duplicated Instructor/Room modals - Both modals were essentially 95% the same. Extracted a shared BulkSessionFieldModal that owns the form shell, sessions list, and future-session checkbox, and submit loop. Then used two wrappers to shrink to 50 lines each and pass config. -193 lines total. 

While testing the changes in the browser I ran into two different scoping issues.
 
- ChangeinstructorModal was querying /api/users?role=${user.role} - the logged-in user's role rather than instructor candidates for the class's facility. The backend requires a facility_id as well, the backend was using the QueryContext, so if a department admin was working on a class outside of the facility they were currently scoped too, there would be discrepancies in the queried instructors. 
- 
- EnrollResidentsModal had the same shape of bug for department admins viewing classes outside of their scoped facility. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1215038908589222?focus=true

### DIFF
--- a/frontend/src/pages/class-detail/BulkCancelSessionsModal.tsx
+++ b/frontend/src/pages/class-detail/BulkCancelSessionsModal.tsx
@@ -11,7 +11,7 @@ import {
     SelectValue
 } from '@/components/ui/select';
 import { FormModal, TonedPanel } from '@/components/shared';
-import API from '@/api/api';
+import { bulkPatchEvents } from '@/api/bulkPatchEvents';
 import { toast } from 'sonner';
 
 export interface BulkCancelSession {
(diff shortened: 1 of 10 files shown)


frontend/src/pages/class-detail/BulkCancelSessionsModal.tsx
carddev81, line 1: 100 percent better

carddev81 (approved):
Tested good and great fixes!

PR #1149, opened 2026-06-02
chore: apply prettier formatting and array-type lint fix

CK-7vn:
This PR just runs prettier on the files that my other PR/branch for the room scoping touches because prettier made a bunch of formatting changes and it would have made it impossible for the reviewer to review without a huge headache.


carddev81 (approved):
Loooks like formatting changes....Looks pretty

PR #1152, opened 2026-06-02
fix: add logic for when the doc id does not exist and fix form per ticket id 674

carddev81:

- [X] No debug/console/fmt.Println statements
- [X] Unnecessary development comments removed
- [X] All acceptance criteria verified
- [X] Functions according to **ticket specifications**
- [X] Tested manually where applicable
- [X] Branch rebased with latest main
- [X] No business logic exists within the database layer

## Description of the change

Added custom conditional logic for when DOC ID does not exist the user name is used instead. Also fixed form to use onSubmit rather than button 

- **Related issues**: Link to related Asana ticket that this closes: [Confirm-by-typing UX gaps in resident dialogs](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1215036799520096?focus=true)

### DIFF
--- a/frontend/src/components/residents/ResidentDialogs.tsx
+++ b/frontend/src/components/residents/ResidentDialogs.tsx
@@ -90,56 +90,66 @@ export function EditResidentDialog({
             description="Update resident information"
             titleClassName="text-foreground"
         >
-            <div className="space-y-4 py-4">
-                <div className="grid grid-cols-2 gap-4">
+            <form
+                onSubmit={(e) => {
+                    e.preventDefault();
+                    void form.handleSubmit((d) => void handleSave(d))(e);
+                }}
+            >
+                <div className="space-y-4 py-4">
+                    <div className="grid grid-cols-2 gap-4">
+                        <div>
+                            <Label htmlFor="edit-first">First Name</Label>
+                            <Input
+                                id="edit-first"
+                                {...form.register('name_first', {
+                                    required: true
+                                })}
+                                className="mt-2"
+                            />
+                        </div>
+                        <div>
+                            <Label htmlFor="edit-last">Last Name</Label>
+                            <Input
+                                id="edit-last"
+                                {...form.register('name_last', {
+                                    required: true
+                                })}
+                                className="mt-2"
+                            />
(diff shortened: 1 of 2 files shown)


corypride (approved):
## 🐛 Bugs Caused by PR #1152

_No PR-caused bugs identified during pre-test analysis._ Code rabbit had a good suggestion that should be made, and this PR is good to go

---

## 🐛 Discovered Bugs — Unrelated to PR #1152

### Bug: `resident-verify` returns 404 when resident has no doc_id
- **Severity**: Medium
- **Blocker**: No — this is a pre-existing bug unrelated to PR #1152. The PR can be merged as-is.
- **Description**: `GET /api/users/resident-verify?user_id=X&facility_id=Y&doc_id=` returns 404 because `GetUserByDocIDAndID` fails on an empty `doc_id`. Surfaced when opening the Transfer dialog for a resident without a doc_id and selecting a destination facility.
- **Repro**: Transfer dialog for any resident with `doc_id = NULL` → select destination facility → 404 in browser console.
- **Location**: `backend/src/database/users.go` — `GetUserByDocIDAndID` always queries `WHERE LOWER(doc_id) = ''` even when doc_id is empty, which never matches a NULL column.
- **Recommendation**: When `doc_id` is an empty string, fall back to looking up by `user_id` alone. The endpoint's purpose is to verify the resident exists and check program transfer conflicts — `doc_id` is just a lookup key, not a security boundary.
  ```go
  // backend/src/database/users.go
  func (db *DB) GetUserByDocIDAndID(ctx context.Context, docID string, userID int) (*models.User, error) {
      user := models.User{}
      query := db.WithContext(ctx)
      if docID != "" {
          query = query.Where("LOWER(doc_id) = ? AND id = ?", strings.ToLower(docID), userID)
      } else {
          query = query.Where("id = ?", userID)
      }
      if err := query.First(&user).Error; err != nil {
          return nil, newNotFoundDBError(err, "users")
      }
      return &user, nil
  }
  ```

---

## 🗓️ Testing completed on:
- [x] Date: 2026-06-04

---

## 📊 Overall Results

GTG!

PR #1154, opened 2026-06-03
fix: remove facility toggle adjust lingering facility toggle 

CK-7vn:

## Description of the change
Removes the top facility-switcher toggle and re-scopes how facility context is resolved so that data is no longer silently filtered by an ambient session facility. 

  - **Frontend**
    - Removed the top facility toggle from `TopNav` and the now-dead facilities fetch in `AuthenticatedLayout`.
    - "View at Facility" drill-in is now **stateless** — it navigates with `?facility_id=N` instead of mutating session state.
    - Department/Super admins default to **statewide** views; facility selection is explicit where a write needs it (class/room creation send
  `?facility_id`; bulk-import and Add-Admin now require a facility selection for all roles).
  - **Backend**
    - `getQueryContext` returns statewide (`FacilityID = 0`) for switch-capable admins with no `facility_id`; facility admins stay pinned to
  their own facility.
    - Removed the `PUT /api/admin/facility-context/{id}` session-switch endpoint; `auth.go` no longer honors an alternate Kratos facility
  trait (claims are pinned to the user's home facility).
    - Write handlers (class, room, single/bulk user, provider import) derive the target facility from the request and reject (`400`) when a
  switch-capable admin omits it (`requireFacilityID` helper).
    - Reads whose visibility is inherently per-facility (open content, calendar, helpful links, resident-for-class lists) fall back to the
  home facility via `facilityScopedQueryContext`.

## Screenshot(s)
image

## Additional context

A lot of these lines are actually from prettier, so if this gets reviewed before we end up merging a prettier PR, then you can append `?w=1` to the "Files changed" URL and it will remove the formatting lines and leave just the logic for review. 

---
  - https://app.asana.com/0/0/1215131303729099

### DIFF
--- a/backend/src/handlers/classes_handler.go
+++ b/backend/src/handlers/classes_handler.go
@@ -16,17 +16,19 @@ func (srv *Server) registerClassesRoutes() []routeDef {
 	resolver := FacilityAdminResolver("program_classes", "class_id")
 	validateFacility := func(check string) RouteResolver {
 		return func(tx *database.DB, r *http.Request) bool {
-			claims := r.Context().Value(ClaimsKey).(*Claims)
-			if claims.canSwitchFacility() {
-				if check == "" {
-					return true
-				}
+			// Facility comes from getQueryContext: the ?facility_id= param for
+			// switch-capable admins, else their home facility. A statewide admin
+			// (facilityID == 0) is allowed for facility-agnostic reads but must
+			// target a specific facility for facility-scoped operations.
+			facilityID := srv.getQueryContext(r).FacilityID
+			if facilityID == 0 {
+				return check == ""
 			}
 			var count int64
 			return tx.WithContext(r.Context()).
 				Table("programs").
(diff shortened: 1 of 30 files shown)


backend/src/handlers/classes_handler.go
```diff
@@ -16,17 +16,19 @@ func (srv *Server) registerClassesRoutes() []routeDef {
...
-			claims := r.Context().Value(ClaimsKey).(*Claims)
-			if claims.canSwitchFacility() {
-				if check == "" {
-					return true
-				}
+			// Facility comes from getQueryContext: the ?facility_id= param for
+			// switch-capable admins, else their home facility. A statewide admin
+			// (facilityID == 0) is allowed for facility-agnostic reads but must
+			// target a specific facility for facility-scoped operations.
+			facilityID := srv.getQueryContext(r).FacilityID
```
carddev81, line 23: tested good

carddev81 (approved):
Looks great and tested good. I made a couple changes and pushed to this code base. Rooms were not being pulled correctly for the facility--fixed all associated issues

carddev81:
need to fix all conflicts and then merge

PR #1157, opened 2026-06-03
chore: run prettier on frontend

CK-7vn:
This PR just runs prettier on the frontend so that we don't continue to get PR's filled with white-sense and formatting changes. This should minimize confusion on reviews on what was changed.


carddev81 (approved):
OMG, so many formatting changes.

PR #1161, opened 2026-06-05
fix: scope eligible residents query to class facility (ID-686)

corypride:
## Fix

Made the backend authoritative: changed the `WHERE` clause in `GetEligibleResidentsForClass` to filter on `c.facility_id` (the class's own facility via the already-joined `program_classes` table) instead of `args.FacilityID`.

This holds regardless of caller role or what `facility_id` is passed in the request.

### `backend/src/database/users.go`
- One-line filter change in `GetEligibleResidentsForClass`

### `backend/tests/integration/class_enrollments_test.go`
- Added `TestEligibleResidentsFilteredByClassFacility` covering:
  - Only unenrolled facility-A students are returned
  - Already-enrolled residents are excluded
  - Deactivated residents are excluded
  - Facility-B residents are excluded even when a department admin passes `facility_id=B`

### `frontend/src/pages/class-detail/EnrollResidentsModal.tsx`
- Already passes `facility_id=${classFacilityId}`
- Kept as-is, but it is no longer the security boundary

### `getQueryContext` facility logic
- Other endpoints depend on it
- Left unchanged

### Integration Test
- Verifies facility isolation is enforced independently of the request's `facility_id`

### Manual Verification
1. Log in as a department admin
2. Open a class assigned to facility A
3. Click **Add Resident**
4. Verify that only facility-A residents are shown

---
  - https://app.asana.com/0/0/1215131303729115

### DIFF
--- a/backend/src/database/users.go
+++ b/backend/src/database/users.go
@@ -185,7 +185,8 @@ func (db *DB) GetEligibleResidentsForClass(args *models.QueryContext, classId in
 		Joins("JOIN facilities_programs fp ON users.facility_id = fp.facility_id").
 		Joins("JOIN program_classes c ON c.program_id = fp.program_id AND c.id = ?", classId).
 		Where("pse.user_id IS NULL"). //not enrolled in class
-		Where("users.role = 'student' AND users.facility_id = ?", args.FacilityID).
+		Where("users.role = 'student'").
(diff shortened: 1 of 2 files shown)


carddev81 (approved):
Tested Good and code looks great

PR #1165, opened 2026-06-08
Feat/provider platform rebased

gtonye:

- [ ] No debug/console/fmt.Println statements
- [ ] Unnecessary development comments removed
- [ ] All acceptance criteria verified
- [ ] Functions according to **ticket specifications**
- [ ] Tested manually where applicable

## Description of the change

This PR introduces full Canvas provider platform integration — surfacing Canvas courses as programs, merging Canvas calendar events into the admin schedule view, and providing facility admins with tooling to link (map) residents to their Canvas accounts.

**Canvas Programs via live API lookup**

When the Programs overview table is fetched, the backend performs a live lookup against every enabled Canvas provider platform and synthesizes a `ProgramsOverviewTable` entry per provider. These entries appear alongside native UnlockEd programs but carry `source: "canvas"` and a stable synthetic ID (`CanvasProgramIDOffset + providerID`). The facility admin program view shows a read-only banner for Canvas programs and suppresses edit controls accordingly.

```mermaid
sequenceDiagram
    participant Client
    participant API as GET /api/programs
    participant NATS as NATS KV (canvas_programs)
    participant Canvas as Canvas API

    Client->>API: GET /api/programs
    API->>API: query DB for native programs
    loop each enabled Canvas provider
        API->>NATS: get cache key canvas_program_{id}
        alt cache hit (< 5 min old)
            NATS-->>API: cached ProgramsOverviewTable
        else cache miss / stale
            API->>Canvas: GET /accounts/{id}/courses?per_page=100
            Canvas-->>API: course list
            API->>NATS: put canvas_program_{id} (TTL 5 min)
            NATS-->>API: ok
        end
    end
    API-->>Client: native programs + Canvas programs merged
```

Results are persisted in a dedicated `canvas_programs` NATS KV bucket with a 5-minute TTL. On a cache miss the handler calls Canvas directly, then writes the result back. Individual program routes (`/programs/:id`, archive-check, history) are guarded by the same `CanvasProgramIDOffset` sentinel and return safe no-op responses for Canvas IDs, so the frontend does not need special-casing beyond the read-only banner.

**Canvas User Matching**

Facility admins must link residents before Canvas enrollment data appears on resident profiles. Canvas users and UnlockEd residents are matched automatically using a composite name-similarity score (Jaro-Winkler 60% + Smith-Waterman-Gotoh 40%):

- **Auto-confirmed** (score ≥ 0.90): mappings are applied without admin review.
- **Ambiguous** (0.50–0.89): surfaced in the Provider User Management UI for admin confirmation.
- **Unmatched** (< 0.50): the admin can manually select an existing resident or create a new one.

Until mappings exist, Canvas enrollment counts reflect 0 mapped users.

**Canvas calendar events on admin schedule**

Canvas calendar events are fetched (paginated via the `Link` header) and merged into the admin calendar. Canvas events render in blue (`blue-700`) and show a "Canvas" badge on the event tile. Clicking a Canvas event opens a `CanvasEventSheet` (separate from the existing `SessionDetailSheet` used for native sessions).

**Other changes**

- `enrollment_types` field added to `ProviderPlatform` (new migration, `[]string` serialized as JSON).
- OIDC client recreation endpoint added for Canvas OIDC re-registration flows.
- Access key is now whitespace-trimmed after decryption (`AfterFind` hook) to avoid auth failures from keys pasted with trailing newlines.
- New `ProviderPlatformManagement` and `ProviderPlatformDetail` admin pages wired into sidebar navigation under a `ProviderAccess` feature flag.

## Screenshot(s)

Screenshots of the different screens are available in [this document](https://docs.google.com/presentation/d/1b8i2V5YmC5bZ4LyrcWRX4tq17fjcYgkrB9v7DxQzgSk/edit?usp=sharing).

### Local setup

To test locally, you will need a running Canvas instance.

You can use the Docker-based one:

1. Clone the repository: https://github.com/instructure/canvas-lms
2. Install Docker and Docker Compose if not already installed.
3. Install Ruby if not already installed.
4. Build the application and start it with Docker Compose. **Note:** Because UnlockEd also uses a Postgres instance, configure the Canvas Docker Postgres to bind to a dedicated port to avoid conflicts.

### Technical notes

- Canvas program IDs use an offset (`100_000_000`) to avoid collisions with native program IDs. All routes that accept a program ID check this sentinel to route to Canvas-specific handlers. Reviewers should verify this boundary is consistent across the frontend (`isCanvasProgram = id >= 100_000_000`) and backend (`CanvasProgramIDOffset`).
- Facility admins must link residents before Canvas data is visible — this is not automated on first sync. The matching UI is the primary workflow for this.
- The greedy bipartite 1:1 matching algorithm ensures a Canvas user and an UnlockEd resident can only appear in one confirmed match pair, preventing duplicate mappings.
- **Known gap:** Canvas course pagination is currently capped at 100 per page. Providers with more than 100 courses will have all pages fetched, but the first live call may be slow before the cache is warm.

#### High-Level Architecture and Philosophy

The Provider system uses an abstraction layer that separates provider logic from the main UnlockEd codebase.

This comes with tradeoffs to consider:
1. It assumes that future providers share a common set of behaviors.
2. New providers must inherit from the base provider class.

This may require drilling into provider APIs and data structures to map them against UnlockEd data models. We have done that mapping work here: https://docs.google.com/spreadsheets/d/15BZiP_TyY-Cg2LqoM22GkLnEK3lU_2UobL64EzukgyM/edit?usp=sharing

#### Program Concept

Canvas is course-based, so building a program dynamically will always involve latency, as the Canvas API is designed to be consumed per course.

We have implemented parallel calls combined with caching to mitigate this, but a Canvas instance with a large number of courses will still present challenges.

Persisting the data would help, but any persistence strategy needs to account for keeping data in sync with Canvas. The provider code includes logic for background jobs to support this.

#### Timezone

We have noticed some discrepancies in time handling. While storing times in UTC is generally good practice, we have seen in other projects that storing the timezone alongside the UTC timestamp can help with display reconciliation on the frontend.

We have also observed that frontend code sometimes uses UTC without applying the user's timezone (see Program Overview and the class header).

For Canvas, we propagate the timezone from the course all the way to the frontend to ensure consistent display between Canvas and UnlockEd.

One suggestion would be to display the timezone to the user to avoid confusion — this is especially worthwhile if users spanning multiple timezones is a realistic scenario.


frontend/src/pages/learning/ResidentOverview.tsx
```diff
@@ -134,10 +135,11 @@ export default function ResidentOverview() {
...
                                                 <td className="py-3 px-5 text-muted-foreground">
-                                                    {prog.attendance_percentage !==
-                                                    undefined
-                                                        ? `${Math.round(prog.attendance_percentage)}%`
-                                                        : '-'}
+                                                    {prog.is_canvas
+                                                        ? '-'
+                                                        : prog.attendance_percentage !== undefined
+                                                          ? `${Math.round(prog.attendance_percentage)}%`
+                                                          : '-'}
```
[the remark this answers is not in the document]
gtonye: This behavior pre-existed the Canvas work and should be resolved with the upcoming rework of the residents' program page.

frontend/src/pages/admin/ProviderPlatformManagement.tsx
```diff
@@ -0,0 +1,384 @@
...
+    ProviderResponse,
+    ServerResponseMany
+} from '@/types';
+import { useAuth, hasFeature } from '@/auth/useAuth';
+import { FeatureAccess } from '@/types';
+import API from '@/api/api';
+import { PageHeader } from '@/components/shared/PageHeader';
+import { EmptyState } from '@/components/shared/EmptyState';
+import { FormModal } from '@/components/shared/FormModal';
+import { EnrollmentTypeSelector } from '@/Components/shared/EnrollmentTypeSelector';
```
carddev81: @gtonye issue with this component not existing. Had to remove to get application to run sucessfully

frontend/src/pages/admin/ProviderPlatformDetail.tsx
```diff
@@ -0,0 +1,698 @@
...
+    DropdownMenuTrigger
+} from '@/components/ui/dropdown-menu';
+import { toast } from 'sonner';
+import {
+    Accordion,
+    AccordionContent,
+    AccordionItem,
+    AccordionTrigger
+} from '@/components/ui/accordion';
+import { EnrollmentTypeSelector } from '@/Components/shared/EnrollmentTypeSelector';
```
carddev81: @gtonye issue with this component not existing. Had to remove to get application to run sucessfully

frontend/src/pages/admin/ProviderPlatformDetail.tsx
```diff
@@ -0,0 +1,698 @@
...
+    const [showOidcInfoModal, setShowOidcInfoModal] = useState(false);
+    const [showArchiveConfirm, setShowArchiveConfirm] = useState(false);
+    const [showRecreateConfirm, setShowRecreateConfirm] = useState(false);
+    const [oidcClient, setOidcClient] = useState<OidcClient | null>(null);
+
+    const {
+        data: providerResp,
+        mutate,
+        isLoading,
+        error
```
carddev81, line 78: @gtonye linting error on line 79: `Unsafe object destructuring of a property with an `any` value.`

frontend/src/pages/admin/ProviderPlatformManagement.tsx
```diff
@@ -0,0 +1,384 @@
...
+export default function ProviderPlatformManagement() {
+    const { user } = useAuth();
+    const navigate = useNavigate();
+    const [showAddModal, setShowAddModal] = useState(false);
+    const { page, perPage, setPage, setPerPage } = useUrlPagination(1, 20);
+
+    const {
+        data: providers,
+        mutate,
+        error,
```
carddev81, line 50: @gtonye linting error on line 52: `Unsafe object destructuring of a property with an `any` value.`

frontend/src/pages/admin/ProviderPlatformManagement.tsx
carddev81: researching this a bit, the problem was the capital C on Components--after changing to @/components/shared/EnrollmentTypeSelector, everything worked as expected

frontend/src/pages/admin/ProviderPlatformDetail.tsx
carddev81: researching this a bit, the problem was the capital C on Components--after changing to @/components/shared/EnrollmentTypeSelector, everything worked as expected

carddev81:
@gtonye when adding a learning platform, I accidentally clicked off of the `Add Learning Platform` modal the modal closed unexpectedly. Can we make the modal stay open if the user clicks any area outside of the modal. 

Also, curious what the difference between the selection `Canvas Cloud` and `Canvas OSS`? If there are no underlying functional differences could we remove one of the selections for Canvas?

carddev81:
@gtonye I expected 'http://localhost:3001/api/v1/accounts/1/users?per_page=1000&enrollment_type[]=student' to return only enrolled students but this endpoint also returned TAs and teachers. Just pasted the output here for example:


[{"id":1,"name":"rich@unlockedlabs.org","created_at":"2026-06-10T10:01:15-06:00","sortable_name":"rich@unlockedlabs.org","short_name":"rich@unlockedlabs.org","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"rich@unlockedlabs.org"},{"id":4,"name":"Sam Student","created_at":"2026-06-10T10:10:42-06:00","sortable_name":"Student, Sam","short_name":"Sam Student","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"student1@example.com"},{"id":5,"name":"Sara Student","created_at":"2026-06-10T10:10:42-06:00","sortable_name":"Student, Sara","short_name":"Sara Student","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"student2@example.com"},{"id":8,"name":"Selena Student","created_at":"2026-06-10T10:19:50-06:00","sortable_name":"Student, Selena","short_name":"Selena Student","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"student4@example.com"},{"id":9,"name":"Sidney Student","created_at":"2026-06-10T10:19:50-06:00","sortable_name":"Student, Sidney","short_name":"Sidney Student","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"student5@example.com"},{"id":10,"name":"Simon Student","created_at":"2026-06-10T10:19:50-06:00","sortable_name":"Student, Simon","short_name":"Simon Student","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"student6@example.com"},{"id":6,"name":"Steve Student","created_at":"2026-06-10T10:10:42-06:00","sortable_name":"Student, Steve","short_name":"Steve Student","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"student3@example.com"},{"id":3,"name":"Tina TA","created_at":"2026-06-10T10:10:42-06:00","sortable_name":"TA, Tina","short_name":"Tina TA","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"ta@example.com"},{"id":2,"name":"Terry Teacher","created_at":"2026-06-10T10:10:41-06:00","sortable_name":"Teacher, Terry","short_name":"Terry Teacher","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"teacher@example.com"},{"id":7,"name":"Tom Teacher","created_at":"2026-06-10T10:19:50-06:00","sortable_name":"Teacher, Tom","short_name":"Tom Teacher","sis_user_id":null,"integration_id":null,"sis_import_id":null,"login_id":"teacher2@example.com"}]

carddev81:
@gtonye when selecting Confirm button on one user within the list, all confirm buttons are updated to Confirmed. I believe this should only change to Confirmed for the single button that I click. See video: 


https://github.com/user-attachments/assets/2366077c-67ad-4e58-9a3c-3dcbf878b896

carddev81:
@gtonye I tried out the Create resident and is not functioning as expected. See video:


https://github.com/user-attachments/assets/502790a9-4c2f-4d02-a0e1-632de8e04839

carddev81:
@gtonye after selecting a single resident to link to Canvas all the 'unmatched' and 'needs review' users list disappear and I am unable to map any more users. I clicked refresh and lists are still empty. 

https://github.com/user-attachments/assets/dd5dc5b0-2e5b-4716-8678-bc1734ea9697

carddev81:
@gtonye just noting this down:  After linking a resident to a canvas user, I noticed there was no way to view which canvas user is tied to unlocked users. For instance which canvas user is the following linked to:

image

carddev81:
@gtonye 

Canvas/Resident User is enrolled within all classes that were created but does not show as an enrollment within the listing. The page should list the resident as enrolled in all of the available classes but the enrollment number is zero. Check out the images below:

image

Canvas Images to show enrollments and courses
image

image

Utilization does have an indicator that shows alerts the admin that this value is not available, but isn't clearly informing the user why.
image

frontend/src/pages/admin/ProviderPlatformManagement.tsx
gtonye: should be resolved with the [most recent commits](https://github.com/UnlockedLabs/UnlockEdv2/pull/1165/commits/2a780df839b974d21b74cf314788eb9581dea3e0)

frontend/src/pages/admin/ProviderPlatformDetail.tsx
gtonye: should be resolved with the [most recent commits](https://github.com/UnlockedLabs/UnlockEdv2/pull/1165/commits/2a780df839b974d21b74cf314788eb9581dea3e0)

frontend/src/pages/admin/ProviderPlatformDetail.tsx
gtonye: should be resolved with the [most recent commits](https://github.com/UnlockedLabs/UnlockEdv2/pull/1165/commits/2a780df839b974d21b74cf314788eb9581dea3e0)

frontend/src/pages/admin/ProviderPlatformManagement.tsx
gtonye: should be resolved with the [most recent commits](https://github.com/UnlockedLabs/UnlockEdv2/pull/1165/commits/2a780df839b974d21b74cf314788eb9581dea3e0)

gtonye:
@carddev81 

> @gtonye when adding a learning platform, I accidentally clicked off of the `Add Learning Platform` modal the modal closed unexpectedly. Can we make the modal stay open if the user clicks any area outside of the modal.


This should be resolved with the [most recent commits](https://github.com/UnlockedLabs/UnlockEdv2/pull/1165/commits/2a780df839b974d21b74cf314788eb9581dea3e0)


> Also, curious what the difference between the selection `Canvas Cloud` and `Canvas OSS`? If there are no underlying functional differences could we remove one of the selections for Canvas?

For the purpose of our integration there is none, the core APIs we are using (courses, users, schedule) are the same. As part of the recent changes  I've unified it under `Canvas`.

gtonye:
@carddev81

Note: We are also in the process of revamping the UI for the manage users screen, so the most recent changes don't include the fixes for some of the selection issues you noted.

There were a few issues in the logic used for the Canvas user linking. Some of the previous tests you performed may have caused irrecoverable stale state in your database, so I suggest trying to see if the user screen reflects the correct linking state — otherwise, starting with a newly created provider connection would be better.

The issues resolved:
- The logic had a silent failure if a resident was previously linked in the same provider
- Creating a resident failed because a Canvas username can have unsupported (in UnlockEd) username characters

This might explain some of the enrollment data discrepancies you observed in the program screen. Please let me know if you're still experiencing that.

**Note:** Canvas relies on the provider middleware — make sure to restart it when testing after pulling the new changes.

backend/migrations/00071_add_enrollment_types_to_provider_platforms.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
```
carddev81: @gtonye was there a specific reason that the version number of the scripts are starting at 00071? The next version number should be 00069?

carddev81:
@gtonye thanks so much for fixing the issues, I am able to map users now. We are getting down to the wire, I'm completing my functionality tests now and want to get my findings to you asap:

**What does not work:**
- Signed in as 'department admin' I should be able to map canvas users to residents located within any facility within the state but I am unable to. I should be able to map to residents within any facility
- Signed in as 'facility admin' I should ONLY be able to map canvas users to residents that are currently located at the facility of the facility administrator--This does not work as expected. There is no link to manage users? Is this the expected behavior to not be able to see the link to Learning Platforms, therefore unable to map users. See pic:

	
image


video:
https://github.com/user-attachments/assets/eecbb6a7-8c5b-4a78-8a02-761fc42527ed

carddev81:
@gtonye Class Schedule times do not line up with what is at the top of the class details page and also the programs listing page is displaying a different value, seems like each page displays a different time?:

**Canvas**
image

image

**Class Details Page**
image

**Class Details Page -- Schedule**
image

Programs Listing
image

Schedule
image

carddev81:
@gtonye I was able to take an action on canvas class, but should not be able to. see video below.


https://github.com/user-attachments/assets/5876c51d-877a-4993-b86d-6236f7264abe

backend/src/handlers/canvas_programs.go
```diff
@@ -0,0 +1,1393 @@
...
+	return result, nil
+}
+
+// fetchCanvasProviderProgram calls:
+//
+//	GET /api/v1/accounts/{accountID}/courses?per_page=100
+//
+// and builds the synthetic program entry.
+func (srv *Server) fetchCanvasProviderProgram(provider *models.ProviderPlatform) (models.ProgramsOverviewTable, error) {
+	coursesURL := provider.BaseUrl + "/api/v1/accounts/" + provider.AccountID + "/courses?per_page=100"
```
carddev81, line 307: @gtonye we had talked about the hard-coded 100 already, but just noting it here. It is concerning and haven't thought about any edge cases but if there are possibly more than 100 courses, we could run into problems.

carddev81:
@gtonye I logged in as facility admin and am unable to view classes within the canvas course. I get an error page popup:

https://github.com/user-attachments/assets/08690ea5-f07d-444b-b171-56c4a9d79a32

backend/src/handlers/classes_handler.go
```diff
@@ -107,6 +134,12 @@ func (srv *Server) handleIndexClassesForFacility(w http.ResponseWriter, r *http.
 	if err != nil {
 		return newDatabaseServiceError(err)
 	}
+	canvasClasses, err := srv.fetchCanvasClassesAllProviders(facilityID)
+	if err != nil {
+		logrus.WithError(err).Warn("failed to fetch canvas classes, returning DB classes only")
+	} else {
+		classes = append(classes, canvasClasses...)
```
carddev81, line 151: @gtonye should we update the metadata of the pagination to include the total? 

`args.Total += int64(len(canvasClasses))
`
Noticed that it is done within the programs_handler.go line 104

backend/src/handlers/user_handler.go
```diff
@@ -543,9 +543,83 @@ func (srv *Server) handleGetUserPrograms(w http.ResponseWriter, r *http.Request,
 		userPrograms[i].CalculateAttendancePercentage()
 		userPrograms[i].Schedule = models.FormatScheduleFromRRule(userPrograms[i].RecurrenceRule)
 	}
+	canvasPrograms := srv.fetchCanvasUserPrograms(userId)
+	userPrograms = append(userPrograms, canvasPrograms...)
```
carddev81, line 547: @gtonye should we update the meta here too. The queryCtx.Total for pagination?

carddev81 (changes requested):
@gtonye review complete, need a few changes per comments outside of this review. @CK-7vn should be sending his review shortly.

backend/src/handlers/user_matching.go
```diff
@@ -0,0 +1,251 @@
...
+	provider, err := srv.Db.GetProviderPlatformByID(int(service.ProviderPlatformID))
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+
+	var failed []string
+	applied := 0
+
+	for _, c := range req.Confirmed {
+		existing, _ := srv.Db.GetProviderUserMapping(int(c.UnlockEdUserID), int(service.ProviderPlatformID))
```
CK-7vn, line 186: This route doesn't have a resolver, so there's no guarantee the admin working on each resident is an admin of that facility. The initial admin check will just check that the admin is in fact an admin, not that they are a department admin. A fac admin in Facility A could Register a resident in Facility B and allow that resident to login to Facility A after the OIDC get's created. So just adding ```
claims := r.Context().Value(ClaimsKey).(*Claims)
	user, err := srv.Db.GetUserByID(c.UnlockEdUserID)
	if err != nil || (!claims.canSwitchFacility() && user.FacilityID != srv.getFacilityID(r)) {
		failed = append(failed, c.CanvasUser.Username)
		continue
	}```
Should do the trick.

backend/src/handlers/user_matching.go
```diff
@@ -0,0 +1,251 @@
...
+	// Score every canvas×unlocked pair
+	candidates := make([]matchCandidate, 0, len(canvasUsers)*len(unlockEdUsers))
+	for ci, cu := range canvasUsers {
+		canvasName := cu.NameFirst + " " + cu.NameLast
+		for ui, u := range unlockEdUsers {
+			score := nameSimilarity(canvasName, u.NameFirst+" "+u.NameLast)
+			candidates = append(candidates, matchCandidate{ci, ui, score})
+		}
+	}
+	sort.Slice(candidates, func(i, j int) bool {
```
CK-7vn: We should be using sort.SliceStable here, and then using a tie-breaker if possible when there may be an instance of 2 people with the same name.

backend/src/handlers/canvas_programs.go
```diff
@@ -0,0 +1,1393 @@
...
+		recurrenceRule = fmt.Sprintf("%s\nRRULE:FREQ=WEEKLY;BYDAY=%s;COUNT=1", dtstart, weekdayToRRuleDay(startAt.UTC().Weekday()))
+	}
+	return
+}
+
+// fetchCanvasCoursesScheduleEvents batch-fetches upcoming calendar events for the given
+// raw Canvas course IDs and returns a map of rawCourseID → synthetic ProgramClassEvent.
+// The synthetic event carries a RecurrenceRule and Duration suitable for passing to
+// services.FormatClassScheduleAndRoom, mirroring how regular class events work.
+func (srv *Server) fetchCanvasCoursesScheduleEvents(
```
CK-7vn, line 1132: The Canvas fetch loops are swallowing every failure as a break break and don't check the resp.StatusCode, fetchAllCanvasPages:136-139 handles it the way I think we can handle it here.

backend/src/handlers/canvas_programs.go
```diff
@@ -0,0 +1,1393 @@
...
+				Status:         status,
+			},
+			FacilityName: provider.Name,
+		})
+	}
+
+	// Determine which facility to scope enrollment counts to.
+	claims := r.Context().Value(ClaimsKey).(*Claims)
+	var facilityID uint
+	if facilityIDStr := r.URL.Query().Get("facility_id"); facilityIDStr != "" {
```
CK-7vn: Instead of reading the facility_id straight from the query. We should derive it from srv.getQueryContext(r).FacilityID

backend/src/handlers/oidc.go
```diff
@@ -51,6 +54,51 @@ func (srv *Server) handleGetOidcClient(w http.ResponseWriter, r *http.Request, l
...
+	log.add("oidc_client_id", id)
+	existing, err := srv.Db.GetOidcClientById(fmt.Sprintf("%d", id))
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	provider, err := srv.Db.GetProviderPlatformByID(int(existing.ProviderPlatformID))
+	if err != nil {
+		return newDatabaseServiceError(err)
+	}
+	// delete from Hydra
```
CK-7vn, line 71: I think creating the new client first then tearing down the old, and then verifying the delete status will help us avoid some potential edge cases in the future.

backend/src/handlers/canvas_programs.go
```diff
@@ -0,0 +1,1393 @@
...
+		if id, parseErr := strconv.Atoi(facilityIDStr); parseErr == nil && id > 0 {
+			facilityID = uint(id)
+		}
+	}
+	if facilityID == 0 && !claims.canSwitchFacility() {
+		facilityID = claims.FacilityID
+	}
+
+	// Concurrently count mapped enrollees for each course, scoped to facility when set.
+	counts := make(map[uint]int64, len(classes))
```
CK-7vn, line 929: Can we cache enrollee/completion results in the bucket added for the program summary and bind the concurrency using a sem similarly to how its done at :895

backend/src/handlers/user_matching_test.go
CK-7vn, line 1: All of our tests tend to live inside of tests/integration.

CK-7vn (commented):
Thanks for this, definitely a good integration, and a lot of work. I did a static analysis of the code, and haven't manually tested the findings I mentioned in my inlines, so if I missed anything that negates my assumptions, please don't hesitate to correct me. I also have evaluated some of coderabbits suggestions and thumbed up ones I would consider worth implementing. 

Most of what I saw cluster around two things.  
1. Identity handling is a super high stakes area given the domain, there's what I believe to be a cross-facility auth gap in the provider-user-mapping, a facility admin can bind a resident outside their facility, and then secondly the matcher decides two people are the "same" on string matching alone. Given from my experience it is not uncommon to have pairs of the same named individuals inside of a facility and or state. We should evaluate whether string matching alone is the best option here. 
2. Silent failure seems to be a recurring pattern which we try to avoid whenever possible once again, once again just because of the domain we're in. We would rather fail loudly and make sure the user is aware of what happened than the user proceeding on and acting based on an incorrect belief. 

Also, there is a lot of duplication inside of canvas_programs.go, and generally we really prefer to make any DB calls inside of the database layer through method calls vs inline when possible.

backend/src/handlers/canvas_programs.go
gtonye: This is resolved with the line below
```
courses, err := srv.fetchAllCanvasPages(provider, coursesURL, 0)
```
which retrieves all the pages, it paginates the calls by `100`

gtonye:
Thank you for the thorough review @CK-7vn and @carddev81.

The PR is ready for another review.

I would appreciate another functional review @carddev81, as the different roles (Department Admin, Facility Admin) should now have the correct access.

## Changelog

* Moved DB calls to a database layer
* Moved tests to the test folder
* Removed duplication in canvas_program.go
* Implemented selected CodeRabbit feedback
* Resolved issue with Department Admin not being able to select residents from any facility

carddev81:
@gtonye noticed that the mapping users page changed signficantly. Looks good. There is an issue with the Canvas User name not being displayed once it is linked to an UnlockEd resident. See picture below and notice the em dash:

The Canvas User names that should be displayed are Sam Student and Sara Student.

image

Before the changes there were no issues with mapping users, this one issue now exists.

carddev81:
@gtonye When clicking a class that has '1 Enrolled' resident, the **Roster** of that class displays 4 enrolled residents. This list should only display the one resident that is enrolled within that class (facility).  See video:



https://github.com/user-attachments/assets/be0658fb-6819-4dae-80f5-980422a099f1

carddev81:
@gtonye inconsistent time issues still exists between the pages.

**Statewide Program View:**
image
>

**Class Details Card**
image

**Class Details > Schedule shows 9:02 AM**
image

**Schedule (Admin) > 6:02 AM**
image

carddev81:
@gtonye I do not remember if we had discussed whether or not the canvas program should be a part of the filtering. I noticed that my Canvas Program is not a selection within the program filter drop down box. See below, I labeled my canvas instance 'Missouri College Canvas Program'

image

carddev81:
@gtonye User mapping issue edge case found:

I clicked Create button within the Unmatched section and first I had assumed that I would see a create resident modal w/form to fill out new resident information--this did not happen and immediately the mapping displayed as 'Pending apply' within the 'Linked residents' table. I clicked apply and it proceeded with the creation. I then clicked Unlink and got the 'Pending apply' back? I am unable to undo the create functionality that I started? See video below:

We had discussed on May 29th @1PM Central (Present at this meeting was  Carolina, you, and Guillaume) that we would see a create resident modal popup and we would be required to fill out a form to add the resident to the application.

https://github.com/user-attachments/assets/fb7f0095-377f-4977-b2e8-088c15bbb2fd

Another odd issue related to the one above is that the resident that was created is still available to be selected even though it is displayed within the list of available selections. See video:

https://github.com/user-attachments/assets/b0c18b7d-feb8-43ec-a72b-cc44e51fed1b

carddev81:
@gtonye expected date of enrollment to be a real date versus 0001-01-01. See pic below:

image

gtonye:
It is ready for a re-review.

Below is the status of all the items:

1. Inconsistent time issues still exist between pages. ✅
2. Class roster shows 4 residents when only 1 is enrolled. (dept admin) ✅
3. Canvas username shows an em dash instead of the name once linked (should be Sam Student / Sara Student). ✅
4. Canvas program ('Missouri College Canvas Program') doesn't appear in the program filter dropdown — open question on whether it should. ⚠️
5. Create in the Unmatched section skips the expected resident-info modal and can't be undone — differs from what we discussed on May 29. Created residents also remain selectable. ✅
6. Enrollment date shows 0001-01-01 instead of a real date. ✅

On

> Canvas program ('Missouri College Canvas Program') doesn't appear in the program filter dropdown — open question on whether it should.

I was unable to repro — I do see my instance in the dropdown (I was signed in as a department admin for my test). Can you try again with the latest changes?

canvas-integration-classes-filter


## Change Log

* Updated user mapping page to allow single user linking or creation for all residents
* Added dynamic Canvas user information retrieval when reviewing linked users
* Fixed issue with Canvas retrieved classes not using the class timezone on display
* Added support for dynamic facility / course pairs

carddev81:
@gtonye a part of listed item 5 is a reoccurring issue

`Create in the Unmatched section skips the expected resident-info modal and can't be undone — differs from what we discussed on May 29. Created residents also remain selectable.` ✅

1. I am having issues with the `Linked residents` section of the `Learning Platform Users` screen.  Note that in the pic below is a resident (StudentSteve) listed within the Linked residents section. From the `Needs review` section when I click `Link existing` next to one of the residents here, `StudentSteve` is available to choose from in this list. Is this intended behavior?

image

image

2. Unable to cancel or clear out the 'Confirm' selections I made from the `Needs review` section. I only have the option of Apply # match(es). See video below.

https://github.com/user-attachments/assets/47bdc0f7-c336-4751-84b9-a469452d51c1

3. Noticed that all available canvas users (teachers, TAs, students, ect.) are being displayed rather than just students. Knowing this I revisited the SQL migration for this PR. Makes me wonder if the column named enrollment_types within the table public.provider_platforms is used or needed? In a previous PR this value was needed, but now it seems like this column value is not being utilized or accessed and the value is currently hardcoded as 'student'. The enrollment_type when querying Canvas API is returning all types of users even though we request just students. See pic below:

image

**From CANVAS**
image

If you feel that this column will not be needed in the future then we should be able to remove the following migration script along with any related code.

00071_add_enrollment_types_to_provider_platforms.sql

4. While running tests against item `Class roster shows 4 residents when only 1 is enrolled. (dept admin) ✅` I stumbled across data inconsistencies related to Total Enrollments across programs and Classes. I created a video to display the inconsistencies. This occurs when I `Unlink` and then `Link existing` residents. The old records will be used in the counting of enrollments and causes false positives. I will post the video shortly. In the video you should be able to see that the enrollment count is off, it should have stayed the same. The deleted (Unlinked) users are being counted along with the new unlinked users.

https://github.com/user-attachments/assets/dc5e4be0-ce86-4fee-b7f4-4377339603e8

@gtonye The number within the Enrollment column on the Programs listing page does not match the number of enrollments within the metrics card on the `Statewide Program View`. Also, to be consistent the classes column value on the Programs listing page does not match the metrics card on the `Statewide Program View`.  These should match exactly? see pics below for example, but you should be able to see the issues I was running into in the video above.

image

image

Also, for facility admins, numbers tend to be off for number of enrollments, please see facility admin video below:

https://github.com/user-attachments/assets/2f03ac4b-fdbd-48b8-a5f5-117e007b5104

5. Note: the classes page is loading really slow when I have more than 10 facilities in the system? This can be seen in the 1st video in number 4 above.

gtonye:
Thank you @carddev81 for taking the time to review in detail.

The recent changes should address all the elements in the feedback.

**1. A resident already in Linked residents (StudentSteve) is still selectable under 'Link existing' from Needs review — intended?**

> Residents awaiting confirmation of their match are now filtered out of the available residents for mapping.

**2. Can't cancel/clear 'Confirm' selections in Needs review — only option is 'Apply # match(es)'.**

> Added a new cancel option for confirmed matches.

**3. All Canvas user types are showing instead of just students. Looks like the `enrollment_types` column on `provider_platforms` is now hardcoded to 'student' and unused — if it's no longer needed, we could drop migration `00071_add_enrollment_types_to_provider_platforms.sql` and related code? What is your recommendation?**

> It was indeed no longer used, so I removed it.

**4. Enrollment counts are inconsistent: Unlink + Link existing keeps counting deleted records (false positives), the Programs listing Enrollment/Classes columns don't match the Statewide Program View metric cards, and facility admin numbers are off (fac admin numbers off could be because of false positives).**

> The program overview page was displaying the number of residents enrolled in the program. This now represents the total number of enrollments across all classes. There was also a missing facility ID scope, which explains the additional discrepancy for facility admins.

**5. New Classes page loads slowly with more than 10 facilities (see first video in #4).**

> This one is harder to fix. Canvas has no concept of a "Facility" — that grouping exists only in UnlockEd. To compute per-facility enrollment counts, the code must fetch every course from Canvas, fetch every active enrollment per course, look up which of those Canvas users are mapped to local residents, and then group those residents by facility. With N courses, this means N Canvas API calls on every page load, with no shortcut available — Canvas provides no bulk enrollment endpoint and no way to query enrollments by resident or group.
>
> ```mermaid
> flowchart TD
>     A([Page load request]) --> B
>
>     B["<b>GET</b> /api/v1/accounts/{id}/courses\nOne call — returns all courses as a list"]
>
>     B --> C
>
>     subgraph loop ["Per-course loop — repeats N times (once per course)"]
>         C["<b>GET</b> /courses/{id}/enrollments\nReturns Canvas user IDs of active students"]
>         C --> D["<b>DB:</b> provider_user_mappings\nMatch Canvas IDs → local residents"]
>     end
>
>     D --> E["Group residents by facility\n(facility concept exists only in UnlockEd)"]
>     E --> F([Per-facility enrollment counts\nOne row per facility × course])
>
>     style B fill:#CECBF6,stroke:#534AB7,color:#26215C
>     style C fill:#F5C4B3,stroke:#993C1D,color:#4A1B0C
>     style D fill:#9FE1CB,stroke:#0F6E56,color:#04342C
>     style E fill:#9FE1CB,stroke:#0F6E56,color:#04342C
>     style loop fill:#F1EFE8,stroke:#888780,color:#2C2C2A
> ```
>
> The underlying limitation is that the Canvas API does not provide a single endpoint to retrieve enrollments across all courses, nor a way to query which courses a given resident is enrolled in. As a result, the page performance scales linearly with the number of courses rather than the number of facilities.
>
> To mitigate this, we've added a cache layer so the computation only runs on the first load and is served from cache on subsequent requests. The tradeoff is that enrollment data may be slightly delayed when changes are made in Canvas — the cache will reflect the previous state until it expires.

carddev81:
@gtonye 

1. A few issues still exist as part of listed item 4 from last review...counts are still not reflective of the actual number of enrollments. See pics below, the total number of enrollments should match the number of currently enrolled. The total enrollments includes historical enrollments if any. Also notice the negative number on the dept admin view. See video, you should be able to spot that the deleted records are being counted as enrollments when signed in as dept. admin. I also screenshot the result set of the provider_user_mappings table to display the deleted_at column.

deptadmincounts

EnrollmentCountIssue

provider_user_mapping

https://github.com/user-attachments/assets/941b890f-1c40-43df-9b6d-becf9f062427

2. Recieving error on initial load of classes page for facility/dept. admins. After mapping residents I immediately went to the Programs > Classes page. See video.

https://github.com/user-attachments/assets/6119ec6c-e962-455c-a493-ad085b12c4c3

3. Didn't catch this on the last reviews, but Import All Users is giving a false positive, when clicked it displays the message 'Users imported successfully. Please check for accounts not created.' No users were actually imported and the underlying API call fails.

import_all

4. Teachers and TAs are still being pulled from the canvas API, is there no way to filter these out?. I noticed in the code the call to attempt to pull only students but this does not seem to work.

teacherstas

Canvasstudents

gtonye:
@carddev81

This should be ready for re-review. The recent changes include fixes for:
* **Inconsistencies in enrollment calculation:** The Programs list page, program details page, and class page are now consistent.
* **Enrollment calculation excludes deleted provider user mappings:** The calculation no longer includes deleted provider user mappings.
* **Error on page load for facility and department admins:** After linking a user, the program list page would often throw an error — this has now been resolved.
* **"Import All Users" has been removed:** This feature had become redundant and has been removed.

**Note:** As discussed offline, there is an inconsistency across Canvas instances where certain API versions appear to handle the `enrollment_type` parameter differently. This would need further investigation, potentially by looking at open issues on the Canvas LMS repository to see if it has been flagged.

carddev81:
@gtonye I was able to duplicate error page consistently. The error occurs the very first time the cache is loaded from the programs page. I captured the logging output for you as well as the devtools. Check out video below:

Here are the steps to reproduce on your machine

1. Unlink all users
2. Log out of unlocked
3. Stop all docker containers or processes
4. Start server back up
5. Link users
6. Go to programs page and click a program
7. See error.

https://github.com/user-attachments/assets/8a7051f4-f9dd-4fa7-a540-f89a9bc79940



```
INFO[0189] /app/backend/src/database/provider_user_mappings.go:20 record not found
[0.249ms] [rows:0] SELECT * FROM "provider_user_mappings" WHERE (user_id = 208 AND provider_platform_id = 8) AND "provider_user_mappings"."deleted_at" IS NULL ORDER BY "provider_user_mappings"."id" LIMIT 1 
INFO[0189] /app/backend/src/database/provider_user_mappings.go:20 record not found
[0.239ms] [rows:0] SELECT * FROM "provider_user_mappings" WHERE (user_id = 209 AND provider_platform_id = 8) AND "provider_user_mappings"."deleted_at" IS NULL ORDER BY "provider_user_mappings"."id" LIMIT 1 
INFO[0189] /app/backend/src/database/provider_user_mappings.go:20 record not found
[0.154ms] [rows:0] SELECT * FROM "provider_user_mappings" WHERE (user_id = 210 AND provider_platform_id = 8) AND "provider_user_mappings"."deleted_at" IS NULL ORDER BY "provider_user_mappings"."id" LIMIT 1 
INFO[0189] /app/backend/src/database/provider_user_mappings.go:20 record not found
[0.224ms] [rows:0] SELECT * FROM "provider_user_mappings" WHERE (user_id = 207 AND provider_platform_id = 8) AND "provider_user_mappings"."deleted_at" IS NULL ORDER BY "provider_user_mappings"."id" LIMIT 1 
{"admin_id":33,"audit":"admin_action","facility_id":1,"facility_name":"Palmer Correctional Center","handler":"handleApplyMatches","ip_address":"172.18.0.14:35446","level":"info","method":"POST","msg":"","path":"/api/actions/provider-platforms/8/apply-matches","role":"department_admin","session_id":"0a660bd4-60a6-4a04-82c4-ad2a064c2d1a","time":"2026-06-18T15:03:31Z","username":"rich.salas"}
{"level":"info","msg":"Init request for provider service","time":"2026-06-18T15:03:31Z"}
{"level":"info","msg":"request: \u0026{GET http://provider-service:8081/api/users/all?id=8 HTTP/1.1 1 1 map[] \u003cnil\u003e \u003cnil\u003e 0 [] false provider-service:8081 map[] map[] \u003cnil\u003e map[]   \u003cnil\u003e \u003cnil\u003e \u003cnil\u003e  {{}} \u003cnil\u003e [] map[]}","time":"2026-06-18T15:03:31Z"}
{"level":"info","msg":"url: http://provider-service:8081/api/users/all?id=8 \n","time":"2026-06-18T15:03:31Z"}
{"level":"info","msg":"Init request for provider service","time":"2026-06-18T15:03:31Z"}
{"level":"info","msg":"url: http://provider-service:8081/api/users?id=8 \n","time":"2026-06-18T15:03:31Z"}
{"level":"info","msg":"request: \u0026{GET http://provider-service:8081/api/users?id=8 HTTP/1.1 1 1 map[] \u003cnil\u003e \u003cnil\u003e 0 [] false provider-service:8081 map[] map[] \u003cnil\u003e map[]   \u003cnil\u003e \u003cnil\u003e \u003cnil\u003e  {{}} \u003cnil\u003e [] map[]}","time":"2026-06-18T15:03:31Z"}
{"level":"debug","msg":"users received from middleware: [{StudentSidney Sidney Student student5@example.com 9 student5@example.com} {StudentSimon Simon Student student6@example.com 10 student6@example.com} {TATina Tina TA ta@example.com 4 ta@example.com} {TeacherTerry Terry Teacher teacher@example.com 2 teacher@example.com} {TeacherTom Tom Teacher teacher2@example.com 3 teacher2@example.com}]","time":"2026-06-18T15:03:42Z"}

```

gtonye:
@carddev81

I'm trying another approach. I suspect the Canvas API is slow, especially on that initial call.

Can you test again with the most recent change and let me know if it's still causing the error?

## Technical Notes

We've moved all program computation inside a goroutine triggered by a cache miss. This is due to Canvas being slow and program computation requiring iteration over all courses to calculate enrollment counts.
We've added a visual signal on the frontend to indicate to the user that the particular program is loading asynchronously.

https://github.com/user-attachments/assets/4649a2b7-ed21-479d-b5b2-362811d2064b

carddev81 (approved):
Looks and works great.

PR #1169, opened 2026-06-11
Feat: Learning Record For Residents

gtonye:

- [ ] No debug/console/fmt.Println statements
- [ ] Unnecessary development comments removed
- [ ] All acceptance criteria verified
- [ ] Functions according to **ticket specifications**
- [ ] Tested manually where applicable

## Description of the change

This PR adds backend persistence for the Learning Record feature allowing residents to log

**Backend**
- DB layer: get / create / update / delete learning record entries
- HTTP handlers for 7 routes under `/api/learning-record/`
- Feature flag for learning record

**Frontend**
- Homepage widget to prompt and promote the functionalityto resident
- Page to manage achievements

## How to test

1. As a **superadmin**, go to Feature Control and enable the Learning Record feature
2. As a **resident**, verify that the Learning Record is visible
3. Create a new achievement and complete all steps through to submission — verify the entry persists after a page reload
4. **Draft edge case** — start filling in an achievement but do not submit; reload the page and verify the draft is restored and flagged as incomplete

## Screenshot(s)

a-learning-record-feature-flag
b-learning-record-resident-home

### Technical design decision — column-per-question schema

The questionnaire is stored in a dedicated table with each question as its own column, rather than a single JSONB column containing the full response payload. This prioritizes maintainability through migrations:

- Adding a new question means adding a column, which is explicit, trackable, and straightforward to roll back
- JSONB would require application-level logic to manage schema evolution inside the blob and coordinate data migrations accordingly
- Column-level constraints, indexing, and query filtering are all simpler with discrete columns


backend/migrations/00072_create_learning_record_tables.sql
```diff
@@ -0,0 +1,39 @@
+-- +goose Up
+CREATE TABLE learning_record_entries (
```
carddev81: @gtonye noting that the table should be qualified with the schema name

`public.learning_record_entries` 

also where appropriate please make the change.

carddev81: @gtonye this table has `created_at` and `updated_at`, but it's missing the rest of our standard auditing set:

- `deleted_at TIMESTAMPTZ`
- `create_user_id INTEGER` — tracks who created the record
- `update_user_id INTEGER` — tracks who last modified it

Can we add these to stay consistent with the rest of our tables (e.g. `program_classes`)? 

Also please make sure that when saving or updating records the values are being set/saved correctly

backend/migrations/00069_create_learning_record_tables.sql
```diff
@@ -0,0 +1,39 @@
+-- +goose Up
```
carddev81, line 1: Was there a specific reason that the version number of the scripts are starting at 00072? The next version number should be 00069?

carddev81: if we can, please use the next available number for consistency

carddev81:
@gtonye pointing out an inconsistency with the grey background. It doesn't span the full viewport like all other pages. See pic below. Click on any other link within the nav bar.

image

backend/migrations/00073_add_learning_record_feature_flag.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
```
carddev81: @gtonye Not sure why we need this extra migration file, we usually try to use just one migration file per PR. Sometimes we run into a transaction issue where 2 is required. For consistency within our codebase can this bit of code be put into 00072 file.

config/dev.nginx.conf
carddev81, line 1: please revert this file back to its original form, this causes local testing to fail

carddev81 (changes requested):
Tested good and found a few minor issues.

frontend/src/components/dashboard/ContinueLearningSection.tsx
```diff
@@ -0,0 +1,217 @@
...
+        id: 'preview-interview-skills',
+        title: 'Interview Skills',
+        statusLabel: 'Not started yet',
+        contentType: 'library',
+        contentTypeLabel: 'Article',
+        durationLabel: '8 min',
+        progressPercent: 0,
+        href: '/knowledge-center'
+    }
+];
```
calisio: there are things hard coded into the homepage - these should be accurate to actual data not just hardcoded into the codebase?

calisio (commented):
things i'd like to see:
- auto scroll to top on left side of screen when i click into the next section of questions (right now it stays at the bottom & it gets confusing) - see video
https://github.com/user-attachments/assets/819960ad-ba10-426a-b20d-fe89a2578e47

- auto scroll to where it is adding content for the right side of the screen - you add stuff in the prompts and then you dont see it it populate in the right side because of the scroll issue

https://github.com/user-attachments/assets/81176f53-9c7c-4c47-b588-a643b9fde14e

- on home screen, dismiss only temporarily dismisses that record

frontend/src/components/dashboard/ContinueLearningSection.tsx
gtonye: I checked with the team and will remove the ContinueLearningSection as it was part of a concept for the home page but is not related to the digital resident transcript.

gtonye:
> things i'd like to see:
> 
> * auto scroll to top on left side of screen when i click into the next section of questions (right now it stays at the bottom & it gets confusing) - see video
>   https://github.com/user-attachments/assets/819960ad-ba10-426a-b20d-fe89a2578e47
> * auto scroll to where it is adding content for the right side of the screen - you add stuff in the prompts and then you dont see it it populate in the right side because of the scroll issue
> 
>  Screen.Recording.2026-06-16.at.1.31.15.PM.mov 
> * on home screen, dismiss only temporarily dismisses that record

Thank you for the review @calisio 

We've made the changes requested. Can you help us better understand this one

> on home screen, dismiss only temporarily dismisses that record

Otherwise here is a demo of the recent interaction changes:

https://github.com/user-attachments/assets/7cf6b139-8313-455a-9254-5146ac2d6db4

## Change Log

* Connected the sections of the preview view with the questions for the resident.
* Removed the unused `ContinueLearning` concept component

calisio:
@gtonye 

https://github.com/user-attachments/assets/3d44ea09-7188-4811-8a19-cce79f385e4a

this is what i was referring to in the dismiss - it only temporarily removes it, but then re-appears when you go back to the homepage. I am okay with this functionality for now given the limited time on the project, so feel free to ignore that comment.

The only super small thing left is that for some reason now the page overflows to the bottom on the home page. 

Screenshot 2026-06-18 at 10 41 37 AM

I also believe that this section is hard coded into the codebase & therefore does not lead to any libraries actually in our tool. We can add this ourselves but since it does not have to do with digital resident transcript could you also remove this component (and/or just remove the hard-coded nature of it?)

Screenshot 2026-06-18 at 10 45 20 AM

The rest looks good

gtonye:
@calisio Please take another look and let us know if the issues are resolved with the most recent changes.

calisio (approved):
looks good on the frontend

carddev81 (approved):

PR #1178, opened 2026-06-16
fix: double booking instructor err

CK-7vn:

- [x] Functions according to **ticket spexifications**

## Description of the change
There was an issue with double-booking instructors, and the eligibility of instructors during assignment. This PR implements a double-booking detection system similar to double-booking rooms. 
- New database/instructor_conflicts.go: expands an instructor's events/overrides into time slots (getInstructorBookingsInRange) and
  detects overlaps against a candidate schedule, honoring per-occurrence instructor overrides, cancellations, and facility timezone.
  - Refactored rooms.go to share the recurrence-window + overlap + class-name logic (prepareConflictWindow,
  buildConflictsFromBookings) between room and instructor checks (DRY). LockAndCheckConflicts now locks and checks room and/or
  instructor; conflicts are tagged with ConflictType.
  - Wired into handleCreateClass, handleUpdateClass (the ticket's "change a class's instructor" path), and handleCreateEvent; returns
  HTTP 409 with a type-aware message (instructor vs room). Added instructor-belongs-to-facility validation on the event path.
  - Frontend: EditClassModal shows an instructor-aware conflict heading; RoomConflict type gains conflict_type.
  - Tests: new instructor_conflict_test.go covering overlap detection, non-overlap, and the 409-on-instructor-change scenario.
 
  - Related issues: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1214430813078163?focus=true

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ jar
 /csvs/zims/*.zim
 /.claude
 .worktrees/
+/frontend/node_modules/*
--- a/backend/src/database/instructor_conflicts.go
+++ b/backend/src/database/instructor_conflicts.go
@@ -1,0 +1,133 @@
+package database
+
+import (
+	"UnlockEdv2/src/models"
+	"fmt"
+	"sort"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/teambition/rrule-go"
+)
+
+func checkInstructorRRuleConflicts(db *DB, req *models.ConflictCheckRequest) ([]models.RoomConflict, error) {
+	w, err := db.prepareConflictWindow(req)
+	if err != nil {
+		return nil, err
+	}
+	bookings, err := db.getInstructorBookingsInRange(req.FacilityID, req.InstructorID, w.rangeStart, w.rangeEnd, w.facilityTZ)
+	if err != nil {
+		return nil, err
+	}
+	return buildConflictsFromBookings(db, w, bookings, req, models.ConflictTypeInstructor), nil
+}
+
+func (db *DB) getInstructorBookingsInRange(facilityID, instructorID uint, rangeStart, rangeEnd time.Time, facilityTimezone string) ([]models.RoomBooking, error) {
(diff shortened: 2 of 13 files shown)


corypride (changes requested):
Core feature tests and works as expected. Small fix and merge conflicts, and this should be good.

**One gap worth fixing before merge:** the create-class conflict dialog (inline "Create New Class" form on the Program Details page) still hardcodes "Room Scheduling Conflict / The selected room has N conflicts" and ignores `conflict_type`. So creating a class with a double-booked instructor in a different room correctly returns a 409 but shows misleading "room" wording. This PR updated `EditClassModal` but not this separate create-class component — it should read `conflict_type` the same way.

**Pre-existing (not this PR):** `EditClassModal`'s duration parse (`/(\d+)h(\d+)m/`) doesn't handle a `"2h"`-form duration, leaving End Time blank so a save writes `duration:"0h0m0s"`; a zero-length event then silently bypasses conflict checks. Not triggered by app-created events (which store `"2h0m0s"`), but a defensive fix would harden the feature.

**Why that matters for this PR specifically:** the whole point of the double-booking feature is to notice when two classes overlap in time. A class with zero length occupies no time at all, so it can never "overlap" anything. Once a class's duration has been zeroed out this way, both the room-conflict and instructor-conflict checks quietly pass it — you could double-book an instructor and the system wouldn't complain, because as far as it's concerned the class takes zero time.

**Why this isn't an everyday production problem:** classes created through the app always save their duration in the full `2h0m0s` format, which the code reads correctly. The shorthand `2h` form only showed up because that's how I wrote it in the test seed data. So under normal use this stays hidden — but it's a fragile spot. If a duration ever ends up in shorthand (bad import, future code change, a migration, hand-edited data), editing that class would silently zero its length and poke a hole in the conflict checks. A small hardening fix — teach the parser to accept the shorthand form, and/or refuse to save a zero-length class — would close that gap.

corypride (approved):
GTG!

PR #1181, opened 2026-06-23
 PostHog analytics instrumentation for pilot tracking

corypride:
# feat: PostHog analytics instrumentation for pilot tracking

## Summary

Instruments the frontend with PostHog to capture the metrics needed for the facility pilot
(Asana: *"To Dos before pilots: Set up posthog graphs"*). Adds a small typed analytics layer,
custom events across the attendance / program / class / enrollment flows, user identification,
session reset on logout, and error tracking. No backend schema or API changes.

User identification is **PII-minimized**: we send **only** `id`, `role`, and `facility_id` as
person properties — no names, usernames, or emails. Event properties (`class_id`, `event_id`,
`present_count`, `duration_seconds`, etc.) describe objects and aggregate counts, never individuals.

## How this maps to the Asana graph requirements

| Asana graph | Status | PostHog insight | What powers it |
| --- | --- | --- | --- |
| **Time to complete an attendance session** | ✅ | "Attendance time — avg & median (sec)" (Trends) | `attendance_session_started` → `attendance_session_completed` with `duration_seconds` |
| **Time to set up a new program or class** | ✅ | "Program setup time" + "Class setup time" (Trends) | `program_creation_started/_completed`, `class_creation_started/_completed` with `duration_seconds` |
| **Time of day attendance recorded vs scheduled** *(OPTIONAL)* | ⚪️ Deferred | "Attendance recording delay" (planned; not built) | Optional in the ticket; intentionally not shipped for the pilot |
| **Login frequency per user** | ✅ | "Stickiness (days active/week)" (Stickiness) | Default pageviews + `posthog.identify()` (stable per-user id) |
| **Drop-off points in key flows** | ✅ | "Attendance funnel" + "Enrollment funnel" (Funnels) | Funnel-ready events — attendance (`started`→`completed`) and enrollment (`enroll_modal_opened`→`enroll_residents_selected`→`enroll_completed`) |
| **Error states hit during common flows** | ✅ | "API errors by endpoint" (Trends/Table) + built-in Error tracking | `api_error` event + `capture_exceptions: true` |
| **Distinct active users per week** | ✅ | "Weekly active users" (Trends) + "Active users by facility" (Table) | Default identify + pageviews |
| **New users added over time** | ✅ | "User lifecycle" (Lifecycle) | Default identify |

## Events shipped

| Event | Properties | Fires from |
| --- | --- | --- |
| `attendance_session_started` | `class_id`, `event_id`, `date` | `event-attendance/index.tsx` (on open) |
[...]

PR #1182, opened 2026-06-24
Programs listing page is set to 20 per page but only displays 10

corypride:

## Description of the change

Fixes the Programs listing page showing only 10 items per page while the "items per page"
control read 20.

**Root cause:** `ProgramsPage` initialized `perPage = 10`, but the shared `Pagination` dropdown
only offers 20 / 40 / 80. `<select value={10}>` matched no option, so the browser displayed the
first option (20) while the list actually sliced 10 rows.

Changes:
- **Frontend:** `ProgramsPage` now defaults to 20 per page (`useUrlPagination(1, 20)`), so the
  dropdown and the rendered slice agree.
- **Backend:** `getPaginationInfo` now defaults `per_page` to 20 (was 10) for both the
  empty-param and parse-failure paths, aligning the API default with the intended page size.
- Audited every other paginated page — they already default to 20 (via the hook default or an
  explicit `20`), so no other page required a change.

Also in this PR:
- **Refresh persistence fix:** while testing, found the Programs page reset to page 1 on hard
  refresh / deep-link. Cause was a first-render `useRef` guard that React StrictMode defeats by
  double-invoking the effect on mount. Replaced it with previous-value tracking so the page resets
  only when a filter/sort/search value actually changes.
- **`go mod tidy` (chore):** `adrg/strutil` is a direct import in `user_matching.go` (Canvas name
  matching) but was still flagged `// indirect` since the Canvas PR; reclassified as direct. No
  dependency change (`go.sum` untouched).
- **Integration-test compile fix:** `tests/integration/testenv.go` had a `:=` redeclaration that
  fails to compile on `main`; fixed here so CI can run (same fix as open PR #1181).

- **Related issues**: Closes [Asana ID-717](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1215770512119836?focus=true).

## Screenshot(s)

Programs page now renders 20 per page by default with the dropdown matching, page 2 shows the
remainder, and the page survives a hard refresh:

- `
test_1_2_page2_remainder
(Programs is admin-facing, so the 1366 x 768 resident-facing requirement does not apply.)

## Additional context

- **What to focus on:** the backend `per_page` default change applies to all paginated endpoints,
  not just programs — reviewers may want to confirm no consumer depended on the old default of 10.
  (The Programs page itself fetches `per_page=100` and paginates client-side, so its own display is
  driven by the frontend default.)
- **`detailed-list` returns 21 for a 20-page:** expected — the handler appends in-progress Canvas
  provider programs after the paginated DB query and bumps `total` accordingly. Not a pagination bug.
- **Known issue not addressed:** `main`'s integration-test package does not compile without the
  `testenv.go` fix; this PR carries the same fix as PR #1181, so whichever merges second will
  collapse the identical hunks.
- **Decision considered and declined:** centralizing the default page size into a shared constant —
  out of scope for this ticket; left the existing per-page hook defaults in place.
- No core features or components were removed.

---
  - https://app.asana.com/0/0/1215310633886744

### DIFF
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -26,10 +26,12 @@ require (
 	gorm.io/gorm v1.31.1
 )
 
-require github.com/xuri/excelize/v2 v2.10.0
+require (
+	github.com/adrg/strutil v0.3.1
+	github.com/xuri/excelize/v2 v2.10.0
+)
 
 require (
-	github.com/adrg/strutil v0.3.1 // indirect
 	github.com/aws/aws-sdk-go-v2/service/signin v1.0.5 // indirect
 	github.com/richardlehane/mscfb v1.0.6 // indirect
 	github.com/richardlehane/msoleps v1.0.6 // indirect
--- a/backend/src/handlers/canvas_programs.go
+++ b/backend/src/handlers/canvas_programs.go
@@ -393,16 +393,16 @@ func (srv *Server) fetchAllCanvasPages(ctx context.Context, provider *models.Pro
 			return nil, err
 		}
 		if resp.StatusCode != http.StatusOK {
-			resp.Body.Close()
+			_ = resp.Body.Close()
 			return nil, fmt.Errorf("canvas API returned %d", resp.StatusCode)
 		}
 		var page []map[string]interface{}
 		if err := json.NewDecoder(resp.Body).Decode(&page); err != nil {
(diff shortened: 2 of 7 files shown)


CK-7vn (approved):
Nice work man.

PR #1187, opened 2026-07-07
Ck 7vn/fac kc

CK-7vn:

## Description of the change
Adds per-facility visibility to the knowledge center. 

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1216014442850259

## Screenshot(s)
image
image

## Additional context

Each content article is now a single row, so that being said, it may be worth thinking about/working over if we still want facility admins to be able to delete/edit videos and helpful links.


backend/src/database/facility_visibility.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, newGetRecordsDBError(err, "facility_visibility_statuses")
+	}
+	return visibilities, nil
+}
+
+func (db *DB) setContentVisibilityForFacilities(args *models.QueryContext, contentID, providerID uint, facilityIDs []uint, visible bool) error {
+	if len(facilityIDs) == 0 {
+		return nil
+	}
+	statuses := make([]models.FacilityVisibilityStatus, 0, len(facilityIDs))
```
carddev81: lines 35 to 48 seems to be some business logic introduced into to the database layer here, is there anyway we can move these lines to the business layer, maybe even creating a service layer similar to ClassesService (/services/classes.go)--this service logic can be reused in service layer

backend/src/database/helpful_links.go
```diff
@@ -8,8 +8,9 @@ import (
 
 type HelpfulLinkResp struct {
 	models.HelpfulLink
-	IsFavorited bool `json:"is_favorited"`
-	IsFeatured  bool `json:"is_featured"`
+	IsFavorited          bool `json:"is_favorited"`
```
carddev81: @CK-7vn no need to apply any changes here but noticed that each openconent type has the same three fields, isfavorited, isfeatured, and visibilityfacilitycount....i wonder if using embedding would be good here or just keep it as is? Just a thought really. Want to be consistent either way.

carddev81 (approved):
@CK-7vn Left a couple comments. Everything works great and looks good.

backend/src/database/helpful_links.go
CK-7vn: No definitely, I think embedding is the right choice!

PR #1190, opened 2026-07-10
Feature flagging at a per-facililty level

corypride:
## Pre-Submission PR Checklist

- [x] No debug/console/fmt.Println statements
- [x] Unnecessary development comments removed
- [x] All acceptance criteria verified
- [x] Functions according to **ticket specifications**
- [x] Tested manually where applicable
- [x] Branch rebased with latest main
- [ ] No business logic exists within the database layer

## Description of the change

Covers **two tickets**: the per-facility feature-flag feature (ID-719) and the follow-up bug report that the flags weren't fully working (ID-740).

### ID-719 — Per-facility feature flags

Adds **per-facility feature flags** so a department/system admin can turn features on/off for an individual facility, layered on top of the existing statewide (global) flags. Effective access = `global_enabled(feature) AND NOT facility_disabled(feature)`; a facility with no override row inherits the global value (rollout is a no-op — no seeding, no behavior change).

- **Migration `00071`** — new `facility_feature_flags` table (`UNIQUE(facility_id, feature)`, FK to `facilities` ON DELETE CASCADE, `enabled` default TRUE, audit columns). Reuses the existing `feature` enum, so no `ALTER TYPE`. Production starts empty = inherit. *(Renumbered from `00070` to avoid collision with a migration that landed on main.)*
- **Model** (`feature_flags.go`) — `FacilityFeatureFlag` override row + `TopLevelFacilityFeatures` and the overview/detail DTOs. No `default` gorm tag on `Enabled` (so a `false` insert isn't dropped).
- **Database layer** (`facility_feature_access.go`) — effective-access resolver (`GetFacilityFeatureAccess`), overview list with the on/off filter (`GetAllFacilitiesFeatureStatus`), right-panel detail (`GetFacilityFeatureDetail`), single-feature upsert with layering guards (`ToggleFacilityFeature`), and the bulk `ApplyFacilityFeaturesToAll`.
- **Enforcement rescope** (`auth.go`, `middleware.go`, `server.go`, `user_handler.go`) — makes `Claims.FeatureAccess` the single source of truth. Resolved once at login: cross-facility admins (system/dept) get the global set; pinned users (facility_admin/students) get their facility's effective set. Fail-open to the global set on a DB error.
[...]

PR #1191, opened 2026-07-13
Audit History

corypride:

## Description of the change

This PR bundles two independent changes plus a required security dependency bump: a frontend
pagination tweak and a backend attendance soft-delete correctness fix (the latter spanning the
attendance feature and all attendance analytics/report reads).

### 1. Audit History pagination → 10 per page

Sets the **Audit History** tab to paginate at **10 entries per page** (previously 20), mirroring the
per-page work done for the Programs list (Asana ID-717). Applies the same lesson from that fix: the
default per-page value must be a selectable option in the dropdown, or the view renders a page size
the control can't represent.

- **`frontend/src/components/Pagination.tsx`** — add an optional `perPageOptions?: number[]` prop
  (default `[20, 40, 80]`) and render the dropdown options from it. All existing callers are
  unchanged (they still get 20/40/80).
- **`frontend/src/pages/class-detail/AuditTab.tsx`** — default the audit view to 10 per page
  (`useUrlPagination(1, 10, 'audit')`) and pass `perPageOptions={[10, 20, 40]}` so 10 is a valid,
  pre-selected option.

No backend change: the frontend sends `per_page` explicitly, `getPaginationInfo` honors it, and
`GetChangeLogEntries` already applies `Offset`/`Limit`.

### 2. Attendance duplicate records — consistent soft-delete

Fixes duplicate attendance records / inflated attendance counts. Root cause was a three-way
interaction in `backend/src/database/events_attendance.go`: attendance is **soft-deleted**, the
re-mark upsert never cleared `deleted_at`, and the raw grid/stat queries never filtered
`deleted_at IS NULL` — so a deleted-then-re-marked resident's soft-deleted row stayed visible in the
grid and every attendance statistic. Fixed per the team decision to **keep soft-delete** (records are
retained for recoverability), made consistent:

- **Upsert revive** — added `deleted_at` to `LogUserAttendance`'s `ON CONFLICT DO UPDATE` columns, so
  re-marking a previously-deleted resident resets `deleted_at` to NULL and revives the **same** row
  instead of updating a hidden, soft-deleted one. The full unique index `idx_event_user_date` is
  intentionally kept — it routes the re-mark into the revive path rather than spawning a second row.
- **`deleted_at IS NULL` on the genuinely-raw reads** — grid join (in the LEFT JOIN ON-clause, so
  enrollments without attendance still appear), attendance rate, attendance flags (all subquery
  joins), consecutive-absences, and both cumulative-rate queries.
- **No migration / no data purge** — existing soft-deleted rows are retained by design and now
  correctly excluded from reads. GORM `.Model(...)` reads (`GetAttendees`, `GetMissingAttendance`,
  `GetAttendanceCountsForEvents`) already auto-scope `deleted_at IS NULL` and were left unchanged.
- **Repo-wide coverage** — the same `deleted_at`-blind raw-query pattern was also fixed in **13 other
  attendance reads** across `dashboard.go` (≥3-absence concern counts), `programs.go` (program/facility
  attendance-rate metrics + resident engagement export), `reports.go` (all three report types), and
  `users.go` (per-user enrollment/weekly/history views), so no attendance surface counts soft-deleted
  rows. LEFT JOINs got the predicate in the ON-clause; raw WHEREs got it appended.

### 3. Dependency bump (security) — clears `govulncheck` GO-2026-5764

Backend-touching PRs are blocked by the `security-go` CI check until this is bumped (pre-existing,
unrelated to the changes above). Upgrades `aws-sdk-go-v2/service/s3` (→ v1.105.0) and its transitive
`aws/protocol/eventstream` (→ v1.7.14) past the patched versions in both scanned modules (`backend`,
`provider-middleware`), with `go work sync` aligning `backend/migrations` and `backend/seeder`.
Dependency metadata only — no source changes. `govulncheck ./...` now reports **0 code-affected
vulnerabilities** in both modules. (Same bump as PR #1190.)

- **Related issues**: Closes Asana ID-XXX (Audit History) and ID-XXX (attendance duplicates).

## Follow-up PR

Only lower-priority **hardening** remains deferred (not required for correctness now that all reads
filter soft-deletes): converting `idx_event_user_date` to a partial unique index
(`WHERE deleted_at IS NULL`) and migrating `date VARCHAR(64)` to a real `date` column type so
inconsistent date strings can't defeat uniqueness.

## Screenshot(s)

audit_page2_11-16


carddev81:
@corypride can you please fix the github conflicts when you get time, I want to make sure your version of code runs

CK-7vn (approved):
Just merge conflicts and then you're all set to go my friend. Good work.

PR #1193, opened 2026-07-14
fix: cap statewide program completion rate at 100% (ID-716)

CK-7vn:

## Description of the change
FetchEnrollmentMetrics counted Completed enrollments in the numerator without requiring enrolled_at IS NOT NULL, but the denominator required it. Completed enrollments with a NULL enrolled_at (common: completing a class never backfills enrolled_at for rows created while the class was not Active) landed in the numerator but not the denominator, so the rate could exceed 100% (reported 130%).

Constrain the numerator to the same enrolled_at IS NOT NULL base as the denominator. Added a regression test asserting the rate never exceeds 100%.

- **Related issues**: https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1215727990979790?focus=true


carddev81 (approved):
Tested good and all program completions stayed at or below 100%!

PR #1201, opened 2026-08-04
Feature flagging at a per-facililty level

corypride:

## Description of the change

Adds **per-facility feature flags** so a department/system admin can turn Knowledge Center, Learning
Platforms, Program Hub & Tracking, and Learning Record on/off for an individual facility, layered on
top of the existing statewide flags. Effective access = the statewide floor minus whatever a facility
has explicitly disabled; a facility with no override row inherits the floor untouched, so rollout is
a no-op for every existing facility. This supersedes an earlier attempt at this ticket (PR #1190,
`cpride/feature_flags`) that was rewritten from scratch per reviewer feedback — this PR is a single
unified page (not two), sub-features actually function, and "apply to all" includes them too.

- **Migration `00071`** — new `facility_feature_flags` table (composite PK `(facility_id, feature)`,
  FK to `facilities` ON DELETE CASCADE, no `default` on `enabled` so a `false` insert is never
  silently dropped). Also flips the `learning_record` statewide floor from off to on, since this page
  is now the only way to control it and the old complaint ("Learning Record can never be turned on")
  would otherwise persist.
- **Model** (`feature_flags.go`) — `FacilityFeatureFlag`, `TopLevelFeatures`, `SubFeatureParent`.
- **Database layer** (`facility_feature_access.go`) — `GetFacilityFeatureAccess` (effective-access
  resolver, with parent→sub-feature cascade), `GetFacilityFeatureOverview` (list panel),
  `GetFacilityFeatureDetail`, `UpsertFacilityFeatureFlag` (with statewide-floor and parent-off
  guards), `ApplyFacilityFeaturesToAll` (propagates top-level *and* sub-features in one transaction).
- **Enforcement rescope** (`auth.go`, `middleware.go`, `user_handler.go`) — `Claims.FeatureAccess` is
  now resolved per-facility for **every** role via one formula, no System Admin carve-out; toggling a
  feature actually gates nav + API access, not just a DB row.
- **Handlers/routes** (`facility_feature_access.go`) — `GET /api/facilities/features`,
  `GET /api/facilities/{id}/features`, `PUT /api/facilities/{id}/features/{feature}`,
  `PUT /api/facilities/features/apply-all`, all behind `newDeptAdminRoute` (system + dept admin only).
  Old global toggle endpoints and their now-dead handler/DB functions removed
  (`feature_access.go` deleted; demo-seed handler carved out into `demo_seed_handler.go`).
- **Frontend** — `FeatureControl.tsx` rewritten as a single Exports-style list+detail page: left card
  (search + compound On/Off filter per feature + facility list with pills/"X of 4 on" counts), right
  card (selected facility's feature cards with nested Knowledge Center sub-toggles + "Apply these
  settings to all facilities" behind a confirm dialog). Route moved from System-Admin-only to
  `deptAdminRoutes`; Sidebar gating changed to `canSwitchFacility`, so System Admin and Department
  Admin see the byte-identical page.

- **Knowledge Center sub-feature gating** (`KnowledgeCenterManagement.tsx`,
  `ResidentKnowledgeCenter.tsx`) — the "Helpful Links" and "Videos" tabs now check
  `hasFeature(user, FeatureAccess.HelpfulLinksAccess)` / `UploadVideoAccess` before rendering, matching
  the backend's per-facility enforcement. Previously these always rendered regardless of the
  sub-feature toggle (see bugs below).
- **Video route gating fix** (`backend/src/handlers/video_handler.go`) — `registerVideoRoutes` gated
  video viewing/favoriting on the parent `OpenContentAccess` only, requiring `UploadVideoAccess`
  solely on the add-video route. Now uses `UploadVideoAccess` uniformly for every video route,
  matching the pattern already used for helpful links (`registerLeftMenuRoutes`), so the sub-feature
  actually restricts viewing, not just uploading.
- **Request Content button** (`ResidentKnowledgeCenter.tsx`, `lib/validation.ts`) — the
  `request_content` sub-feature previously had no corresponding UI anywhere; the backend endpoint
  (`POST /api/open-content/request-content`) existed but nothing called it. Added a "Request Content"
  button to the resident-facing Knowledge Center, gated on
  `hasFeature(user, FeatureAccess.RequestContentAccess)`, opening a form (new `requestContentSchema`)
  that posts `{ content }` to the existing endpoint. This was the last of the four sub-features to
  actually have a working UI toggle.

Commits: backend (queries → enforcement → routes → tests) → frontend rewrite → a cascade bug fix
found during manual testing → a Knowledge Center sub-feature UI-gating fix → a video-route gating fix
→ a Request Content button for residents → CodeRabbit review fixes → a favorites feature-gating fix
and the ResidentHome empty-section fix, both found during a later resident-driven pass (see below).
No unrelated commits carried along.

- **Related issues**: Closes Asana ID-719.

## Screenshot(s)

image

## Additional context

- **Testing** (25/26 test-kit cases passed, 1 bug found and fixed, 1 pending): full backend
  integration suite passes, including the new facility feature flag tests and a regression test for
  the cascade fix below. Driven end-to-end via Playwright across **SuperAdmin, a department admin,
  a facility admin, and a resident student**: list/pills/counts, the compound filter, search
  (substring + case-insensitive, with a graceful empty state when the selected facility falls outside
  the filter), per-facility toggles (including the acting admin's own sidebar updating live with no
  reload), the apply-all confirm/cancel flow, and role gating (facility admin blocked at both the
  route and API level), all with 0 console errors. Only **migration rollback** is still pending — not
  run against the shared dev DB; verify on a throwaway DB before merge.
- **Second verification pass, driven as an actual resident** (not an admin previewing the resident
  pages): confirmed a student's claims carry the per-facility set; that revoking `helpful_links` /
  `upload_video` 401s `/api/helpful-links` and `/api/videos` and hides both tabs; that turning off
  only the parent drops all three sub-features from a student's claims with *no override rows of
  their own* (the cascade computed, not read) and 401s every Knowledge Center endpoint; and that the
  Request Content button reaches its handler. This pass is what surfaced the fourth and fifth bugs
  below — neither was reachable from the admin-side testing above.
- **Bug found and fixed in this same PR**: disabling a parent feature (e.g. Knowledge Center) at a
  facility correctly excluded it from a resident's effective access, but its 3 sub-features
  (`request_content`, `helpful_links`, `upload_video`) stayed effectively enabled because their own
  override rows were untouched — `GetFacilityFeatureAccess` had no parent→child cascade. This is
  exactly the "sub-features must actually work" defect flagged in the prior PR #1190's review. Fixed
  before this PR was opened; see commit `58606988` and its regression test.
- **Second bug found and fixed, later**: the backend correctly gated sub-feature APIs (e.g.
  `GET /api/helpful-links` 401s when `helpful_links` is off), but the frontend never hid the
  corresponding tab/button — the "Helpful Links" tab stayed visible and clickable in both the
  admin/facility-admin Knowledge Center management page and the resident-facing Knowledge Center, just
  silently 401ing. No data leak, but misleading UX and inconsistent with the top-level nav's own
  `OpenContentAccess` gating. Fixed by adding `hasFeature` checks to both components.
- **Third bug found and fixed, later**: unlike `helpful_links`, the `upload_video` sub-feature never
  actually gated anything beyond the "Add Video" button — `GET /api/videos` was only gated on the
  parent `OpenContentAccess`, so disabling `upload_video` left video viewing (API and UI) fully
  functional for both admins and residents. Fixed by gating all video routes on `UploadVideoAccess`
  (matching the helpful-links pattern) and hiding the Videos tab/content in both frontend components.
- **Fourth bug found and fixed, later**: gating the video *routes* turned out not to be sufficient —
  `GET /api/open-content/favorites` and `/api/open-content/favorite-groupings` are gated on the
  parent `OpenContentAccess` only (correctly, since libraries must keep working), and their queries
  are fixed three-branch UNIONs over libraries + videos + helpful links with no per-content-type
  filter. So with `upload_video` off, a resident's Favorites tab still rendered a favorited **video**
  card — title, description, channel, thumbnail — while `GET /api/videos` and `/api/videos/{id}` both
  correctly 401'd. The same gap applied to `helpful_links`. Found by driving the app as an actual
  resident rather than an admin. Fixed by plumbing the caller's effective feature set through
  `models.QueryContext` and emptying the video / helpful-link branches of both favorites queries when
  the corresponding sub-feature is off. Favorites are **hidden, never deleted** — the
  `open_content_favorites` rows are untouched, so re-enabling the feature restores them.
- **Fifth bug found and fixed, later**: `ResidentHome.tsx` (previously flagged here as an unrelated
  follow-up) rendered two Knowledge Center shells when the feature was off — the "Pick Up Where You
  Left Off" section and the Favorites sidebar, the latter reading "No Favorites Yet". Neither had a
  `.length > 0` guard of its own, so both rendered empty with "View All" buttons pointing at a gated
  route. Both are now gated on `OpenContentAccess`, and the page's three Knowledge Center SWR keys
  are conditional so the requests are skipped rather than fired-and-discarded. Folded in here rather
  than deferred, since it's the same per-facility gating story.
- **Dead fetches removed from the `/home` loader** (found while chasing the 401s the fix above did
  *not* silence): `getStudentLevel1Data` issued six blocking requests per navigation, but
  `ResidentHome` reads only `topUserContent` / `topFacilityContent` from it — helpful links,
  favorites and featured libraries come from the page's own SWR hooks, and nothing anywhere read
  `libraryOptions`. The four surplus requests were fetched, JSON-serialized into loader data, and
  discarded, and each 401'd with Knowledge Center off. Removed them and gated the two that remain on
  `OpenContentAccess`. `/home` now loads with **zero** application console errors in both states, and
  the route transition makes a third of the requests it used to. The loader feeds this one route only
  (`app-routes.tsx`), and `ResidentHomeData` was trimmed to what is actually returned.
- **Database-layer logic**: the guard/cascade rules (statewide-floor check, parent-off check, cascade
  computation) live in the database layer alongside the queries, mirroring the existing global
  `feature_access.go`'s established pattern in this codebase. Flagging for reviewer preference — left
  the "no business logic in the database layer" box unchecked accordingly.
- A ready-to-run test kit (seed script + detailed test cases + results) lives in `pr_notes/PR719/`.

---
  - https://app.asana.com/0/0/1216014970136949
  - https://app.asana.com/0/0/1216726852749170


backend/migrations/00071_add_facility_feature_flags.sql
```diff
@@ -0,0 +1,29 @@
...
+    update_user_id INTEGER,
+    created_at timestamp with time zone,
+    updated_at timestamp with time zone,
+    PRIMARY KEY (facility_id, feature),
+    FOREIGN KEY (facility_id) REFERENCES public.facilities(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE INDEX idx_facility_feature_flags_feature ON public.facility_feature_flags USING btree (feature);
+
+-- Per-facility control now supersedes the statewide staged-rollout gate that
```
carddev81: why are we setting this to TRUE? or better was there a reason we were setting this to FALSE before?

carddev81 (changes requested):
1. For learning platforms turned off there is a call to canvas occurring even thought 
server-1            | {"error":"canvas API returned 401","level":"warning","msg":"failed to fetch canvas courses for provider 1, skipping","time":"2026-08-06T18:30:41Z
2. per @calisio make sure to set all flags to false, turned off by default. So I would leave out the flag update to true (just remove the sql out of the mi8gration file)

backend/migrations/00071_add_facility_feature_flags.sql
corypride: Good catch — per @calisio's guidance that flags should stay off by default, I've removed the UPDATE ... SET enabled = TRUE/FALSE statements entirely. This migration now only adds the facility_feature_flags table; learning_record stays at whatever migration 00069 already set it to (FALSE), and facilities can opt in individually once this table exists.

carddev81:
@corypride Just checked the feature flagging pr...there still is the call to canvas being made even though I turned learning platforms off. It is when I click on programs or classes, you can tell by the following output line:

server-1 | {"error":"canvas API returned 401","level":"warning","msg":"failed to fetch canvas courses for provider 1, skipping","time":"2026-08-06T18:30:41Z

it fails to execute because I do not have canvas completely set up but if i did have it set up it would pull classes from canvas even though i turned that feature off.

carddev81 (approved):
@corypride code looks good and tested good.

PR #1203, opened 2026-08-05
Add Learning record privacy notice

corypride:

## Description of the change

Residents entering achievements into Learning Records had no on-screen explanation of how their data
is used. This adds a persistent, low-visual-weight transparency notice to both resident-facing
surfaces. It is informational only — not dismissible, and it never gates or delays starting an entry.

| File | Change |
|---|---|
| `data/learningRecordResidentCopy.ts` | + `LEARNING_RECORD_PRIVACY_NOTICE` (`short` / `full`) — one constant, so the two placements can't drift and final wording is a one-file edit |
| `components/learning-record/LearningRecordPrivacyNotice.tsx` | **new** — wraps the existing shadcn `Alert`, styled to match `IncompleteEntryReminder` |
| `digital-transcript/DigitalTranscriptHome.tsx` | `short` variant, in **both** prototype branches, above the "You have X achievements saved here" heading and the "Add achievement" CTA |
| `digital-transcript/DigitalTranscriptWysiwygEntry.tsx` | `full` variant, directly above the My Achievement / My Experience / My Future progress bar |
| `data/learningRecordResidentCopy.ts` (again), `data/faqData.ts`, `student/ResidentHome.tsx` | rewrote four existing strings that contradicted the notice — see below |

Frontend only — no migration, model, handler, database-layer or seeder changes.

Two non-obvious bits, both commented in place — they look like cruft but aren't:

- **`role="note"`** overrides `Alert`'s built-in `role="alert"`, which is an assertive live region and
  would make screen readers interrupt on every page load.
- **`w-auto`** on the entry-form notice — `Alert` ships `w-full`, which combined with `mx-4` makes the
  box 32px wider than its container, pushing the right border past the edge where `overflow-hidden`
  clips it. The sibling progress `Card` needs no override because `Card` has no `w-full`.

### Where to see each change

Log in as a **resident**. `learning_record` ships disabled (migration `00069`), so enable it in
`/feature-control` first — note a flag flipped directly in SQL won't take effect until the server
restarts, since it's read from an in-memory cache populated at boot and refreshed only by the admin
toggle.

|---|---|---|
| `/learning-record-funnel` | Notice added | Below the `<h1>`, **above** "You have X achievements saved here" and the Add-achievement CTA |
| `/learning-record-categories` | Notice added | Same notice, above the start card — confirms both prototype variants |
| `/learning-record-funnel/entry` | Notice added | Directly above the My Achievement / My Experience / My Future progress bar. Check the **right** border sits inside the panel, and that you can type immediately — it must not gate the form |
| `/learning-record-categories/entry` | Notice added | Same; no progress bar in this variant, which is correct |
| `/home` | Start-card copy | Green "Log a new achievement" card now ends "…facility staff don't see your individual entries" |
| `/home` → "What does this mean?" | Two FAQ strings | Dialog shows both rewritten answers |
| `/learning-record-funnel` → "View all achievements" → "What does this mean?" | Same two strings | Proves the shared dialog reaches this surface too |
| `/faqs` → "Managing My Account" | All three FAQ strings | Open *print or share*, *Where is my record saved?*, and *Is my activity private?* **one at a time** — the accordion is `type="single"`, so opening one closes the last |

Empty-state check: log in as a resident with zero entries and load either landing route — the notice
still renders above the empty state.

- **Related issues**: Closes Asana ID-807

## Screenshot(s)

All resident-facing at 1366×768.

entry_categories
entry_funnel
help_modal_both_answers
start_card_blurb
achievements_sheet_modal
landing_funnel
landing_categories
faq_activity_private
faq_print_share
faq_saved_here

## Additional context

**Copy approved by Carolina** (2026-08-06). Both variants live in one constant, so any future wording
change is a single edit.

### Four existing strings said the opposite — fixed here

The app already told residents things the notice contradicts. Three of them sit in the **same**
"Managing My Account" accordion on `/faqs` — all three questions visible in one frame, so a resident
opening them in turn met the contradiction directly. Carolina approved the rewrite and asked that the "what does this
mean" modal and the "view all achievements" popout be updated too — which they were by construction,
since the two Learning Record strings are single-sourced and the modal and the FAQ render the same
constants (**five render sites**: both landing variants, the achievements popout, the resident home
start card, and `/faqs`).

| Was | Now |
|---|---|
| "Your record stays private and is only seen by people you give it to." | "The file you create is only seen by the people you give it to." — scoped to the PDF |
| "It is not shared with anyone automatically. Nothing leaves this app unless you choose to print or share it." | Keeps the print/share promise, adds the aggregate carve-out **verbatim from the approved notice**, so the two cannot drift |
| "The UnlockEd team and authorized facility staff can see this information." | Scoped to library and course activity, with Learning Record entries explicitly excluded |
| Resident home start card: "…Nothing leaves this app unless you choose to print or share it." | "…Your answers are saved here as you go, and facility staff don't see your individual entries." |

The fourth one is worth a note. It sat in JSX wrapped across five source lines, so no single line held a
matching phrase and two grep sweeps missed it — it was caught by eye, in a screenshot, sitting directly
above its own "What does this mean?" link and contradicting the modal one click away. It's now moved
into the copy module as `LEARNING_RECORD_START_CARD_BLURB`, so **every resident-facing Learning Record
privacy claim is in one file** where a plain grep will find it. If you're checking this kind of thing in
future, normalize whitespace per file before matching; line-based grep gives false confidence on JSX.

Note the aggregate use the notice discloses **isn't built yet** — nothing currently aggregates
`learning_record_entries`. The wording is deliberately permissive ("*may* be used"), so it's accurate
now, stays accurate once aggregation ships, and won't need re-approval. Carolina was told this before
approving.

### Two pre-existing bugs found while testing — neither fixed here

This diff touches no backend files; both reproduce on `main` without it.

1. **`PUT /api/learning-record/draft` returns 500 for every resident, silently** — the autosave label
   still reads "Saved". `UpsertLearningRecordDraft` (`database/learning_record.go:99-102`) infers
   `ON CONFLICT (user_id, client_id)`, but the only matching unique index is partial
   (`... WHERE deleted_at IS NULL`, migration `00069`) and Postgres can't infer a partial index →
   `SQLSTATE 42P10`. It's a planning failure, not a row conflict: it fails even for a brand-new key
   with zero conflicting rows. Reflection content is safe (that saves via `POST`/`PUT /entries`); only
   the draft/resume row is lost, and only cross-device since the frontend also keeps it in
   `localStorage`. Fix is one clause —
   `TargetWhere: clause.Where{Exprs: []clause.Expression{clause.Expr{SQL: "deleted_at IS NULL"}}}`.
   It must be `TargetWhere`, not `Where`: GORM emits the former as the conflict-target predicate, the
   latter after `DO UPDATE SET`, where it wouldn't help. Happy to open a separate PR.

---
  - https://app.asana.com/0/0/1217097620641819

### DIFF
--- a/frontend/src/components/learning-record/LearningRecordPrivacyNotice.tsx
+++ b/frontend/src/components/learning-record/LearningRecordPrivacyNotice.tsx
@@ -1,0 +1,52 @@
+import { ShieldCheck } from 'lucide-react';
+import { Alert, AlertDescription, AlertTitle } from '@/components/ui/alert';
+import { LEARNING_RECORD_PRIVACY_NOTICE } from '@/data/learningRecordResidentCopy';
+import { cn } from '@/lib/utils';
+
+export interface LearningRecordPrivacyNoticeProps {
+    /**
+     * `short` — Learning Record landing page.
+     * `full` — achievement entry form (expanded wording, same substance).
+     */
+    variant: 'short' | 'full';
+    /** Caller owns spacing; the notice itself sets no margin. */
+    className?: string;
+}
+
+/**
+ * Persistent transparency notice telling residents how their Learning Record
+ * data is used. Informational only — deliberately not dismissible and never a
+ * gate: it must not block or delay starting an entry.
+ *
+ * Copy lives in `LEARNING_RECORD_PRIVACY_NOTICE` so both placements stay
+ * consistent and Carolina's final wording is a one-file change.
+ */
+export function LearningRecordPrivacyNotice({
+    variant,
+    className
+}: LearningRecordPrivacyNoticeProps) {
+    const copy = LEARNING_RECORD_PRIVACY_NOTICE[variant];
+
+    return (
+        <Alert
+            /*
+             * `Alert` hardcodes role="alert" (an assertive live region), which
+             * would interrupt screen readers on every page load. This notice is
+             * passive, so downgrade to `note`. Safe because `Alert` spreads
+             * props after its own role.
+             */
+            role="note"
+            data-slot="learning-record-privacy-notice"
+            className={cn('border-[#556830]/30 bg-[#556830]/5', className)}
+        >
+            <ShieldCheck
+                className="text-[#556830] dark:text-primary"
+                aria-hidden
+            />
+            <AlertTitle className="text-foreground">{copy.title}</AlertTitle>
+            <AlertDescription>
+                <span className="text-sm leading-relaxed">{copy.body}</span>
+            </AlertDescription>
+        </Alert>
+    );
+}
--- a/frontend/src/data/faqData.ts
+++ b/frontend/src/data/faqData.ts
@@ -104,7 +104,7 @@ export const FAQ_CATEGORIES: Record<string, FAQEntry[]> = {
         },
         {
             question: 'Is my activity on UnlockEd private?',
-            answer: 'The UnlockEd team and authorized facility staff can see this information.'
+            answer: "The UnlockEd team and authorized facility staff can see your library and course activity. Your Learning Record entries are separate — staff don't see your individual entries there."
         }
     ],
     'Getting Help and Troubleshooting': [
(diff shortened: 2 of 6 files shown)


carddev81 (approved):
@corypride Great Work!! Looks good and tested great!!!!!

PR #1206, opened 2026-08-11
Ck 7vn/tutor merge

CK-7vn:

## Description of the change
This branch has the adapter layer for the tutor application while feature-flagging the HiSET Tutor. 

---
  - https://app.asana.com/0/0/1217418401578380


backend/migrations/00072_add_ai_tutor_feature_flag.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
+-- +goose NO TRANSACTION
+ALTER TYPE feature ADD VALUE IF NOT EXISTS 'ai_tutor';
+
+INSERT INTO public.feature_flags (name, enabled)
+VALUES ('ai_tutor', FALSE)
+ON CONFLICT (name) DO NOTHING;
+
+-- +goose Down
+DELETE FROM public.feature_flags WHERE name = 'ai_tutor';
```
carddev81: missing the drop of the enum type ai_tutor, reference the script named: 00072_add_resident_programs_feature.sql for how to do the goose Down for database type enums. A little tricky, but has been consistent for the most part here within the good Down

backend/migrations/00072_add_ai_tutor_feature_flag.sql
```diff
@@ -0,0 +1,10 @@
+-- +goose Up
```
carddev81: Make sure to increment your file here after you do a rebase with main, a couple migrations has happened since this PR was submitted

backend/migrations/00072_add_ai_tutor_feature_flag.sql
corypride, line 1: Make sure to rename this

carddev81 (changes requested):
1. When running make init, process fails because unable to prepare context: path ".../wkspcs/pr_reviews/ai/unlocked-hiset-ai" not found. I believe we should still be able to run `make init` or `make dev` when running Unlockedv2 w/out the tutor. We shouldn't fail the process. Currently the user has to checkout ai/unlocked-hiset-ai in order to run successfully. If hiset tutor is unavailable what should the iframe shouldn't display nginx 502 page rendered inside the UnlockEd chrome.
2. Unable to turn on AI Tutor. (If you rebase with latest from main this issue should be fixed)
3. Feature does not exist within the facility feature cards also the number of features should be 5 now. 

Image

4. Initial logon of resident causes error when clicking AI Tutor link, after first click and then click again on the same link it works.

Image

5. clicking students name in top right corner causes the window to double, see pic below.

Image

tutor-integration.md
corypride: This may be better suited in the docs directory.

corypride (changes requested):
---

## 📝 My Notes (from testing)

### Questions about AI Tutor as a whole
- Is `tutor-integration.md` meant to stay as a permanent reference doc, or is it a one-time handoff note that should get trimmed/archived after this merges? Its own header calls it "Context for Part B... uncommitted" and reads like a punch list of what's proven vs. not — worth a quick call from the author either way, since `README.md:17` already links to it as if it's permanent, and Findings 1/2 above are only worth fixing if it's staying.

### Wanted to test but couldn't

> Roles (automatic, by UnlockEd role): staff (`system_admin`/`facility_admin`/`department_admin`) land in the tutor's teacher view; residents get the student experience. `/tutor/teacher` is staff-gated server-side. A SuperAdmin (the `system_admin` role or the username `SuperAdmin`) additionally gets a Students picker to impersonate any student for QA, and a Stop impersonating button to return. No one else sees the picker.

**Blocked — no workaround.** Still need to check the teacher view, the student view, and the SuperAdmin impersonation picker + "Stop impersonating," but there's no way to turn the feature on through the app to get there (see Finding 3 — it's not just the per-facility toggle, there's no UI path to the statewide flag at all). Not testable until that's fixed.

### Observations
- `tutor-integration.md` says: `TUTOR_FEATURE_FLAG=ai_tutor`, plus one AI key (`ANTHROPIC_API_KEY`/`GEMINI_API_KEY`/`OPENAI_API_KEY`) — set in `.env`, compose reads them through. Turns out `TUTOR_FEATURE_FLAG` isn't actually something you set yourself — it's hardcoded in `docker-compose.yml`. Rest of what this section gets wrong is in Finding 1.
- Mastra/node errors in the log on `make dev-tutor` — app still builds despite them. Haven't dug into what's throwing these yet, flagging so it doesn't get lost.

Image

- Unable to turn Tutor on as SuperAdmin — real bug, not user error, and it's a hard blocker: tried the per-facility toggle, then "Apply to All Facilities" too — neither works, because neither one is capable of setting the flag the tutor actually checks. Details in Finding 3.
- Took a few seconds for the tutor to render on the first click of the AI Tutor nav link — render time sped up on every try after that. Confirmed this is just `make dev-tutor` compiling routes on first hit (Next.js dev server), not a real perf issue — not worth chasing.
- View as student: *(didn't get here either)*

---

## 🔍 Findings

### Finding 1: `tutor-integration.md` env-setup instructions don't warn that a missing/mismatched AI key fails silently

- **File**: `tutor-integration.md:108-109`
- **Issue**: Tells a dev to set one AI key (`ANTHROPIC_API_KEY`/`GEMINI_API_KEY`/`OPENAI_API_KEY`) with no mention of what happens if it's missing or mismatched with the active provider. In practice: `AI_PROVIDER` defaults to `anthropic` (`docker-compose.yml:137`), so `GEMINI_API_KEY`/`OPENAI_API_KEY` alone (without also setting `TUTOR_AI_PROVIDER`) silently fails. The tutor boots clean either way — every AI route guards with `isAIConfigured()` and returns a plain JSON "unconfigured" response, no thrown error, nothing in `docker compose logs`. The only signal is an in-app message on first chat send.
- **Fix**: Amend `tutor-integration.md:108-109` to state that `AI_PROVIDER` defaults to `anthropic`, and that a missing/mismatched key fails silently (no terminal error).

### Finding 2: `tutor-integration.md` schema section is stale — table count and provisioning mechanism are both wrong

- **File**: `tutor-integration.md:111-115`
- **Issue**: Says "the tutor owns only `tutor.*` (8 tables)" — actual count is 21 (verified by counting `CREATE TABLE` statements across `lib/db/migrations/*.sql` in the sibling repo). It also describes provisioning via `docker/apply-schema.mjs` reading `docker/tutor-schema.sql` — neither file exists anymore; `docker/migrate.mjs` replaced them (applies the Drizzle migrations from `lib/db/migrations` directly).
- **Fix**: Update the table count (or drop the specific number to avoid re-staling) and replace the `apply-schema.mjs`/`tutor-schema.sql` description with `docker/migrate.mjs`. Also update the doc's 4 remaining `00072_add_ai_tutor_feature_flag.sql` references — that migration was renumbered to `00074` locally to resolve a version collision with `main`'s `00072_add_resident_programs_feature.sql`.

### Finding 3: 🔴 BLOCKER — there is no way to turn AI Tutor on through the admin UI at all

- **Severity**: Blocker. Not "the per-facility toggle has a bug" — there is no UI action, anywhere in this app, that a reviewer or admin can take to make the tutor actually work.
- **Root cause**: the tutor-service's kill-switch (`lib/adapters/identity/index.ts:48-55`, `isFlagEnabled`, called from `require-user.ts:20`) checks *only* the statewide `feature_flags` table — never `facility_feature_flags`. But every write path in `FeatureControl.tsx` only ever writes `facility_feature_flags` rows:
  - The per-facility switch → `PUT /api/facilities/{id}/features/{feature}` → writes a `FacilityFeatureFlag` override row.
  - "Apply to All Facilities" → `PUT /api/facilities/features/apply-all` → `ApplyFacilityFeaturesToAll` (`backend/src/database/facility_feature_access.go:184`) → also only writes `FacilityFeatureFlag` override rows to every other facility.
  - Checked every handler in `facility_feature_access.go` — the *only* function anywhere in the backend that touches the statewide `feature_flags` table is a read-only `GetFeatureAccess()`. Nothing writes to it. The migration's one-time `INSERT ('ai_tutor', FALSE)` is the only thing that has ever set that row.
  - So no combination of UI actions can ever satisfy the tutor's check — not the per-facility toggle, not "Apply to All Facilities," regardless of how many facilities exist.
- **Confirmed live**: `feature_flags.ai_tutor.enabled = FALSE` (statewide, untouched by any UI action) with `facility_feature_flags` override `TRUE` for facility 1 → 6 consecutive `403 Forbidden` on `/tutor/api/me` 

Image
 
Image

Also confirmed "Apply to All Facilities" doesn't change this — it never touches the statewide row (in this dev DB there's only one facility, so the bulk-copy itself was a no-op too, but that's incidental — the underlying gap exists regardless of facility count). The *only* way I got it working was a direct SQL `UPDATE feature_flags SET enabled = TRUE` against Postgres, which no admin/reviewer has a UI path to do (`images/ai-tutor-working-statewide-on.png`).
- **Fix direction**: `isFlagEnabled`/`isTutorEnabled` need facility context — either move the check to after the caller's `facilityId` is resolved and query facility-aware (mirroring `resolveFeatures`'s override-wins-over-statewide logic), or add a second facility-aware check after the existing fast statewide pre-check. This blocks any real QA of the feature until it's fixed — not something to work around with a manual DB edit for review purposes.

**Not included above** (already covered by Rich's review or CodeRabbit, no need to restate): AI Tutor missing from `TOP_LEVEL_FEATURES`/the facility overview cards; the migration missing a `Down` for the `ai_tutor` enum value.

backend/migrations/00075_add_ai_tutor_feature_flag.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
```
carddev81, line 1: Migration file needs to be incremented to 00075 due to main having 74

backend/seeder/seeder
carddev81, line 1: This file can be removed

carddev81 (changes requested):
1. Unlocked fails to start when starting Unlocked with the path ai/.... See video below too for issue.

Steps to reproduce:
1. Run unlocked `make init` w/out ai folder next to it
  - successfully starts and ai tutor works as expected
2. Run unlocked `make dev` w/ai folder next to it
  - unlocked fails to start

getting this issue:

```
rev_proxy-1         | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
rev_proxy-1         | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
rev_proxy-1         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
rev_proxy-1         | 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
rev_proxy-1         | 10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
rev_proxy-1         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
rev_proxy-1         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
rev_proxy-1         | /docker-entrypoint.sh: Configuration complete; ready for start up
rev_proxy-1         | nginx: [emerg] host not found in upstream "tutor-service" in /etc/nginx/conf.d/default.conf:65
rev_proxy-1 exited with code 1
```
In order to get it to work i have to drop the containers and volumes and start fresh. I can then run make dev and it run successfully

https://github.com/user-attachments/assets/e192466e-6199-4e3e-9859-e5b30d1dc181

I will verify the other two commands work a. `make dev-tutor` and `make dev-registry`

frontend/src/pages/admin/FeatureControl.tsx
```diff
@@ -536,6 +544,42 @@ export default function FeatureControl() {
                                         </div>
                                     );
                                 })}
+                                <div className="bg-background border border-border rounded-lg p-6">
```
carddev81: See line 47-76, noticed that FEATURE_CARDS wasn't utilized here. I suggest for consistency adding the configuration/meta for the Ai tutor feature here

carddev81 (changes requested):
NOTE: Using 'eval-harnesscontent' branch getting the following errors:
 
1. Unlocked fails to start when starting Unlocked with the path ai/.... See video below too for issue.Steps to reproduce:
    1. Run unlocked `make init` w/out ai folder next to it  
      - successfully starts and ai tutor works as expected
    2. Run unlocked `make dev` w/ai folder next to it
      - unlocked fails to start

getting this issue:

```
rev_proxy-1         | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
rev_proxy-1         | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
rev_proxy-1         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
rev_proxy-1         | 10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
rev_proxy-1         | 10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
rev_proxy-1         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
rev_proxy-1         | /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
rev_proxy-1         | /docker-entrypoint.sh: Configuration complete; ready for start up
rev_proxy-1         | nginx: [emerg] host not found in upstream "tutor-service" in /etc/nginx/conf.d/default.conf:65
rev_proxy-1 exited with code 1
```
In order to get it to work i have to drop the containers and volumes and start fresh. I can then run make dev and it run successfully

https://github.com/user-attachments/assets/e192466e-6199-4e3e-9859-e5b30d1dc181

I will verify the other two commands work a. `make dev-tutor` and `make dev-registry`

2. Something to think about in the future, is if the tutor is missing from the app the tutor will take down the entire app. For example someone that forks our repo will be unable to start unlocked repo due to the tutor image being private and therefore won't be able to pull that image, we need to make the app ignore the failure and still run w/out the image. The reason i bring this up is due to the latest change `depends_on: tutor-service` this doesn't fix the nginx issue of not starting, but if the image is unable to be pulled the rev_proxy will never start.
3. I know we don't use these often, but `make kolibri` and `make install-dep` will not work unless you do the same as you did for make dev
4.  Just found this, not sure if this is new? I'm able to serve the tutor app by typing http://localhost/tutor. I see in the middle ware it says it fixes this but doesn't? The /tutor serves the app shell with no session.

Image

carddev81:
@CK-7vn 
Issue when running `make dev` with tutor path `ai/unlocked...`, when clicking on AI Tutor getting the message `AI Tutor is temporarily unavailable - Please try again in a few minutes.`

Steps to reproduce:
1. Run unlocked make init without ai folder existing
2. Stop unlocked
3. Run unlocked make dev with ai folder existing
4. See error

ResidentSide

carddev81 (changes requested):
@CK-7vn a couple of issues:

1. ([previous review](https://github.com/UnlockedLabs/UnlockEdv2/pull/1206#discussion_r3875474535), think it may have gotten skipped) See line 47-76, noticed that FEATURE_CARDS wasn't utilized here. I suggest for consistency adding the configuration/meta for the Ai tutor feature here.
2. We talked about this [Issue](https://github.com/UnlockedLabs/UnlockEdv2/pull/1206#issuecomment-5453755741)
3. Concerning the `make` commands--If a build of the backend, frontend, or provider-middleware throws an error the global build does not stop and it keeps going even though its unsuccessful. It should fail--keeps the same behavior as before. To test this all you have to do is add this line or any bad line to a dev.Dockerfile `RUN this_command_does_not_exist`.

corypride (changes requested):
Approving. Three open items from my review, none of them blocking — noting them here for the record:

Migration rollback (00075): The Down script's enum rename/recreate/cast sequence runs with no transaction safety net (NO TRANSACTION applies to the whole file, Up and Down alike). Not a new pattern this PR introduces — 00072 already shipped the identical structure. Worth a follow-up ticket to harden both, not a blocker here.

Personalization scope: app/page.tsx / app/teacher/page.tsx still advertise "personalized feedback," and the Learn section still calls out to generate it — but since no resident gets a real tutor.profiles row, it delivers generic content instead. Confirmed with Rich this is a "nay" for MVP (scope is Learn & Practice only), so this is really a "the copy and code should catch up to the scope decision" item, not a functional bug. Filing separately.

Impersonation ("View as student"): Real bug — the safety banner doesn't render after using the picker as designed, only after a hard reload. But it's not reachable in the shipped config: NODE_ENV=production is baked into the Dockerfile, ALLOW_IMPERSONATION isn't set anywhere in docker-compose.yml/.env.example, so canImpersonate is false for everyone and the picker route itself is unreachable. I could only reproduce it by manually flipping the env var for testing. Given that, and that this feature's being pulled from the PR entirely per our conversation, I'm not blocking on it — just flagging that until the removal actually lands, this is dead code with a live bug in it, not a shipped one.

carddev81 (approved):
@CK-7vn Good work. This one was a doozy!

corypride (approved):

PR #1207, opened 2026-08-13
PostHog Behavioral Graphs for SMWRC Learning Record Pilot

corypride:
# feat: PostHog deployment tagging, PII reduction, and Learning Record question-level analytics

Branch: `cpride/posthog_learning_record`
Commits: `3b70a101` (analytics pipeline), `5c16e6d5` (Learning Record instrumentation, ID-806)

## Summary

Two related changes. The first makes the analytics pipeline trustworthy: facilities actually
report, every event says which deployment it came from, resident text stops leaking into
event properties, and page-load request failures become visible. The second, ID-806, adds
question-level behavior tracking to the Learning Record form on top of that foundation.

The second is only meaningful because of the first. The pilot's SQL over
`learning_record_entries` already shows *what* residents saved; these events show *how* they
worked — which questions take long, which chip selectors get skipped, how many records
someone adds in a sitting — and that is only readable if facility traffic can be separated
from staging testing.

No backend schema or API changes.

## Where to see each change

| --- | --- | --- |
| `/learning-record-funnel/entry` | Question-level analytics on the funnel form | No visible UI change. With the `learning_record` flag on and a real PostHog key, editing fields and moving between the three steps emits `lr_question_left` / `lr_step_completed`; finishing an entry emits `lr_entry_completed`. |
| `/learning-record-categories/entry` | Intentionally **not** instrumented | Same form, non-funnel variant. The tracker is gated on `isFunnel`, so no `lr_*` events fire here. Confirm that's still what we want. |
| Any page that loads data | SWR failures now report `api_error` | Break the API (stop the backend, or block a request in devtools) and confirm one `api_error` per failed GET, with `endpoint` collapsed to `/api/programs/:id/classes` form. |
| Any page, any click | Autocapture element text is masked | `$autocapture` events still fire, but `$el_text` no longer carries the text of the clicked element — previously resident names in tables and links. |
| Login / logout | Identity is deployment-namespaced | `distinct_id` reads `<state>:<user id>`, e.g. `staging:1`. Logout clears identity and immediately re-registers the deployment tags. |
| Local dev | Analytics is off by default | With a blank `VITE_DEPLOYMENT`, no requests go to `us.i.posthog.com` at all. |

---

# Reviewer notes

## Ordering — the insights cannot be built before this merges

This is the part that is easy to get backwards. The PostHog dashboard work is **downstream**
of the deploy, not a prerequisite for it:

1. Push this branch and open the PR.
2. Merge to `main`.
3. Staging deploys and sends its first tagged events.
4. **Only now** can the PostHog insights be built or repointed. PostHog creates property and
   event definitions lazily on first ingest — `deployment`, `state`, `endpoint`, and all four
   `lr_*` events do not appear in any filter dropdown until at least one has arrived. Building
   the filters early doesn't fail loudly; the properties simply aren't offered and it looks
   broken.
5. Cut the state branches; facility data arrives; verify it is tagged as expected.

**The one pre-merge PostHog config item is already done: session replay is verified OFF**
(Settings → Session replay → "Record user sessions"). That check mattered because screen
recording in a corrections context would capture resident information displayed on screen, and
because the `defaults: '2026-01-30'` preset passes session-recording configuration while the
actual toggle is project-side — so it needed verifying rather than assuming. **Events already
transmitted cannot be un-sent**, which is why it preceded the merge rather than following it.

No PostHog configuration blocks this PR. Everything else that needs configuring is downstream
of the deploy and is written up in **`POSTHOG_FOLLOWUP_PR_NOTE.md`** (same directory), phased by
what unblocks each step.

## ⚠️ Open question for reviewers — consent language vs. individual attribution

**Please weigh in on this before approving.** It is the one thing in this PR I do not think should
be settled without the person who owns the resident-facing copy.

`main` now carries `frontend/src/data/learningRecordResidentCopy.ts` — the actual resident
transparency notice, added in `3df61c7b` and shown on both the Learning Record landing page and the
achievement entry form. It commits to three facts, including that responses are combined *"with all
identifying information removed"* and that *"Nothing you write can be traced back to you
individually."*

Where the `lr_*` instrumentation actually sits against that:

| Claim in the notice | Instrumentation |
| --- | --- |
| Facility staff won't see individual answers | ✅ No answer text is ever sent. Counts, booleans, durations, fixed enums only. |
| Responses combined "with all identifying information removed" | ⚠️ **Not aggregate.** `posthog.identify()` attributes every `lr_*` event to a specific resident. |
| "Nothing you write can be traced back to you individually" | ⚠️ Literally true for content. But `char_count` is derived from what they wrote. |

Concretely, PostHog ends up holding per-resident behavioral rows — *this resident spent 45s on
`q5`, picked 3 chips, answered 8 of 10* — readable by anyone with project access. That is
individual-level behavioral data. It is not answer content, and it is genuinely useful for the
pilot, but it is not the anonymized aggregate the notice describes.

Worth noting the copy file marks itself **"FIRST DRAFT — pending Carolina's final copy,"** so the
wording is still moving. That is exactly why I did not resolve this unilaterally in either
direction: I did not want to quietly narrow working instrumentation to match a draft, nor quietly
ship instrumentation that outruns what residents were told.

**If reviewers want the narrower version, there are two independent levers:**

1. **Drop `char_count`** — one property, the only one derived from written content. Cheap; costs the
   "are residents writing a sentence or a paragraph?" signal.
2. **Send `lr_*` unattributed** — decouple these events from the identified person so they are
   aggregate-only. More invasive, and it would break `entry_index_in_session` and
   `seconds_since_session_start`, which are inherently per-person.

Either can land in this PR or a follow-up. Tell me which and I'll do it.

### Autocapture stays on, its element text is masked — `mask_all_text: true`

`posthog-js` ships `autocapture: true` with `mask_all_text: false` (verified in the installed
1.347.2 bundle — the flag is passed into the autocapture element-props builder, so masking
changes what gets serialized rather than whether the event fires). Unmasked, every
`$autocapture` click carries `$el_text`: clicking a resident's name in a table sent that name
to PostHog. That is a larger exposure than the `api_error` query strings this branch also
fixes, and it would have gone live the moment facilities started reporting.

| Option | Effect | |
| --- | --- | --- |
| Disable autocapture entirely | No `$autocapture` events. Pilot metrics unaffected. | |
| **Keep clicks, drop text** | Click and rageclick events still recorded, element text stripped. | ✅ chosen |
| Accept as-is | Resident names may appear in `$el_text`. | |

Masking is free here: none of the 8 pilot metrics depend on autocapture. Stickiness, WAU, and
Lifecycle all run on `$pageview` + `identify`, which are separate mechanisms.

### Still open — element *attributes* and personal-data properties

`mask_all_text` covers text content only. Element **attributes** are a second channel: with
only this flag set, an `aria-label` or `title` such as "Delete achievement for \<name\>" is
still captured. `mask_all_element_attributes: true` closes it at the same zero cost.

Also worth evaluating: `mask_personal_data_properties` (default off), which redacts
personal-data patterns in properties such as URLs — confirm its exact coverage against
PostHog's docs before relying on it.

**Both are deliberately not in this PR** rather than added silently. They should be decided
before facilities report. Happy to add the attribute flag to this PR if reviewers prefer.

### Query strings stripped from `api_error`

`url` previously carried the full request path including query string, which meant resident
search terms went to PostHog, and the "errors by endpoint" breakdown fragmented into one row
per typed character (`?search=Sm`, `?search=Smi`, `?search=Smit`). Now `url` is path-only and
`endpoint` additionally collapses numeric and UUID segments to `:id`.

### Learning Record properties are counts and enums only

`learningRecordAnalytics.ts` is the single place that reads entry values, and it reads only
lengths, counts, booleans, and fixed enums — never answer text. Any future property on an `lr_*`
event carrying resident text is a regression.

See the open question above for the part this does *not* settle: the events are individually
attributed, which is a weaker guarantee than the "anonymized, aggregate" framing in the resident
notice.

## Why `distinct_id` is namespaced

`user.id` is a per-database autoincrement — unique within a deployment, not across them. While
only staging reported there was one id space and the bare id was safe. With four states
reporting, a bare id would make stlouis user 1 and maine user 1 the same PostHog person:
person properties (`role`, `facility_id`, `features`) would go last-write-wins between
facilities, and weekly-active-users — the pilot's headline metric — would undercount by
collapsing distinct people into one.

There is a second-order effect worth knowing: the project's test-account filter is
`role is not system_admin`, and `role` is a person property. Without namespacing, a shared
person record would flip between `system_admin` and `facility_admin` and that filter would
intermittently drop a real pilot user.

Persons created before this change keep bare numeric ids and will not merge with the
namespaced ones. That is fine — the dashboard filter drops all pre-deploy data anyway. It does
mean a tester who used the app both before and after this change shows up as two persons, which
inflates the person counts currently on the dashboard (4 persons entering the attendance funnel,
3 the enrollment funnel — a couple of people, counted more than once).

## Reference — what the code adds

| Property | Kind | Value | Notes |
| --- | --- | --- | --- |
| `deployment` | Event (super) | `production` \| `development` | On every event incl. `$pageview`, `$exception` |
| `state` | Event (super) | `stlouis` \| `maine` \| `alaska` \| `mocode` \| `demo` \| `staging` | `unknown` if unset |
| `facility_id` | Event (super) **and** person | number | Registered at identify. Cleared on logout and not re-registered — it is user-scoped, unlike `deployment`/`state`, which are re-registered after `reset()` |
| `features` | Person | sorted comma-joined `feature_access` | New |
| `endpoint` | Event (`api_error`) | e.g. `/api/programs/:id/classes` | New; ids collapsed |
| `url` | Event (`api_error`) | e.g. `/api/programs/12/classes` | **Changed:** query string removed, canonicalized |
| `distinct_id` | Identity | e.g. `stlouis:1` | **Changed:** was the bare user id |

New events (ID-806), all on the funnel form only:

| Event | Properties |
| --- | --- |
| `lr_entry_started` | `entry_index_in_session` |
| `lr_question_left` | `question_key`, `step_id`, `kind`, `duration_seconds`, `touched`, `answered`, `selection_count`, `char_count` |
| `lr_step_completed` | `step_id`, `step_index`, `duration_seconds`, `answered_count` |
| `lr_entry_completed` | `duration_seconds`, `answered_count`, `total_fields`, `entry_index_in_session`, `seconds_since_session_start`, `submitted_at` |

Which deployments send data, after this change:

| Branch | Deployment | Reports? |
| --- | --- | --- |
| `stlouis` / `maine` / `alaska` / `mocode` | `production` | **Yes** (new) |
| `main` | `development` / `staging` | Yes (unchanged) |
| `demov2` | `development` / `demo` | No — seeded fake data is noise |
| local development | — | No, unless `VITE_DEPLOYMENT` is set deliberately |
| forks | — | No — secrets resolve empty, key fails the `phc_` check |

Analytics initializes only when the key is `phc_`-prefixed **and** `VITE_DEPLOYMENT` is set.
That single rule rejects the CI placeholder, rejects the `.env.example` sample, fails closed on
forks where the secret resolves empty, and stops a developer who pastes a real key into `.env`
from silently sending until they also name a deployment — at which point their events are
tagged and filterable rather than anonymous noise.

## Rollback

The code side reverts in one line: set `POSTHOG_KEY=placeholder_for_other_deployments` in the
frontend branch of `.github/workflows/container_builds.yml` and rebuild. The frontend rejects
any key that is not `phc_`-prefixed, so the deployments go quiet immediately without touching
the PostHog project.

The PostHog-side changes are all non-destructive filters and breakdowns; reverting them is just
changing them back. Note that **events already transmitted cannot be un-sent** — which is why
the session-replay check belongs before merge.

## Known limitations, not addressed here

Documented so they aren't rediscovered as surprises:

- **`lr_step_completed` never fires for the last step.** It fires when the resident navigates
  *away* from a step, and `entryCompleted` flushes the final step's questions without emitting
  a step event. So the completion funnel must be
  `lr_entry_started → achievement → experience → lr_entry_completed`; adding a `future` step
  would read as total abandonment. Whole-entry time is available via `duration_seconds` on
  `lr_entry_completed`; last-step time is not. Fixable, but deliberately out of scope here.
- **Aborted requests report as errors.** `api.ts` uses a 10-second `AbortController` and its
  catch block treats an abort like any other failure — `api_error` with `status: 0` and
  `message: "signal is aborted without reason"`. A genuine timeout is worth recording; a
  navigation-cancelled request is not. Currently indistinguishable. Pre-existing, unchanged.
- **Non-JSON error bodies become the `message`.** When a response isn't JSON, `api.ts` puts the
  raw body in `message` — a 502 sends roughly 700 characters of nginx HTML as an event
  property. Pre-existing, unchanged.
- **`duration_seconds` measures wall clock**, not active editing time. A resident who opens the
  form and is pulled away inflates it, with no cap or outlier trimming. This is why medians are
  the numbers to quote.
- **`*_started` events double-fire in dev** because React StrictMode invokes effects twice.
  Production builds fire once. Affects raw `*_started` counts only, not durations or funnel
  conversion.

## Follow-up PR

Nothing below blocks this PR. Full detail — including click-by-click steps and what unblocks
each phase — is in **`POSTHOG_FOLLOWUP_PR_NOTE.md`**. Summary of that scope:

- **Steps 2–7:** verify the new properties landed, then add the `deployment` filter to the pilot
  dashboard and repoint the "API errors by endpoint" and facility filters. Step 3 is the one
  with a deadline: until it's done, every dashboard tile mixes facility sessions with the team's
  staging testing.
- **Expect the dashboard to go empty when that filter lands.** Every number on it today is
  internal testing — three to four persons at near-100% conversion, and a 0.5-second average
  "program setup time" that can only be a scripted form fill. The filter hiding all of it is the
  filter working. The note records the current per-tile baseline so this isn't mistaken for a
  regression.
- **Steps 8–11:** build the four Learning Record insights — completion funnel, median time per
  question, answer/skip rate, entries per sitting. Blocked until the `learning_record` flag is
  on somewhere that reports.
- **Corrections to `POSTHOG_DASHBOARD_SETUP.md`:** the §0 event reference is now stale, and §0
  and §2c document three properties (`scheduled_start_at`, `hours_after_session`,
  `recording_delay_bucket`) that have zero occurrences in the codebase and were never shipped —
  so anyone following §2c builds an insight that stays permanently empty.

**Not in the follow-up — already resolved as pre-work:** "Attendance time — avg & median (sec)",
§2a's "core metric", was showing *no matching events*. Cause was the dashboard's saved date range,
not the insight: the test data is sparse and months old, so a rolling window missed it entirely. At
All time the tile renders normably. Verified in code that no application change was needed —
`attendance_session_completed` always carries `duration_seconds` as a number, and `useFlowTimer`
measures the correct span. Recorded in Step 2 of `POSTHOG_CONFIG_UPDATES.md` along with the
known-good insight configuration.

---
  - https://app.asana.com/0/0/1217097620641814

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,3 +35,11 @@ CORS_ALLOWED_ORIGINS=http://localhost,http://127.0.0.1
 
 VITE_PUBLIC_POSTHOG_KEY=your_posthog_project_api_key
 VITE_PUBLIC_POSTHOG_HOST=https://us.i.posthog.com
+
+# Analytics is OFF unless VITE_DEPLOYMENT is set AND the key above is a real
+# phc_-prefixed project key. Deployed images get these at build time; leave them
+# blank for local development. To deliberately send from local, set
+# VITE_DEPLOYMENT=local — every event is then tagged so it can be excluded in
+# PostHog rather than mixed into facility data.
+VITE_DEPLOYMENT=
+VITE_STATE=
--- a/.github/workflows/container_builds.yml
+++ b/.github/workflows/container_builds.yml
@@ -127,8 +127,6 @@ jobs:
                     VITE_STATE=mocode
                   elif [[ "${GITHUB_REF}" == "refs/heads/demov2" ]]; then
                     # Public demo runs on seeded fake data — noise, not signal.
-                    POSTHOG_KEY=placeholder_for_other_deployments
-                    POSTHOG_HOST=placeholder_for_other_deployments
                     VITE_DEPLOYMENT=development
                     VITE_STATE=demo
                   else
--- a/.github/workflows/container_builds.yml
+++ b/.github/workflows/container_builds.yml
@@ -109,10 +109,14 @@ jobs:
               case $path in
                 "frontend/")
                   echo "Building frontend image"
-                  POSTHOG_KEY=placeholder_for_other_deployments
-                  POSTHOG_HOST=placeholder_for_other_deployments
+                  # Deployments that report to PostHog get the real key; every
+                  # event is tagged with VITE_DEPLOYMENT + VITE_STATE so the
+                  # facilities can be told apart from staging. Builds that should
+                  # not report get a placeholder key, which the frontend rejects
+                  # because it is not phc_-prefixed.
+                  POSTHOG_KEY=${{ secrets.POSTHOG_KEY }}
+                  POSTHOG_HOST=${{ secrets.POSTHOG_HOST }}
                   VITE_DEPLOYMENT=production
-                  # Set environment-specific secrets
                   if [[ "${GITHUB_REF}" == "refs/heads/stlouis" ]]; then
                     VITE_STATE=stlouis
                   elif [[ "${GITHUB_REF}" == "refs/heads/maine" ]]; then
@@ -122,11 +126,12 @@ jobs:
                   elif [[ "${GITHUB_REF}" == "refs/heads/mocode" ]]; then
                     VITE_STATE=mocode
                   elif [[ "${GITHUB_REF}" == "refs/heads/demov2" ]]; then
+                    # Public demo runs on seeded fake data — noise, not signal.
+                    POSTHOG_KEY=placeholder_for_other_deployments
+                    POSTHOG_HOST=placeholder_for_other_deployments
                     VITE_DEPLOYMENT=development
                     VITE_STATE=demo
                   else
-                    POSTHOG_KEY=${{ secrets.POSTHOG_KEY }}
-                    POSTHOG_HOST=${{ secrets.POSTHOG_HOST }}
                     VITE_DEPLOYMENT=development
                     VITE_STATE=staging
                   fi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,8 @@ services:
     environment:
       - VITE_PUBLIC_POSTHOG_KEY=${VITE_PUBLIC_POSTHOG_KEY}
       - VITE_PUBLIC_POSTHOG_HOST=${VITE_PUBLIC_POSTHOG_HOST}
+      - VITE_DEPLOYMENT=${VITE_DEPLOYMENT}
+      - VITE_STATE=${VITE_STATE}
 
   kiwix:
     image: jamescherti/kiwix-serve:latest
--- a/frontend/src/api/api.ts
+++ b/frontend/src/api/api.ts
@@ -3,7 +3,7 @@ import {
     ServerResponseOne,
     ServerResponseMany
 } from '@/types/server';
-import { ANALYTICS_EVENTS, captureEvent } from '@/lib/events';
+import { ANALYTICS_EVENTS, analyticsUrl, captureEvent } from '@/lib/events';
 
 class API {
     private static getCSRFToken(): string {
@@ -118,7 +118,7 @@ class API {
             captureEvent(ANALYTICS_EVENTS.ApiError, {
                 status: errCode ?? 0,
                 method,
-                url,
+                ...analyticsUrl(url),
                 message: error.message || 'An error occurred'
             });
             return {
--- a/frontend/src/api/swrFetcher.ts
+++ b/frontend/src/api/swrFetcher.ts
@@ -1,0 +1,47 @@
+import { ANALYTICS_EVENTS, analyticsUrl, captureEvent } from '@/lib/events';
+
+/**
+ * Global SWR fetcher.
+ *
+ * Lives here rather than inline in `main.tsx` so failures can report to the same
+ * `api_error` event the `API.*` wrapper uses. SWR powers most of the app's
+ * page-load GETs, so before this every one of those failures was invisible to
+ * analytics — "error states hit during common flows" was missing the most common
+ * flow of all, loading a page.
+ *
+ * Errors are captured and then rethrown unchanged, so SWR's own error handling
+ * behaves exactly as it did.
+ */
+export const swrFetcher = async (url: string): Promise<unknown> => {
+    let res: Response;
+    try {
+        res = await fetch(url, {
+            credentials: 'include',
+            headers: {
+                'X-Requested-With': 'XMLHttpRequest',
+                Accept: 'application/json',
+                'Content-Type': 'application/json'
+            }
+        });
+    } catch (err) {
+        // Network-level failure: no HTTP status exists, so report 0 to match the
+        // convention in api.ts.
+        captureEvent(ANALYTICS_EVENTS.ApiError, {
+            status: 0,
+            method: 'GET',
+            ...analyticsUrl(url),
+            message: err instanceof Error ? err.message : 'Network error'
+        });
+        throw err;
+    }
+    if (!res.ok) {
+        captureEvent(ANALYTICS_EVENTS.ApiError, {
+            status: res.status,
+            method: 'GET',
+            ...analyticsUrl(url),
+            message: res.statusText || `HTTP ${res.status}`
+        });
+        throw new Error(res.statusText);
+    }
+    return res.json() as Promise<unknown>;
+};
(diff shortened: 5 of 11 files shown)


carddev81 (approved):
looks good!!

PR #1209, opened 2026-08-17
feat: restructure program class and programs per tickets 745 and 751

carddev81:

## Description of the change

Initial changes to account for the architectural changes from the original structure of programs and classes. Database changes and all silos of code changes were completed.

- **Related issues**: Link to related Asana ticket that this closes: [Update Program/Class structure](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1216770584725146?focus=true) and [Rethink hierarchy of Program/Class/Session](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1216727586577379)
 

## Screenshot(s)

image

image

image


backend/src/database/program_classes.go
```diff
@@ -789,7 +806,7 @@ func (db *DB) BulkCancelSessions(req *models.BulkCancelSessionsRequest, facility
 		Name string `gorm:"column:name"`
 	}
 	var classInfos []classInfo
-	if err := tx.Table("program_classes").
+	if err := tx.Table("program_class_cohorts").
```
CK-7vn: We should join to program_classes here for the name, i'm pretty sure. Getting a 500 issue on Bulk cancellation of sessions. 

Image

Must have just missed this one.

backend/src/database/class_enrollments.go
```diff
@@ -438,7 +518,7 @@ func (db *DB) CheckSchedulingConflicts(classID int, userIDs []int) ([]models.Con
 		}
 
 		for _, enrollment := range existingEnrollments {
-			if int(enrollment.ClassID) == classID {
+			if int(enrollment.ClassID) == cohortID {
```
CK-7vn: Unless i'm reading the migration wrong here, I think we actually need to use enrollment.CohortID instead of enrollment.ClassID, I know I deliberately tried to cause a conflict and it wasn't flagging and I'm quite sure it was because of this. Also on 525

provider-middleware/program_classes.go
```diff
@@ -41,7 +41,7 @@ func (sh *ServiceHandler) activateScheduledClasses(ctx context.Context) error {
 
 	var classIDs []int
 	if err := sh.db.WithContext(ctx).
```
CK-7vn, line 43: I think we need to change this SQL call and update all program_classes prefixed tables to program_class_cohorts prefixed calls.

provider-middleware/program_classes.go
CK-7vn: I believe this would actually be cohort_id now after the change.

backend/src/handlers/program_class_tier_handler.go
```diff
@@ -0,0 +1,188 @@
...
+	return writeJsonResponse(w, http.StatusOK, class)
+}
+
+// withResolvedCreditTypes fills in the inherited credit types for display, so callers do
+// not have to know the empty-means-inherit rule to render a class correctly.
+func (srv *Server) withResolvedCreditTypes(class *models.ProgramClass, args *models.QueryContext) *models.ProgramClass {
+	if len(class.CreditTypes) > 0 {
+		return class
+	}
+	var programTypes []models.ProgramCreditType
```
CK-7vn: This is small, and trivial, but should we move it out of the handler layer?

backend/src/models/program_classes.go
```diff
@@ -103,65 +212,82 @@ func (c *ProgramClass) AfterUpdate(tx *gorm.DB) (err error) {
...
 	UserID            uint                    `json:"user_id" gorm:"not null"`
 	EnrollmentStatus  ProgramEnrollmentStatus `json:"enrollment_status" gorm:"size:255" validate:"max=255"`
 	ChangeReason      string                  `json:"change_reason" gorm:"size:255" validate:"max=255"`
 	EnrolledAt        *time.Time              `json:"enrolled_at"`
 	EnrollmentEndedAt *time.Time              `json:"enrollment_ended_at"`
 
-	User  *User         `json:"user" gorm:"foreignKey:UserID;references:ID"`
-	Class *ProgramClass `json:"class" gorm:"foreignKey:ClassID;references:ID"`
+	User   *User               `json:"user" gorm:"foreignKey:UserID;references:ID"`
+	Cohort *ProgramClassCohort `json:"class" gorm:"foreignKey:CohortID;references:ID"`
```
CK-7vn: Should we change this json tag to cohort?

frontend/src/types/program.ts
CK-7vn: I think these should be changed from program_class prefix to class_ prefix.

backend/src/database/users.go
```diff
@@ -142,14 +142,15 @@ func (db *DB) GetTransferProgramConflicts(ctx context.Context, id uint, transfer
...
 		inner join users u on u.id = pce.user_id
 			and u.id = ?
-		inner join program_classes pc on pc.id = pce.class_id
+		inner join program_class_cohorts pc on pc.id = pce.cohort_id
 			and pc.facility_id = u.facility_id
+		inner join program_classes cl on cl.id = pc.class_id
 		inner join programs p on p.id = pc.program_id
 		where enrollment_status = 'Enrolled'
-			and pc.name not in (select name from transfer_facility_programs)`
+			and cl.name not in (select name from transfer_facility_programs)`
```
CK-7vn: This says resolved....but I think it still needs to be changed to p.name not in...

frontend/src/pages/programs/ClassManagementForm.tsx
```diff
@@ -540,6 +612,59 @@ export function ClassManagementFormInner({
...
+                    id="parent-class"
+                    className={`mt-1 w-full rounded-md border border-input bg-background px-3 py-2 text-sm ${focusRing}`}
+                    value={selectedClassId}
+                    onChange={(e) => {
+                        const value = e.target.value;
+                        setSelectedClassId(value);
+                    }}
+                >
+                    <option value="">Select a class…</option>
+                    {existingClasses.map((c) => (
```
CK-7vn: We should filter out archived classes here, just to make sure no one sees them. 
```ts
{existingClasses.filter((c) => !c.archived_at).map((c) => (
```
Or do it server side and accept &archived=false

backend/src/database/delete_guards.go
```diff
@@ -54,24 +54,24 @@ func (db *DB) ProgramBlockingChildren(programID int) (models.DeleteBlockingChild
...
 WITH cls AS (
-  SELECT id FROM program_classes
+  SELECT id FROM program_class_cohorts
    WHERE program_id = ? AND deleted_at IS NULL
 )
 SELECT
   (SELECT COUNT(*) FROM cls) AS classes,
   (SELECT COUNT(*) FROM program_class_enrollments
-     WHERE class_id IN (SELECT id FROM cls) AND deleted_at IS NULL) AS enrollments,
+     WHERE cohort_id IN (SELECT id FROM cls) AND deleted_at IS NULL) AS enrollments,
```
CK-7vn, line 68: I think this only counts a class as blocking if it still has a cohort under it (the cls CTE pulls from program_class_cohorts). If a class has zero cohorts, like one that got created but never had a cohort attached, would this miss it entirely? And I don't think DeleteProgram touches program_classes at all either. Could that leave an orphaned class row behind after a program delete?

backend/seeder/main.go
```diff
@@ -244,36 +256,53 @@ func seedTestData(db *gorm.DB) {
 		}
```
CK-7vn, line 256: Actually looking at this more, I think availableCohorts can pull in two sibling cohorts under the same class for one user (nothing dedupes by ClassID here), and if both land on Enrolled or EnrollmentCompleted we'd hit the new unique constraints on (user_id, class_id). It's only logged, not fatal, so make seed still "succeeds," just quietly drops some rows. Might be worth deduping by ClassID before we sample?

CK-7vn (changes requested):
Jeesh man, this was a serious serious PR, really good job on it. Few small things, nothing too serious. Nice work man, for real.

backend/src/database/program_classes.go
carddev81: @CK-7vn good catch!! I dug into this one and this was an existing error, program_classes doesn't have the instructor_id column here, it is within the program_class_events table.  There were many other issues that I noticed along the way:

image

backend/src/database/class_enrollments.go
carddev81: @CK-7vn amazing!!! Thank you so much for flagging this. Good eye. So many changes.

backend/src/services/classes.go
CK-7vn: Gotta update these I believe, should be CohortID.

backend/src/database/program_classes.go
```diff
@@ -205,14 +227,14 @@ func (db *DB) UpdateProgramClass(content *models.ProgramClass, id int, conflictR
...
-			Where("class_id = ?", existing.ID).
+			Where("cohort_id = ?", existing.ID).
 			Update("instructor_id", content.InstructorID).Error; err != nil {
 			trans.Rollback()
 			return nil, nil, newUpdateDBError(err, "program class event instructor")
 		}
 	}
 
-	if err := trans.Session(&gorm.Session{FullSaveAssociations: false}).Model(&models.ProgramClass{}).Where("id = ?", existing.ID).Updates(existing).Error; err != nil {
+	if err := trans.Session(&gorm.Session{FullSaveAssociations: false}).Model(&models.ProgramClassCohort{}).Where("id = ?", existing.ID).Updates(existing).Error; err != nil {
```
CK-7vn, line 237: I believe the AfterUpdate will require this to be a map[string]interface{} whenever status changes.

backend/src/database/reports.go
```diff
@@ -198,16 +199,17 @@ func (db *DB) GenerateProgramOutcomesReport(ctx context.Context, req *models.Rep
...
+			cl.name AS class_name,
 			pc.status AS status,
 			pc.credit_hours AS credit_hours,
 			pc.capacity AS capacity,
 			COUNT(CASE WHEN pce.enrollment_status = 'Enrolled' THEN 1 END) AS active_enrollments,
 			COUNT(CASE WHEN pce.enrolled_at BETWEEN ? AND ? THEN 1 END) AS range_enrollments
 		`, req.StartDate, req.EndDate).
-		Joins("LEFT JOIN program_class_enrollments pce ON pce.class_id = pc.id AND pce.deleted_at IS NULL").
+		Joins("JOIN program_classes cl ON cl.id = pc.class_id").
+		Joins("LEFT JOIN program_class_enrollments pce ON pce.cohort_id = pc.id AND pce.deleted_at IS NULL").
```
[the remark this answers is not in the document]
CK-7vn: This actually hasn't been fixed I don't think. This still selects pc.credit_hours directly with no coalesce.

backend/src/database/reports.go
```diff
@@ -198,16 +199,17 @@ func (db *DB) GenerateProgramOutcomesReport(ctx context.Context, req *models.Rep
 func (db *DB) GenerateProgramClassBreakdown(ctx context.Context, req *models.ReportGenerateRequest, programID, facilityID uint) ([]models.ProgramClassBreakdownRow, error) {
 	var rows []models.ProgramClassBreakdownRow
 
-	tx := db.WithContext(ctx).Table("program_classes pc").
+	tx := db.WithContext(ctx).Table("program_class_cohorts pc").
 		Select(`
-			pc.name AS class_name,
+			cl.name AS class_name,
 			pc.status AS status,
 			pc.credit_hours AS credit_hours,
```
CK-7vn: ** Reference on coderabbit comment ** 
I would just coalesce here, so that the credit_hours is being calculated correctly. 
```go
COALESCE(ps.credit_hours, cl.credit_hours) AS credit_hours,
```

CK-7vn (commented):
Just a couple more until your home free my friend! Still amazed at how good of a job you did on this, had to of been a serious pain in the butt. 

Also, not to be a pain, but would you mind fixing a bug completely unrelated to you? Just surfaced it during testing today, it should be in backend/src/handlers/login_flow.go:96-104: 
```go
redirect, err := getKratosRedirect(resp)
if err != nil {
    err := s.Db.UpdateFailedLogin(user.ID)   // := creates a NEW err, shadowing the outer one
    log.infof("Failed login attempt for %d at %s", user.ID, time.Now())
    if err != nil {
        log.error("error updating failed login attempts", err)
    }
    return NewServiceError(err, resp.StatusCode, "Invalid login")  // this "err" is the shadowed one (nil), not Kratos's real error
}
```
Just rename the inner error to like...updateErr or something, that way it doesn't overwrite the one we need. I just figured since you're already in the backend, if you wouldn't mind. No biggie if not just let me know and I'll make a note of it for later.

backend/src/database/program_classes.go
carddev81: this code was dead for awhile. it never fired the hook off. completely removed it and rewrote a piece for this specific scenario. Good catch.  Dead code removed.

CK-7vn (approved):
Nice work man! This was an absolutely insane PR. Great work.

PR #1211, opened 2026-08-19
fix: save Learning Record entries before finishing, fix double-counted question events, add facility tagging (ID-827)

corypride:
Branch: `cpride/posthog_lr_fixes`
Follows: PR #1207 (`cpride/posthog_learning_record`), merged as `b7512cd0`
Tickets: **[ID-827](https://app.asana.com/1/1201607307149189/project/1212887022738973/task/1217097620641814)** (filter by facility), plus fixes supporting **[ID-830](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1217564912951082)** (Learning Record insights)

## Summary

**A resident's work could be lost three ways, and none of them looked like failure.** Clicking
*Finish* navigated away immediately, which unmounted the form and cancelled the pending 500 ms
autosave, so anything typed in the final half-second never reached the server. A write that
*failed* resolved as success, so the resident saw "Saved" and Finish navigated home on an entry
that was never persisted. And the autosave and the Finish flush could race, with the loser's
newer answers dropped under the same green label.

Finish now saves first and leaves only once the save succeeds; a failed write surfaces as an
error and keeps the resident on the form; and writes are serialized so they cannot overwrite each
other.

Everything else is invisible to residents and staff:

- **Analytics correctness** — the Learning Record events were counting some questions twice, and
  the form's final step never reached the step funnel at all, plus three smaller issues raised in
  review of #1207 and merged without being addressed.
- **ID-827, filter by facility** — `facility_id` is a per-database autoincrement, so maine's
  facility 1 and St. Louis's facility 1 collapse into one row in any breakdown. Events now also
  carry `facility` (`maine:1`) and `facility_name`.
- **A new `lr_session_ended` event** so ID-830's "total time in the tool per session" graph can be
  built from a real measurement rather than a proxy.

> **Before building the per-question charts:** the Learning Record events maine has sent since
> Aug 18 count some questions twice. This PR stops that. Start those charts from the date this
> deploys — the affected data is about one test entry, so it isn't worth correcting.

No backend, schema, or API changes. The only non-frontend files in the diff are Go module
manifests from an unrelated security bump — see [Unrelated: Go dependency bump](#unrelated-go-dependency-bump).

## Where to see each change

| --- | --- | --- |
| `/learning-record-funnel/entry` | Finish saves before it closes | Type into the last field and click **Finish** immediately. Reload the entry — the final keystrokes are there. Previously they were gone. |
| `/learning-record-funnel/entry` | Finish fails safely | Stop the backend, then click **Finish**. It stays on the form and shows the save error rather than navigating home on an unsaved entry. |
| `/learning-record-funnel/entry` | Failed autosaves show an error | Stop the backend and type. The autosave label reads **error**, not "Saved". Previously a failed write reported success. |
| `/learning-record-funnel/entry` · home | Failed deletes keep the entry | Stop the backend, delete an entry. It stays, with an error toast. Previously the row vanished locally and returned on the next reload. |
| `/learning-record-funnel/entry` | One analytics event per question | No visible change. With a real PostHog key, a completed entry now emits exactly 10 `lr_question_left`, 3 `lr_step_completed`, 1 `lr_entry_completed` — the third step event is new, see below. |
| `/learning-record-funnel/entry` | Final step reaches the step funnel | No visible change, but **`lr_step_completed` data changes shape**. The event fires on *leaving* a step and the last step is never left, so `future` emitted none on any entry ever recorded. Entry completion now flushes it. A step the resident skipped flushes with `duration_seconds: 0`. |
| `/learning-record-funnel/entry` | New `lr_session_ended` | Open the form, save nothing, navigate away — one event with `entries_completed: 0`. Nothing captured abandoned sessions before. |
| Login / any event | Facility tagging (ID-827) | Events now carry `facility` = `<state>:<id>` and `facility_name` alongside the existing `facility_id`. |
| Any page that loads data | SWR JSON failures report | A 200 response with an empty or malformed body now emits `api_error` instead of failing silently to analytics. |
| `demov2` | Public demo stops reporting | The demo build now gets a placeholder PostHog key, which the frontend rejects. |

---

# Reviewer notes

## The duplicate-event bug

Two code paths emitted the same event for the same question. `noteEdit` emitted
`lr_question_left` for the previous field whenever the resident moved to a new one, and
`stepChanged` then re-emitted **every** field of that step — including the one already reported.

Walking the first step (`programName`, `completionDate`, `whatMadeYouFinish`):

| Action | Emitted, before this PR |
| --- | --- |
| type in `programName` | — |
| type in `completionDate` | `programName` · `touched: true` · real duration |
| click Next | `programName` **again** · `touched: true` · `duration_seconds: 0`, then `completionDate`, then `whatMadeYouFinish` |

The duplicate carried `duration_seconds: 0` because the timing pointer had already moved on, but
`touched` came from a set and was still `true`. That combination is what made it harmful rather
than merely noisy:

- **Median time per question** filters on `touched = true` specifically to keep zeros out. The
  duplicate put a zero inside the filter, dragging the median down for every question except the
  last one edited in each step.
- **Answer rate per question** divides answered events by total events. The denominator doubled
  for touched questions but not untouched ones, so the rate was distorted per question and varied
  with the order the resident happened to fill the form in.

**The fix is structural, not a patch.** `noteEdit` no longer emits anything — it only accumulates
timing. `stepChanged` and `entryCompleted` are now the only emitters, and `touched` is derived
inside the emitter rather than passed in by the caller, so a second event for an already-reported
question cannot be reintroduced by a future edit.

Timing also got more accurate on the way through. It was previously read off whichever editing run
happened to be open, so a question the resident returned to reported only its last run. It now sums
each question's editing runs across the whole entry, with each run ending at the last keystroke so
idle time after they stop typing isn't charged to the question.

Making `entryCompleted` an emitter also closed an older gap. `lr_step_completed` fires when a
resident *leaves* a step, and the last step is never left — so `future` emitted none on any entry
ever recorded, and step-level timing covered two thirds of the form. The same gap hit the middle
step from the other direction: the progress card is an ungated tablist and `activeStep` starts at 0,
so resuming a saved entry and jumping to the last step left `experience` unreported, dropping its
questions from the answer-rate denominator.

`entryCompleted` now flushes every step that hasn't reported. The existing `reportedSteps` guard
makes an already-reported step a no-op, so the flush is unconditional and cannot double count. A
completed entry emits **3** `lr_step_completed`, not 2. Duration is real for the step on screen and
`0` for one never visited — those zeros keep the denominator complete and are expected in
`lr_step_completed` timing.

Two smaller corrections in the same file:

- **`total_fields` was 10 against a 9-field completion requirement.** The local loop walked every
  field while `entryIsComplete` excludes the optional `completionDate`, so `answered_count` could
  not reach `total_fields` without a date — capping any completion-rate insight built on the pair
  at 90%. Both now come from the helpers `entryIsComplete` itself uses.
- **`beginSession` re-bases the session start and entry count**, not just the one-shot. Its
  cleanup always emits, so an earlier window has already been reported by the time it runs;
  carrying that window's start time or count forward double counted it into the next
  `lr_session_ended`. StrictMode walks exactly that path, and so would `isFunnel` flipping while
  the component stays mounted.

## Finish now saves before it completes

`validateFinishRequirements` is `async` and awaits `persistActiveRow()`. Order is validate
completeness → persist → record completion, so an entry that failed to save no longer produces a
completion event and no longer overstates the funnel.

This is also what fixes the data loss described in the summary — the same await covers both, since
the pending autosave is exactly what was being discarded.

Notes for review:

- `persistActiveRow` reads the live session and returns early when the row already matches what is
  committed, so this adds no redundant write on the common path.
- `FunnelFinishHandlers.validateFinishRequirements` now returns `Promise<boolean>`.
  `funnelOnFinish` keeps its `() => void` type and is adapted at the call site with
  `() => void handleFinish()`, matching the existing `onClick={() => void handleDownload()}`
  convention one line above — so the Shell and the button component are untouched.
- The await opens a window for a second click, so `handleFinish` is guarded by an in-flight ref.
  The tracker separately refuses a duplicate completion event; the ref is what prevents a double
  navigation.
- No new UI. The existing autosave indicator already renders `saving` / `saved` / `error`, and
  finishing now drives it.

## Saving now fails loudly

**A failed write was reported as success.** `API.fetchWithHandling` resolves `{ success: false }`
rather than throwing, so `apiCreateEntry` returned null and `apiUpdateEntry` returned false on a
500, a timeout, or offline — and `upsertCommittedEntry` awaited them and resolved normally anyway.
Callers read that as success. `upsertCommittedEntry` and `deleteCommittedEntry` now reject, and
every caller handles it.

**Concurrent saves could drop answers.** The debounced autosave and the Finish flush overlap by
design — the timer's save can still be on the wire when Finish fires. `upsertCommittedEntry`
chooses POST vs PUT from a `client_id → backend id` map written only *after* a create resolves, so
two racing calls for a row with no id yet both POST, and the unique index on
`(user_id, client_id)` rejects the loser. That surfaced as "Saved" with the loser's newer answers
gone. `persistActiveRow` now serializes attempts through a queue; `writeActiveRow` holds the write
itself and reads the live session, so a caller that waited writes what the resident has typed by
then rather than what was on screen when it queued.

Notes for review:

- **The autosave label could lie.** A superseded save resolving late repainted `saved` over a
  newer `saving`, and an in-flight save could repaint `saved` over the `pending` of an edit made
  while it was on the wire. Save intents now claim a ticket at schedule time and report an outcome
  only while they still hold the latest.
- **Finish cancels the pending debounce** before flushing, since the flush writes the same row.
  Deliberately *after* the completeness check, so an incomplete row keeps its autosave and partial
  answers still reach the server.
- **A failed delete used to remove the row locally**, where it reappeared on the next hydrate —
  data loss in reverse. Local state now clears only once the server confirms.
- `lastWrittenRef` records confirmed writes so a queued attempt can skip a redundant PUT.
  `entries` alone can't do this: a queued attempt resumes on the microtask queue, ahead of the
  re-render carrying the previous attempt's `setEntries`, so it would always read stale.

## ID-827 — filter by facility

The ticket asks for "a facility flag for being able to filter insights by not only state deployments
but also state and the facility within." Facility filtering technically worked already — `facility_id`
is on every event — but **breakdowns were wrong**, which is the part that matters.

`facilities.id` is a `SERIAL`
([00002_create_init_tables.sql:12-19](backend/migrations/00002_create_init_tables.sql#L12-L19)), a
per-database autoincrement. Each state's database starts at 1, so `facility_id: 1` names a different
facility depending on which deployment sent it, and a breakdown merges them into one row. This is
the same collision that already forced `distinct_id` to be namespaced to `maine:78`.

Two more schema facts shaped the fix: `facilities.name` has **no unique constraint** and is editable
from the admin UI (so it can collide across states, and a rename splits history), and there is **no
slug or external id column** — the only durable unique key is the pair (deployment, id).

So `identifyUser` now sends three facility properties instead of one, built as a single object so
the person and event copies cannot drift:

| Property | Role |
| --- | --- |
| `facility` (new) | `maine:1` — the stable key to filter and break down on. Survives renames, never collides across states. |
| `facility_name` (new) | `BCF` — display label, so dashboard rows read as names rather than opaque ids. |
| `facility_id` (unchanged) | Kept so the existing pilot dashboard filter keeps working. |

Notes for review:

- **`facility` only exists on events sent after this deploys.** Nothing backfills, so a `facility`
  breakdown starts at the deploy date.
- **Pre-login events carry no facility at all** — the login screen's `$pageview`, an `api_error`
  from an unauthenticated request. Correct, since there is no facility yet, but it means a
  facility-scoped total reads lower than an unscoped one.
- **`facility_name` names a building, not a person.** Worth stating plainly given the context: it is
  an institution name, not resident data.
- `resetAnalytics` needed no change — `posthog.reset()` clears super properties and only the
  deployment tags are re-registered, so facility correctly disappears on logout and returns at the
  next identify. Facility is user-scoped, unlike `deployment`/`state`.

## New event: `lr_session_ended`

ID-830 asks for "total time spent in the tool per session." Nothing measured it:
`seconds_since_session_start` only rides on `lr_entry_completed`, so a resident who opened the form
and abandoned it contributed **nothing** — and abandonment is exactly what a friction study is
looking for. The alternative, PostHog's `$session_duration`, is the whole-app session on a
30-minute inactivity timeout, so it counts time spent anywhere else in the app.

The form already recorded `sessionStartMs` at mount; it simply never reported when the session
ended. `lr_session_ended` now fires once when the resident leaves, carrying `duration_seconds` and
`entries_completed`.

- **Both routes out are covered**: `pagehide` for tab/browser close and bfcache, the effect cleanup
  for in-app navigation, which is how Finish leaves. A one-shot in the tracker makes the overlap
  harmless. `beforeunload`/`unload` are deliberately not used — unreliable, and they break bfcache.
- **It also improves the "achievements per session" graph**, which can now break down on
  `entries_completed` — a real per-session total including a **zero bucket** for sessions that saved
  nothing, rather than being inferred from `entry_index_in_session`.
- **Dev-only wrinkle**: React StrictMode double-invokes effects, so the throwaway cleanup emits one
  ~0-second event before the real one. `beginSession()` re-arms the one-shot so the real event still
  fires. Production runs the effect once. Filter `duration_seconds > 0` if dev noise gets in the way.
- Both properties stay inside the module's privacy rule: a duration and a count, no content.

## Smaller fixes from the #1207 review

- **Secrets no longer interpolated into shell source.** `POSTHOG_KEY` / `POSTHOG_HOST` move to the
  build step's `env:` block. Rendered directly into the `run:` script, a value containing shell
  syntax would execute on the runner.
- **The demo placeholder key now exists.** The comment in `container_builds.yml` claimed builds
  that shouldn't report get a placeholder key — no such branch existed, so `demov2` was reporting
  as `deployment=development, state=demo` on the real key. Three documents already stated demo was
  silent; this makes that true rather than rewriting the docs to match the leak. The primary
  dashboard filter (`deployment equals production`) already excluded demo, but the documented
  alternative (`state is not in staging, unknown`) would have let seeded demo data into the pilot
  numbers.
- **SWR JSON decode failures report.** `res.json()` sat outside both the network and HTTP error
  handlers, so an OK response with a bad body reached the user as an error while analytics saw
  nothing — the exact gap `swrFetcher.ts` was added to close.
- **`vite-env.d.ts` documents `local`.** `.env.example` tells developers to set
  `VITE_DEPLOYMENT=local`; the type declaration listed only `production` and `development`.

#### First — make the app report at all

A normal dev checkout sends **nothing**. `initAnalytics` returns early unless the key is
`phc_`-prefixed *and* a deployment is named, so every check below is silently a no-op until
`frontend/.env` has all four of these:

```
VITE_PUBLIC_POSTHOG_KEY=phc_...            # a real key; the .env.example sample is rejected
VITE_PUBLIC_POSTHOG_HOST=https://us.i.posthog.com
VITE_DEPLOYMENT=local
VITE_STATE=maine                            # without this, stateTag() falls back to `unknown`
```

That last line matters for the facility check specifically: leave `VITE_STATE` blank and the
events read `facility: unknown:<id>`, which looks like a bug but is the documented fallback.

The Learning Record checks also need the `learning_record` feature flag **on** — the form is
behind it, and the `lr_*` events only exist inside the form.

Use the dev tools and posthog activity chart to see events.

## Unrelated: Go dependency bump

`d3be631a` bumps `klauspost/compress` to v1.18.7 across all four Go modules to clear
**GO-2026-5841**. `go.mod` / `go.sum` only — no Go source changes, which is why the summary above
still says no backend changes:

```
backend/go.mod              backend/migrations/go.mod    backend/seeder/go.mod
backend/go.sum              backend/migrations/go.sum    backend/seeder/go.sum
provider-middleware/go.mod  provider-middleware/go.sum
```

It rides along because it was already on the branch. It is independent of the Learning Record
work and can be reviewed on its own — the *Scan Go Dependencies for Vulnerabilities* and both
`golangci-lint` checks cover it.

## Not in this PR

- **The consent question from #1207 is still open.** It asked reviewers to decide whether
  person-attributed `lr_*` events and `char_count` match resident copy promising "all identifying
  information removed." The only review was an approval with no comment on it, so nothing was
  decided and both still ship. Worth settling before the pilot quotes any of this data.

### DIFF
--- a/frontend/src/api/swrFetcher.ts
+++ b/frontend/src/api/swrFetcher.ts
@@ -43,5 +43,18 @@ export const swrFetcher = async (url: string): Promise<unknown> => {
         });
         throw new Error(res.statusText);
     }
-    return res.json() as Promise<unknown>;
+    try {
+        return (await res.json()) as unknown;
+    } catch (err) {
+        // An OK response with an empty or malformed body still fails the page.
+        // Without this the request looks successful to analytics while SWR
+        // surfaces an error to the user.
+        captureEvent(ANALYTICS_EVENTS.ApiError, {
+            status: res.status,
+            method: 'GET',
+            ...analyticsUrl(url),
+            message: err instanceof Error ? err.message : 'Invalid JSON'
+        });
+        throw err;
+    }
 };
--- a/frontend/src/hooks/useTranscriptDraft.ts
+++ b/frontend/src/hooks/useTranscriptDraft.ts
@@ -105,6 +105,19 @@ export function useTranscriptDraft() {
         return fresh;
     }, []);
 
+    /**
+     * Commit an entry to the server, creating or updating based on whether this
+     * client_id has a backend id yet.
+     *
+     * Rejects when the write fails. `API.fetchWithHandling` resolves
+     * `{ success: false }` rather than throwing, so `apiCreateEntry` returns null
+     * and `apiUpdateEntry` returns false on a 500, a timeout, or offline — which
+     * left this function resolving normally on a failed write. Callers read that as
+     * success: the resident saw "Saved", `lr_entry_completed` fired, and Finish
+     * navigated home with the entry never persisted. This is the only place that
+     * can turn a failed write back into a signal, so it throws and lets callers
+     * decide.
+     */
     const upsertCommittedEntry = useCallback(async (entry: TranscriptEntry) => {
         const trimmedEntry = {
             ...entry,
@@ -112,32 +125,46 @@ export function useTranscriptDraft() {
         };
         const backendId = entryBackendIds.current.get(entry.id);
         if (backendId !== undefined) {
-            await apiUpdateEntry(backendId, trimmedEntry);
+            if (!(await apiUpdateEntry(backendId, trimmedEntry))) {
+                throw new Error('failed to update learning record entry');
+            }
             setEntries((prev) =>
                 prev.map((e) => (e.id === entry.id ? trimmedEntry : e))
             );
         } else {
             const result = await apiCreateEntry(trimmedEntry);
-            if (result) {
-                entryBackendIds.current.set(entry.id, result.backendId);
-                setEntries((prev) => {
-                    const idx = prev.findIndex((e) => e.id === entry.id);
-                    if (idx >= 0) {
-                        const next = [...prev];
-                        next[idx] = result.entry;
-                        return next;
-                    }
-                    return [...prev, result.entry];
-                });
+            if (!result) {
+                throw new Error('failed to create learning record entry');
             }
+            entryBackendIds.current.set(entry.id, result.backendId);
+            setEntries((prev) => {
+                const idx = prev.findIndex((e) => e.id === entry.id);
+                if (idx >= 0) {
+                    const next = [...prev];
+                    next[idx] = result.entry;
+                    return next;
+                }
+                return [...prev, result.entry];
+            });
         }
         dispatchEntrySessionUpdated();
     }, []);
 
+    /**
+     * Remove an entry, server-side first.
+     *
+     * Rejects when the delete fails, for the same reason `upsertCommittedEntry`
+     * does — `apiDeleteEntry` returns false rather than throwing. Previously the
+     * local removal ran unconditionally, so a failed delete made the row disappear
+     * while it still existed on the server, and it reappeared on the next hydrate.
+     * Dropping local state only after the server confirms keeps the two in step.
+     */
     const deleteCommittedEntry = useCallback(async (id: string) => {
         const backendId = entryBackendIds.current.get(id);
         if (backendId !== undefined) {
-            await apiDeleteEntry(backendId);
+            if (!(await apiDeleteEntry(backendId))) {
+                throw new Error('failed to delete learning record entry');
+            }
             entryBackendIds.current.delete(id);
         }
         setEntries((prev) => prev.filter((e) => e.id !== id));
--- a/frontend/src/lib/events.ts
+++ b/frontend/src/lib/events.ts
@@ -28,6 +28,9 @@ export const ANALYTICS_EVENTS = {
     LrQuestionLeft: 'lr_question_left',
     LrStepCompleted: 'lr_step_completed',
     LrEntryCompleted: 'lr_entry_completed',
+    // Emitted when the resident leaves the form. The only measure of total time
+    // in the tool that also counts residents who never saved an entry.
+    LrSessionEnded: 'lr_session_ended',
     // Errors & friction
     ApiError: 'api_error'
 } as const;
@@ -70,6 +73,19 @@ function stateTag(): string {
  */
 function analyticsDistinctId(userId: number): string {
     return `${stateTag()}:${userId}`;
+}
+
+/**
+ * Namespace a facility to its deployment (ID-827).
+ *
+ * `facilities.id` is a per-database SERIAL, exactly like `users.id`, so a bare
+ * id names a different facility in every state — breaking any insight down by
+ * `facility_id` merges maine's facility 1 with stlouis's into one row. This is
+ * the key to filter and break down on; it survives facility renames, which
+ * `facility_name` does not.
+ */
+function facilityTag(facilityId: number): string {
+    return `${stateTag()}:${facilityId}`;
 }
 
 /**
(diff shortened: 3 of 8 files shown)


carddev81 (approved):
good stuff! good job!!

PR #1214, opened 2026-08-22
fix: instrumentation for learning record acceptance criteria

corypride:

## Description of the change

Six acceptance criteria needed behavioural data from PostHog. Four of them could not be answered
from what the app was sending, and two of those would have reported numbers that were **wrong**
rather than missing — which is the worse failure, because nothing on the dashboard would have
looked broken.

No new event names. The same five `lr_*` events, with properties added and four defects fixed.

Companion doc: [What we are sending](https://docs.google.com/document/d/1nGSmn7CKhHDTiF3zYI6htkK9S0NrtGC0M0Q7iRcBqvY/edit?tab=t.0) — the payload reference,
the definition of a *sitting*, and a table mapping each acceptance criterion to the events and
properties that answer it.

### Criterion 6 could not be answered at all

*"Total time spent in the tool per session (distribution)"*

The `lr_session_ended` window lived inside `DigitalTranscriptWysiwygEntry`, which unmounts on every
list ↔ entry navigation (`DigitalTranscriptHome.tsx:352`, `DigitalTranscriptEntryPage.tsx:133-155`).
So one sitting emitted several partial events with nothing on the payload to rejoin them,
`entry_index_in_session` restarted at 1 on each visit, and `seconds_since_session_start` measured
from the mount rather than from the start of the sitting. Three symptoms, one cause.

The session now lives in `learningRecordSession.ts`, a module-level singleton that outlives any
mount. `leaveForm` no longer emits: it stops active time accruing and arms a 30-minute resume
window, so a trip out to the list keeps one sitting intact. `endSession` fires once per sitting —
on `pagehide`, on the window lapsing, or on recovery of a sitting a reload interrupted.

`active_seconds` accrues only while the form is mounted, visible and not idle, so it excludes
background-tab and list-page time. The idle cutoff is **five minutes**, deliberately generous:
`lr_question_left` already discards think time, so a short cutoff here would leave nothing in the
dataset showing hesitation — the exact friction signal the story is built on.

### Criterion 5 was double counting

*"Distribution of number of achievements entered per resident, to see if residents stop at one or
add more"*

`this.completed` guards one `entryStarted` → `entryCompleted` pass, and the tracker is rebuilt on
remount. So re-opening a saved achievement and pressing Finish again emitted a second
`lr_entry_completed` with nothing to tell the two apart — an *edit* read as a new achievement.
That can invert the exact question the criterion exists to answer.

Completions now dedupe against a list of achievement ids in `localStorage`. Nothing new goes on the
wire: the ids never leave the browser.

A rejected re-Finish skips the completion flush, but the resident is still in the form, so leaving
afterwards still emits that step's questions flagged `abandoned: true` (verified in the browser).
Criterion 5 is fully protected — one achievement, one `lr_entry_completed`. Criterion 3 must use
**Unique users** rather than event counts, which the build-order doc already specifies, so a
resident revisiting their own entry cannot inflate a selection rate.

### Criterion 3 was missing the population it is defined over

*"Percent of residents who select at least one option vs. leave it blank"*

Questions only reported when a step was left *for another one* or an entry completed. Nothing
flushed the step the resident was standing on when they gave up — so someone who left a chip
question blank **and** abandoned emitted nothing at all, and the blank-leavers were missing from
the denominator the criterion is defined over.

`abandonedCurrentStep` now flushes that step from the effect cleanup. Only the current step:
unvisited ones would arrive as rows of zeros and invent a denominator nobody saw. Those rows carry
`abandoned: true`, so timing tiles can exclude what answer-rate tiles need to keep.

### Criterion 1 measured typing span, not time on question

*"Average and distribution of time spent per question/step, to surface which questions take
unusually long"*

Per-question `duration_seconds` runs first keystroke to last keystroke. A question someone stared
at for a minute and answered in ten seconds reported ten — so the metric was blind to exactly the
questions it was meant to surface, because hesitation is where confusion shows.

New `seconds_before_first_edit`: the quiet gap between the previous keystroke anywhere in the step
(or the step opening, for the first question engaged) and the first keystroke here. Derived from
state the tracker already held — no focus handlers, no coupling to form markup.

True per-question dwell time would need focus/blur across ~10 field components, which this module
deliberately avoids so it does not depend on the form's markup. Hesitation and typing span as two
separate numbers is more useful than one number silently conflating them.

### Criterion 2 measured only the final visit

*"Average time to complete a single achievement entry"*

`entryStarted` re-bases its timer and fires again when a resident returns to the same achievement,
so entry duration reported the last visit alone and systematically understated the work — and it is
the number the facilitator cross-reference leans on.

New `cumulative_seconds`: editing time banked per achievement and summed across every visit,
excluding time away. Additive, so `duration_seconds` keeps its meaning and there is no second
cutover to explain.

### Also fixed, found while verifying the above

- **bfcache restore left the sitting dead.** `pagehide` reports and tears down; the page returns
  with the editor still mounted, so `enterForm` never runs again and everything the resident did
  afterwards was unattributed. A `pageshow` listener that outlives teardown starts a new sitting.
- **A second `leaveForm` re-stamped the close time**, inflating `duration_seconds` 60s → 660s in
  the probe that caught it.
- **Learning Record storage is per-origin and logout clears none of it**, so the next resident on a
  shared workstation was offered the previous one's unfinished achievement *by name* — on the page
  whose own notice promises the opposite — and could silently resume their sitting. Records are now
  stamped with a SHA-256 digest of the resident's analytics id (never the id itself, so nothing
  readable is left in `localStorage` after logout) and discarded when someone else signs in. That
  also covers a browser closed without logging out at all, which clearing-on-logout cannot, and
  keeps a returning resident's own work. Records carrying no stamp are treated as someone else's.
- **Banked time for an already-counted entry was never released.** `entryCompleted`'s early return
  for a re-Finished achievement skipped `takeEntryActiveMs`, so that entry's `entryTiming` record
  kept growing on every later re-open and abandon with nothing left to ever read or clear it —
  unbounded storage growth on exactly the achievements residents revisit most. Now cleared on that
  same early-return path.
- **`cumulative_seconds` had no idle cutoff.** Both contributions — the visit banked in
  `abandonedCurrentStep` and the final visit added in `entryCompleted` — were raw wall clock with
  nothing capping an unattended tab, so a resident who left an achievement open overnight could
  report as having spent hours completing it, skewing the criterion-2 average. Now capped at the
  same 5-minute idle cutoff as `active_seconds`.
- **`lr_entry_started` wasn't deduped, only `lr_entry_completed` was.** `entryStarted` emitted, and
  bumped `lr_session_ended.entries_started`, on every open of an achievement — including reopening
  one already completed to edit it — while `entryCompleted` and its `entries_completed` counter
  correctly skip a re-Finish. A resident who reopens one saved achievement twice reported 3 started
  against 1 completed on the same session event. `entryStarted` now skips both the counter bump and
  the emit for an already-completed row id, so the two counters describe the same population.
- **The sitting's `sessionStorage` key could drift mid-sitting.** `storageKey()` resolved which
  prototype variant's key to use fresh on every call, from the same mutable route global the rest
  of the module reads — but the sitting itself is a long-lived singleton that survives navigation
  by design. A prototype switch mid-sitting silently retargeted later writes (and the clear in
  `endSession`) at the other variant's key, orphaning the record under the original one, which a
  later visit to that route could recover and report a second time. The key is now resolved once,
  when the sitting opens, and held until `endSession` clears it.
- **A bfcache restore left the analytics tracker numbered against the sitting that had just ended.**
  The bfcache fix above deliberately starts a new sitting on `pageshow` without the editor
  unmounting — but `LearningRecordTracker` lives in that same component's `useRef` and is likewise
  never recreated, and `entryStarted` (the only place that numbers `entryIndex`) is gated to run
  once per row id, so it never re-ran either. An achievement left open across the freeze reported
  `lr_entry_completed` with `entry_index_in_session` from the *old* sitting alongside
  `seconds_since_session_start` from the *new* one — breaking criterion 4's "first entry of the
  sitting" identification, and leaving the new sitting's `entries_started`/`entries_completed` pair
  asymmetric if that entry went on to complete. `learningRecordSession.ts` now dispatches a
  `learning-record-session-restarted` event on restart (the same `window.dispatchEvent`
  convention `transcriptEntrySessionStorage.ts` already uses for this class of problem), and the
  tracker's new `reattachToSession` renumbers `entryIndex` and reports a fresh `lr_entry_started`
  without touching question/step timing already accumulated.
- **`noteEntryStarted` numbered entries by completions, not starts.** It incremented
  `entriesStarted` but returned `entriesCompleted + 1` as `entryIndex` — the two counters only
  happened to agree when nothing had ever been abandoned. A resident who opened and abandoned two
  achievements before ever finishing one had both report `entry_index_in_session: 1`, since neither
  had completed. That breaks criterion 4's "first entry of the sitting" identification the same way
  the two bfcache fixes above do, and it fed directly into `reattachToSession`, which calls this
  same function. Now returns `entriesStarted` itself, and `0` (not `1`) when no sitting is active,
  so a missing session is distinguishable from a real first entry — `LearningRecordTracker` already
  treats `entryIndex === 0` as its own "nothing open" sentinel.
- **The owner-change purge missed three retired key sets.** `allDigitalTranscriptStorageKeys()` only
  enumerated the current `FUNNEL_STORAGE_KEYS`/`CATEGORIES_STORAGE_KEYS`, so
  `DIGITAL_TRANSCRIPT_LEGACY_STORAGE` (pre–variant-split), `DIGITAL_TRANSCRIPT_B_STORAGE` (the
  removed guided-flow variant), and `STALE_TRANSCRIPT_A_STORAGE_KEYS` were never cleared when a new
  resident signed in — despite `purgeDigitalTranscriptStorage`'s own comment claiming to drop
  "every" record. `migrateDigitalTranscriptLegacyStorage()`, the only other code that touches those
  keys, has no call site anywhere in the app, so data written under any of them had no path to ever
  being cleared: on a shared workstation it would outlive every owner change, which is exactly the
  leak this PR's owner stamp exists to close. All three are now included in the purge.
- **A `localStorage` failure could skip the matching `sessionStorage` removal for the same key.**
  `purgeDigitalTranscriptStorage()`'s loop wrapped both stores' `removeItem` in one `try`, so if
  `localStorage.removeItem` threw (quota, private mode, sandboxed access), the `sessionStorage`
  removal for that same key never ran — and `session`, the previous resident's sitting, is the one
  key actually held in `sessionStorage`. A partial failure could leave that sitting in place while
  everything else was cleared. Each store now gets its own `try`/`catch` per key.
- **The two dashboard CTAs were raw `<a href>`**, i.e. full document loads, which fire `pagehide`
  and flush the sitting before the new document can resume it. Now `<Link to>`.

## Changes

- **New `learningRecordSession.ts`** — the sitting: resume window, reload recovery, active-time
  accumulation, single emit. Listeners (`visibilitychange`, `pagehide`, `pageshow`, activity) are
  registered by the module, not the component: registered in the effect they would be removed the
  moment the editor unmounted, so a resident who returned to the list and *then* closed the tab
  would produce no event at all now that leaving the form no longer emits.
- **`learningRecordAnalytics.ts`** — session fields removed; completion dedupe; per-entry time
  banking; `abandonedCurrentStep`; `abandoned` and `seconds_before_first_edit`.
- **`DigitalTranscriptWysiwygEntry.tsx`** — `enterForm`/`leaveForm`, the abandon flush, and the row
  id threaded into `entryStarted`/`entryCompleted`.
- **`ResidentHome.tsx`** — the two CTAs converted to `<Link to>`. The file already navigated
  SPA-style at `:184`/`:356` and reserves `href` for external links (`:159`), so the anchors read as
  an oversight rather than a decision.
- **`AuthProvider.tsx`** — the storage owner check, called (and awaited, since the digest hash is
  computed asynchronously via Web Crypto) before anything reads Learning Record storage.
- **`types/digital-transcript.ts`** — `session`, `countedEntries` and `entryTiming` keys, the
  version constant, and the owner stamp (a SHA-256 digest of the analytics id, not the id itself).
  Both prototype namespaces are purged explicitly:
  `getDigitalTranscriptStorageKeys()` resolves against the *current route*, and at sign-in that is
  `/home`, so it would only ever return the funnel keys and would quietly leave the categories ones
  behind.
- **`lib/events.ts`** — `LrSessionEnded` comment, and `analyticsDistinctId` exported so the owner
  stamp and the PostHog identity share one definition.

**The sitting record uses `sessionStorage`, not `localStorage` like its neighbouring transcript
keys.** Draft content should follow a resident across tabs; a clock must not, or a second tab
resumes the first tab's sitting and both emit. The dedupe and timing records *do* use `localStorage`
deliberately — editing an old achievement in another tab must not be counted twice either.

### Which events answer which acceptance criterion

| Acceptance criterion | Read from | Watch out for |
|---|---|---|
| 1. Time per question/step, and which take unusually long | `lr_question_left`: `question_key`, `duration_seconds`, `seconds_before_first_edit`, `touched`. `lr_step_completed`: `step_id`, `step_index`, `duration_seconds` | A question's `duration_seconds` is a typing span and excludes think time; `seconds_before_first_edit` is the hesitation before engaging it, and is null for questions never typed in. Filter `touched = true` and `abandoned = false` for clean timings. Step rows are honest wall clock. |
| 2. Average time to complete a single achievement entry | `lr_entry_completed`: `cumulative_seconds` (use this), `duration_seconds` | `duration_seconds` covers only the resident's last visit to that achievement; `cumulative_seconds` sums every visit, capped at 5 minutes of idle time per visit so an unattended tab can't inflate it. Compare facilitator estimates against `cumulative_seconds`. |
| 3. Selection rate per pill/checkbox question, and average options selected | `lr_question_left`: `question_key`, `kind = tags`, `answered`, `selection_count` | Count residents with Unique users, not events. Keep `abandoned = true` rows in the denominator: that is the "left it blank" population the criterion asks for. Option ceilings differ by question. |
| 4. Distribution of time-to-first-entry across residents | `lr_entry_completed`: `seconds_since_session_start`, `entry_index_in_session` | `entry_index_in_session = 1` is the first entry of a sitting, so a resident with several sittings contributes several rows. For each resident's genuine first, take `min(timestamp)` per person in SQL. |
| 5. Achievements per resident — do they stop at one or add more | `lr_entry_completed`, grouped by person | One event per achievement: re-finishing an existing one is discarded in the browser, so counts are achievements and not Finish presses. Grouping per resident needs SQL; Trends cannot count distinct on a property. |
| 6. Total time in the tool per session | `lr_session_ended`: `active_seconds` (use this), `duration_seconds`, `form_visits` | `active_seconds` excludes hidden-tab, idle and list-page time. `duration_seconds` spans the whole sitting including all three. See "What a sitting is". |

What rides along on every event (lr_session_ended payload)

| Prop | Status | Meaning |
|---|---|---|
| `duration_seconds` | scope changed | Wall clock, first form open → last form close. |
| `entries_completed` | scope changed | Counts across the sitting. |
| `active_seconds` | new | Mounted, visible, not idle. Criterion 6's "time spent". |
| `form_visits` | new | Times the form was opened this sitting. > 1 proves stitching. |
| `entries_started` | new | Pairs with `entries_completed` for abandonment. Deduped the same way `entries_completed` is: reopening an already-completed achievement bumps neither. |
| `ended_reason` | new | `navigated_away` \| `page_hidden` \| `recovered`. |

Also new: cumulative_seconds on lr_entry_completed; seconds_before_first_edit and abandoned on lr_question_left; abandoned on lr_step_completed. Every value is a count, a whole-second duration, a boolean or a fixed enum — within the privacy contract at learningRecordAnalytics.ts and the resident consent copy at learningRecordResidentCopy.ts:8,17. No session id and no row id are ever sent: all stitching and deduping happens on the client.
## Known limits

- **Entry from outside the app starts a new sitting** — bookmark, typed URL, arrival from another
  site. Those should be new sittings, so this is a boundary rather than a defect.
- **`duration_seconds` straddles the cutover** — still wall clock, but a sitting rather than a
  mount, so later values are systematically larger. Prefer `active_seconds`.
- **A browser kill defers the event** to the next visit, so `recovered` events land in the day they
  were sent rather than the day they happened.
- **The 5-minute idle cutoff is a judgement call.** Applied to both `active_seconds` and
  `cumulative_seconds`. It preserves ordinary thinking pauses and caps unattended screens, but six
  minutes of genuine deliberation loses the tail. Treat both as a conservative floor on effort.
- **Per-question dwell time is still unavailable** — see criterion 1 above. `seconds_before_first_edit`
  is a hesitation proxy, not dwell.
- **Adjacent flows remain uninstrumented**: the list page itself, PDF export, delete/cancel, table
  sort, and the categories prototype.

---
  - https://app.asana.com/0/0/1217564912951082

### DIFF
--- a/frontend/src/auth/AuthProvider.tsx
+++ b/frontend/src/auth/AuthProvider.tsx
@@ -3,7 +3,8 @@ import { INIT_KRATOS_LOGIN_FLOW, User } from '@/types';
 import { AuthContext, fetchUser, handleLogout } from '@/auth/useAuth';
 import { useNavigate } from 'react-router-dom';
 import { tabSessionManager } from '@/session/tabSession';
-import { identifyUser } from '@/lib/events';
+import { analyticsDistinctId, identifyUser } from '@/lib/events';
+import { setDigitalTranscriptStorageOwner } from '@/types/digital-transcript';
 
 export const AuthProvider: React.FC<{ children: React.ReactNode }> = ({
     children
@@ -26,6 +27,14 @@ export const AuthProvider: React.FC<{ children: React.ReactNode }> = ({
                     await handleLogout();
                     return;
                 }
+                // Before anything reads Learning Record storage: discard it if
+                // it belongs to a different resident. Facility workstations are
+                // shared, the keys are per-origin, and logging out clears none of
+                // them — so without this the next person to sign in inherits the
+                // previous one's unfinished achievement and their sitting.
+                setDigitalTranscriptStorageOwner(
+                    analyticsDistinctId(authUser.id)
+                );
                 identifyUser(authUser);
                 setUser(authUser);
             }
--- a/frontend/src/auth/AuthProvider.tsx
+++ b/frontend/src/auth/AuthProvider.tsx
@@ -32,7 +32,7 @@ export const AuthProvider: React.FC<{ children: React.ReactNode }> = ({
                 // shared, the keys are per-origin, and logging out clears none of
                 // them — so without this the next person to sign in inherits the
                 // previous one's unfinished achievement and their sitting.
-                setDigitalTranscriptStorageOwner(
+                await setDigitalTranscriptStorageOwner(
                     analyticsDistinctId(authUser.id)
                 );
                 identifyUser(authUser);
--- a/frontend/src/lib/events.ts
+++ b/frontend/src/lib/events.ts
@@ -28,8 +28,12 @@ export const ANALYTICS_EVENTS = {
     LrQuestionLeft: 'lr_question_left',
     LrStepCompleted: 'lr_step_completed',
     LrEntryCompleted: 'lr_entry_completed',
-    // Emitted when the resident leaves the form. The only measure of total time
-    // in the tool that also counts residents who never saved an entry.
+    // Emitted once per sitting, not per form visit: the session survives the trip
+    // out to the list page and ends on pagehide, on a lapsed 30-minute resume
+    // window, or when a later visit recovers one a reload interrupted. Owned by
+    // learningRecordSession.ts, not by the tracker, because the tracker dies with
+    // the editor component. Still the only measure of time in the tool that also
+    // counts residents who never saved an entry.
     LrSessionEnded: 'lr_session_ended',
     // Errors & friction
     ApiError: 'api_error'
@@ -71,7 +75,7 @@ function stateTag(): string {
  * last-write-wins between facilities, and would undercount weekly-active-users —
  * the pilot's headline metric — by collapsing distinct people into one.
  */
-function analyticsDistinctId(userId: number): string {
+export function analyticsDistinctId(userId: number): string {
     return `${stateTag()}:${userId}`;
 }
 
--- a/frontend/src/pages/student/ResidentHome.tsx
+++ b/frontend/src/pages/student/ResidentHome.tsx
@@ -1,4 +1,9 @@
-import { useLoaderData, useNavigate, useSearchParams } from 'react-router-dom';
+import {
+    Link,
+    useLoaderData,
+    useNavigate,
+    useSearchParams
+} from 'react-router-dom';
 import useSWR from 'swr';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import {
@@ -582,11 +587,11 @@ export default function ResidentHome() {
                                             learningRecordOutlineButtonClassName
                                         }
                                     >
-                                        <a
-                                            href={`${DIGITAL_TRANSCRIPT_ENTRY_PATH}?intent=new`}
+                                        <Link
+                                            to={`${DIGITAL_TRANSCRIPT_ENTRY_PATH}?intent=new`}
                                         >
                                             Log achievements
-                                        </a>
+                                        </Link>
                                     </Button>
                                     <Button
                                         type="button"
@@ -654,8 +659,8 @@ export default function ResidentHome() {
                                                 }
                                                 className="gap-1.5 bg-white px-4 text-[#556830] hover:bg-white/90"
                                             >
-                                                <a
-                                                    href={`${DIGITAL_TRANSCRIPT_ENTRY_PATH}?intent=new`}
+                                                <Link
+                                                    to={`${DIGITAL_TRANSCRIPT_ENTRY_PATH}?intent=new`}
                                                 >
                                                     {!heroIsFirstTime ? (
                                                         <Plus
@@ -664,7 +669,7 @@ export default function ResidentHome() {
                                                         />
                                                     ) : null}
                                                     {heroCtaLabel}
-                                                </a>
+                                                </Link>
                                             </Button>
                                         </div>
                                     </CardContent>
--- a/frontend/src/pages/student/digital-transcript/DigitalTranscriptWysiwygEntry.tsx
+++ b/frontend/src/pages/student/digital-transcript/DigitalTranscriptWysiwygEntry.tsx
@@ -55,6 +55,7 @@ import {
     TOP_SKILLS_MAX
 } from './transcriptReflectionConfig';
 import { LearningRecordTracker } from './learningRecordAnalytics';
+import { enterForm, leaveForm, noteActivity } from './learningRecordSession';
 
 /** Maps a TranscriptEntry patch key to its corresponding preview field id. */
 function patchKeyToPreviewField(key: keyof TranscriptEntry): string | null {
@@ -296,7 +297,6 @@ export function DigitalTranscriptWysiwygEntry({
     const analyticsRef = useRef<LearningRecordTracker | null>(null);
     analyticsRef.current ??= new LearningRecordTracker();
     const analyticsStartedForRef = useRef<string | null>(null);
-    const entriesCommittedThisSessionRef = useRef(0);
 
     const committedIds = useMemo(
         () => new Set(entries.map((e) => e.id)),
@@ -557,11 +557,10 @@ export function DigitalTranscriptWysiwygEntry({
             setSaveErrorRowId(null);
             // Success path only, mirroring the attendance/program convention: the
             // row is complete, saved, and the resident is finishing. The tracker
-            // refuses a second completion for the same entry, so the session count
-            // only advances when an event was actually emitted.
-            if (analyticsRef.current?.entryCompleted(row)) {
-                entriesCommittedThisSessionRef.current += 1;
-            }
+            // refuses a second completion for the same entry and reports the
+            // accepted one to the session itself, so the sitting's count advances
+            // only when an event was actually emitted.
+            analyticsRef.current?.entryCompleted(row, id);
             return true;
         }, [
             persistActiveRow,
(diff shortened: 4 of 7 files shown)


carddev81 (approved):
@corypride looks good, and am going to approve and merge, but if you have time check this. The finding was a possible duplicate of criterion #4.  I think it is okay though.

PR #1216, opened 2026-08-25
fix: hide Knowledge Center FAQ content and guided tour when KC is disabled

corypride:

## Description of the change

Residents were seeing FAQ questions and a guided tour for features their facility doesn't even have turned on. A resident on a facility with only Program tracking enabled would open the FAQ and see questions like "What is UnlockEd's Knowledge Center?" or "Where is my record saved?" (Learning Record) — features they can't access and that don't apply to them at all.

This PR makes the FAQ (and the guided tour) only show content for features the facility actually has turned on:

- **Knowledge Center off** → hides the "Overview" and "Finding and Using Content" FAQ sections, drops the Knowledge Center mention from the "What's next for UnlockEd?" answer, and removes the "Start Tour" button (the guided tour only walks through Knowledge Center screens, so it doesn't make sense without it). Also stops the tour from popping up automatically the first time a resident logs in, if Knowledge Center is off.
- **Learning Record off** → hides the two Learning Record FAQ questions ("What does 'print or share' mean?" and "Where is my record saved?"), and trims the "Is my activity on UnlockEd private?" answer so it doesn't reference Learning Record when the resident doesn't have that feature.
- Also found and fixed one more spot that was leaking Knowledge Center content: "Does UnlockEd have a way to track my progress?" reads like generic account copy, but the answer is actually describing the Knowledge Center library experience — it's now hidden when Knowledge Center is off too.

Nothing else in the FAQ needed changing — the rest (name/username, password reset, general help questions) applies to everyone regardless of which features are on.

### Where to see it

https://github.com/user-attachments/assets/e03372cb-91f2-4faa-8b93-21978db3ba17

Log in as a resident, then open **FAQ** from the left nav (or go to `/faqs`). Toggle Knowledge Center and Learning Record on/off for the resident's facility from `/feature-control` as an admin, then reload the FAQ each time to see it update. A video of this walkthrough is attached to the PR.

- **Related issues**: [Asana Task](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1216768705639775)]

## Additional context

This only touches frontend copy/visibility logic — no backend, database, or migration changes.

### DIFF
--- a/frontend/src/data/faqData.ts
+++ b/frontend/src/data/faqData.ts
@@ -10,7 +10,7 @@ interface FAQEntry {
     extra?: string;
 }
 
-export const FAQ_CATEGORIES: Record<string, FAQEntry[]> = {
+const KNOWLEDGE_CENTER_CATEGORIES: Record<string, FAQEntry[]> = {
     Overview: [
         {
             question: "What is UnlockEd's Knowledge Center?",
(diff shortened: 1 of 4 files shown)


frontend/src/data/faqData.ts
```diff
@@ -80,53 +80,75 @@ export const FAQ_CATEGORIES: Record<string, FAQEntry[]> = {
...
+                ? [
+                      {
+                          question:
+                              'Does UnlockEd have a way to track my progress?',
+                          answer: "Right now, UnlockEd is a library of resources you can explore freely. Some of these resources may have more of a learning path built into them, but the tool doesn't guide you along or track your progress like some other courses do. We're considering ways to help users track their learning in the future. We're always eager to hear what you'd find useful, so please let us know!"
+                      }
+                  ]
+                : []),
+            {
+                question: 'Is my activity on UnlockEd private?',
```
carddev81: This Q/A is KC related, can be removed too.

carddev81 (changes requested):
@corypride good work! Just one slight change. Found something that should not be displayed to user when KC is turned off. If you disagree let me know.

frontend/src/data/faqData.ts
corypride: Good catch! I agree.

carddev81 (approved):
Looks good!!

PR #1220, opened 2026-09-11
Repair the first-login tour and make client crashes diagnosable (ID-846)

corypride:

## What went wrong

A resident who had just finished their first login tapped **Start my record** and got the generic
"Something went wrong" page. Nothing about it reached our logs, so there was no way to tell from
Maine's deployment what had actually failed.

## Why

Every time the achievement editor opened, it saved a small amount of information into the browser.
Some devices are set to refuse that — a locked-down browser policy is ordinary on facility
equipment — and the refusal took the whole page down instead of being shrugged off.

Two things to be straight about. This is the best explanation we have, found by reading the code; we
did not manage to make it happen on a test machine. And it was not limited to the **Start my
record** button as an earlier version of this description said — the failing save ran every time the
editor opened, by any route, so more residents could have hit it than first thought.

## What changed

- The achievement editor now works on a device that refuses to store anything.
- The same protection was applied everywhere else the app stores information, including the login
  page, which could previously be left with no way to sign in at all.
- Crashes now report themselves to the server instead of vanishing. The error page shows the
  resident a reference number they can quote.
- Three permission refusals that used to happen silently now leave a log line.
- Passwords are no longer written into the logs.
- The first-login guided tour works again. It has been silently dead for every resident whose first
  login goes through a password reset — a real bug, just not this one. Expect it to reappear.
- The achievements list no longer says a program name is missing when the resident has entered one.

## Open items

The ticket says refreshing made it work, and none of what we fixed behaves that way. Something else
may remain, and the new crash reporting is what will name it if it does.

**⚠️ Needs a decision outside this PR.** Passwords have been going into the logs from `main` for some
time, so existing logs already contain real ones. This stops new ones; it does nothing about what is
already written. Someone needs to decide on retention and disclosure.

## How to check it

| Where | What to do | What should happen |
|---|---|---|
| `/home` → **Start my record** | Set the browser to block site data for this origin first (Chrome: Settings → Privacy → Site settings → "Don't allow sites to save data") | The editor opens and typing still saves. Previously: the error page. |
| `/login` | Same blocked-storage browser | The login form appears and sign-in works. |
| First login as a resident | Log in with an account that still has to set a password | After setting it you land on the home page with the guided tour running. |
| Any page that errors | Force an error | The error page shows a **Reference** timestamp, and the server log gains a matching client-error line with the details. |
| Learning Record list | Save an achievement named `a` | The row reads "Untitled achievement" and matches the steps-answered count beside it. Previously it said "Not added yet". A normal name like `Welding 101` shows as typed. |
| Backend logs | Sign in with a wrong password | The logged form shows `"password":"[REDACTED]"`. |

- **Related issues**: addresses [Asana ID-846](Asana ID-846).

---
  - https://app.asana.com/0/0/1218469139443738

### DIFF
--- a/backend/src/handlers/auth.go
+++ b/backend/src/handlers/auth.go
@@ -139,6 +139,13 @@ func (s *Server) authMiddleware(next http.Handler, resolver RouteResolver) http.
 		// Call to named or custom resolver by routedef
 		if resolver != nil {
 			if !resolver(s.Db, r.WithContext(ctx)) {
+				log.WithFields(log.Fields{
+					"handler":     "authMiddleware",
+					"method":      r.Method,
+					"path":        r.URL.Path,
+					"user_id":     claims.UserID,
+					"facility_id": claims.FacilityID,
+				}).Warn("route resolver denied access to resource")
 				http.Error(w, "User is not allowed to view this resource", http.StatusUnauthorized)
 				return
 			}
--- a/backend/src/handlers/auth.go
+++ b/backend/src/handlers/auth.go
@@ -132,6 +132,28 @@ func (s *Server) authMiddleware(next http.Handler, resolver RouteResolver) http.
 		}
 
 		if claims.PasswordReset && !isAuthRoute(r) {
+			// Logged because this gate was completely silent, and it is not a
+			// benign one. Every endpoint the reset page itself calls is exempt
+			// via isAuthRoute, so reaching here means something else asked for a
+			// route it cannot have yet -- including /api/client-errors, which is
+			// not exempt: a crash reported by a resident who still carries
+			// password_reset never reaches the log, because the reporter gets
+			// this response instead of the handler (client_error_handler.go).
+			//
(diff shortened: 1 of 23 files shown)


carddev81 (changes requested):
Weird issue with 'Not added yet' for the Program column...it is not displaying the changes I made....see video for complete picture: 

https://github.com/user-attachments/assets/d5ab4757-d836-4107-8267-9923a5d61c52

2. transcriptEntrySessionStorage.ts line 369 and 376 are functions that make calls to the session storage, I noticed that you created a safeSessionStorage.ts file that utilizes the functions within but here at these lines you don't use those functions? For consistency could we use these here too?

3. There seems to be a lot of dead unused code within this learning record section. Should we remove this now or just create a ticket for it?  I say just create a ticket for later on. Stuff in the backend too is dead pertaining to this section.

carddev81 (approved):
Good to go!!

PR #1223, opened 2026-09-15
Add account_created_at variable to things that posthog tracks

corypride:
Adds `account_created_at` to the PostHog person properties set at login, so pilot
metrics can be segmented by when an account was set up — separating residents and
staff stood up during pre-pilot setup from those added later. Today those two
populations are indistinguishable in every insight.

Frontend only. `users.created_at` is already served by `GET /api/auth`
(`backend/src/handlers/auth.go:227`) and already typed on `User`, so no backend,
API or migration change was needed. The change is three lines in
`frontend/src/lib/events.ts` — `identifyUser` is the only `posthog.identify` call
site in the app.

## Where to see this change

This is an analytics change with no UI surface, so the "route" is the PostHog UI. I created a sample insight with the variable to show that it is being sent from v2.
image
 
image

## Quickest way to confirm it is firing

Add one dashboard tile — this needs no PostHog configuration at all:

1. New insight → **Trends**
2. Series: `All events` (or `Pageview`), math set to **Unique users**
3. **Filters** panel → **Person properties** → `account_created_at` → operator **is set**
4. Display as **Number**, title it `Users reporting account_created_at`
5. Save & add to dashboard

Nonzero means it is firing. It starts at 0 and climbs as people log in, so it also
doubles as a rollout progress indicator.

Use the **Filters** panel rather than the funnel icon on the series row — the
funnel scopes to a single series, the panel scopes to the whole insight.

**`is set` works on an untyped property**, so this verification does not depend on
the DateTime step below. That step only matters for actual date filtering.

## Verifying by hand

1. Log in as any user.
2. In PostHog open that person — distinct_id is `<state>:<user id>`, e.g. `test:35`.
3. `account_created_at` should be listed, matching that user's `users.created_at`.

**Three traps that make this look broken when it isn't:**

- **A brand-new account shows nothing until its forced password reset is
  finished.** First login lands on `/reset-password`, which sits in `globalRoutes`
  (unauthenticated) — `AuthProvider` never mounts there, so nothing is identified.
  Complete the reset, land in the app, then check. Verified: 0 property sends on
  the reset screen, correct value immediately after.
- **Switch accounts using the app's logout, not by navigating to `/login`.**
  posthog-js keeps the previous user's `distinct_id` and silently refuses to
  re-identify to a different one. The app's logout calls `resetAnalytics()`, which
  is what clears it. Skipping it means the second user's properties never send.
- **`account_created_at = 2026-09-15` will never match.** The value is a full
  RFC3339 timestamp (`2026-09-15T17:42:17.486383Z`). See the config step below.

## One-time PostHog config, needed only for date filtering (not in this repo)

Go to the property definitions page — **Data → Property definitions** (under the
**Schema** heading; older PostHog builds call this Data management → Properties).
Select `account_created_at`, **Edit**, and set the property type to **DateTime**.

Until then PostHog treats it as a string, `=` is exact string equality, and the
date operators are not offered at all. Once typed, the available operators are
**is date before** and **is date after** — there is no "is date equals", so a
single day is expressed as `after X` AND `before X+1`. A date-only value is read
as midnight, so `is date before 2026-09-16` captures all of the 15th.

The property only appears on that page once PostHog has ingested at least one
event carrying it, which doubles as a useful diagnostic: if it is missing, nothing
carrying the property has reached that project yet.

## Scope: this is forward-looking, not retroactive

Two limits worth knowing before anyone builds a dashboard on it:

- **Historical events are not covered by a property filter.** This project uses
  PostHog's persons-on-events mode (default post-June-2024), which stamps person
  properties onto each event at ingestion. Events sent before this deploys carry
  no `account_created_at` and are excluded from any filter on it, even after the
  user logs in. For history, build a **cohort** instead — cohorts evaluate against
  current person properties and do reach back over old events.
- **Users who never log in again never get the property**, so they land in neither
  cohort and disappear from both series of any comparison, with no visual
  indication. If the pre-pilot population includes people who have since drifted
  off, that is exactly the group the analysis would want counted.

Closing that second gap needs a one-off backfill: read `id` / `created_at` from
the users table and POST `$identify` to PostHog with `distinct_id` =
`<state>:<id>`, populating every existing user without waiting for logins. Not in
this PR — flagging it as the follow-up that makes the pre-pilot comparison
complete rather than partial.

## Testing

- `tsc --noEmit` and eslint/prettier clean.
- Verified against the real wire payload (not mocks) using the local analytics QA
  harness, for three paths: existing staff, existing resident, and a
  freshly-created resident driven through the forced password reset. In each case
  the value sent matched that user's `created_at` from `/api/auth` to the same
  instant, and the existing `role` / `features` / `facility` properties were
  unchanged.
- PostHog accepted the captures: `/e/` and `/i/v0/e/` both returned `200`.
- PII guarantee re-checked: no username appears anywhere in the captured payload.


carddev81 (approved):
Looks good.

PR #1224, opened 2026-09-17
feat: add essential ed implemetation per id 808

carddev81:

## Description of the change

Implemented Essential Education LMS into unlocked. Many changes to backend and frontend. Mimic'd what Canvas LMS was doing, buy pulling programs and classes.

NOTE TO DEV: Get with @carddev81 for how to test.

- **Related issues**: Link to related Asana ticket that this closes: [Implement Essential Ed into Unlocked](https://app.asana.com/1/1201607307149189/project/1209460078641109/task/1217118441878654)

Extra notes: 

# Running and testing the learning platform integrations

Two providers are read live from their APIs and projected as synthetic programs and classes:
**Canvas** and **Essential Education**. Neither is synced into our tables.

Canvas setup already lives in `README.md` (the `- _Canvas_:` bullet) — start there for Canvas.
This file covers Essential Education, which is not documented anywhere else, and adds the Canvas
gotchas the README does not mention.

---

### There is nothing to run

Unlike Canvas, Essential Ed has **no container, no `make` target and no seed**. It is a hosted
sandbox, so the entire setup is pointing UnlockEd at it: a URL and a token. Don't go looking for
a compose service.

There is also no migration — `provider_platforms.type` is a plain `varchar(100)`, so
`essential_ed` needs nothing added.

### 1. Enable the feature flag

Enable `provider_platforms` for your facility. Every live-provider call is gated on it, and so is
the sidebar item, so the page below is invisible without it.

### 2. Add the platform

**Sidebar → Learning Platforms → Add Learning Platform**

| Field | Value |
| --- | --- |
| Name | anything — it becomes the program name (use `Essential Ed`) |
| Platform Type | **Essential Education** (last of the four options) |
| Provider Instance URL | `https://api.essentialed.com/v20181129` |
| Access Token | the sandbox token — **ask Rich** |

Then **Add Platform**.

Two things to get right on the URL:

- **No trailing slash.**
- **No leading or trailing whitespace.** Pasting a URL with a leading space used to produce
  `first path segment in URL cannot contain colon` on every call. The value is trimmed on save and
  on read now, but it is still the sort of thing worth looking at twice.

The token is sent as the `X-API-Token` header, not a bearer token
(`essential_ed_programs.go`, `essentialEdTransport.authorize`).

> **TODO:** where to mint a fresh token in the Essential Ed console
> (`app.essentialed.com/manage/unlocked-labs-api-test-district` → ?). Don't reuse an old one that
> has been passed around.

### 3. Map users — do not skip this

The Add dialog gets you a connection, but **every class will read 0 enrolled until residents are
mapped to Essential Ed users.** `provider_user_mappings` starts empty, and the enrollee counters
intersect the provider's students against that empty set. This looks exactly like a broken
integration and is not one.

**Learning Platforms → click the platform row → Manage Users**

Matching runs automatically when the page loads — there is no "match" button to find. **Refresh**
re-runs it. You get three buckets:

- **Auto-confirmed** — matched on name similarity. Confirm each.
- **Needs review** — possible matches. Confirm, or pick a different resident.
- **Unmatched** — no match. **Create** a resident, or **Link existing**.

Per row: **Confirm**, **Link existing**, **Create**. Per bucket: **Confirm all**. When you're
done, **Apply N matches** → **Apply matches**.

### 4. Check it worked

1. **Programs** page lists the Essential Education program next to Canvas.
2. Open it — classes are **College Prep I** and **High School Prep I**.
3. Log in as a resident you mapped — **My Programs** shows those classes.

### Sandbox reference

- District **303**; schools **6592** and **6593**.
- 3 students: `student1` / `student2` / `student3`, ids **4329946–48**.
- 2 classes, both in school 6592: **College Prep I** (69742) and **High School Prep I** (69743).
  Student1 and Student2 are in both; Student3 is in neither, so a resident mapped to Student3
  correctly shows nothing.
- **Classes can only be created in the web console.** There is no `POST /classes` — the API
  creates and updates *users* only.


backend/src/handlers/canvas_programs.go
```diff
@@ -1535,6 +1660,59 @@ type canvasScheduleEvent struct {
...
+		return events
+	}
+	if loc == nil {
+		loc = time.UTC
+	}
+	for day := start; !day.After(end); day = day.AddDate(0, 0, 1) {
+		if !course.startDt.IsZero() && day.Before(course.startDt.Truncate(24*time.Hour)) {
+			continue
+		}
+		if course.endDt != nil && day.After(*course.endDt) {
```
CK-7vn: course.endDt is parsed as bareUTC (Unless i'm reading it wrong, I just went to the time package and read the code), and day.After is using the facility timezone, so I'm pretty sure this will drop the meeting on the actual last day of class silently.

backend/src/models/provider_platforms.go
CK-7vn, line 70: Would it be worth adding a url.Parse(...)Scheme != https in this hook and the BeforeUpdate hook just incase so nothing gets sent as plaintext if for some reason it ends up going to a non-https link?

CK-7vn (commented):
Hey man, great great work, another home run here, manual testing went great. I did read and confirm a couple of the code rabbit findings that I don't think are wrapped up yet, I would just fix those, and I left 2 comments, deferring to your professional opinion on those, I'll re-review as soon as you're done and get it merged hopefully before you're on PTO.

CK-7vn (approved):
Nice work man, fixed stuff I didn't even notice.

PR #1225, opened 2026-09-17
fix: first time login tour and add logging to account for issue noted…

corypride:
## What went wrong

A resident who had just finished their first login tapped **Start my record** and got the generic
"Something went wrong" page. Nothing about it reached our logs, so there was no way to tell from
Maine's deployment what had actually failed.

## Why

Every time the achievement editor opened, it saved a small amount of information into the browser.
Some devices are set to refuse that — a locked-down browser policy is ordinary on facility
equipment — and the refusal took the whole page down instead of being shrugged off.

Two things to be straight about. This is the best explanation we have, found by reading the code; we
did not manage to make it happen on a test machine. And it was not limited to the **Start my
record** button as an earlier version of this description said — the failing save ran every time the
editor opened, by any route, so more residents could have hit it than first thought.

## What changed

- The achievement editor now works on a device that refuses to store anything.
- The same protection was applied everywhere else the app stores information, including the login
  page, which could previously be left with no way to sign in at all.
- Crashes now report themselves to the server instead of vanishing. The error page shows the
  resident a reference number they can quote.
- Three permission refusals that used to happen silently now leave a log line.
- Passwords are no longer written into the logs.
- The first-login guided tour works again. It has been silently dead for every resident whose first
  login goes through a password reset — a real bug, just not this one. Expect it to reappear.
- The achievements list no longer says a program name is missing when the resident has entered one.

## Open items

The ticket says refreshing made it work, and none of what we fixed behaves that way. Something else
may remain, and the new crash reporting is what will name it if it does.

**⚠️ Needs a decision outside this PR.** Passwords have been going into the logs from `main` for some
time, so existing logs already contain real ones. This stops new ones; it does nothing about what is
already written. Someone needs to decide on retention and disclosure.

Uploading yhs-hvda-hif (2026-09-16 22_24 GMT-5).mp4…

https://drive.google.com/file/d/1GX1TBQr-6l3857MPa3OCJBq_tuFxkCnx/view?usp=sharing


carddev81 (approved):
Tested and looks good. Same as what you merged into main already

